PR #2, opened 2013-10-22
Ignore duplicate names even if configured.

fingon:
This will drop duplicate names from search path - they won't do anything. (Obviously, getting rid of duplicate names to start with would be the correct step, but this makes my home network look prettier right now.. _grin_)


fingon:
Hm, guess netifd fix took care of it in any case.

PR #3, opened 2013-12-05
Fixes for UCI loading and possible segfault

dedeckeh:
Hi Steven,

I have worked on two commits; the first commit (47bcdec) fixes the UCI loading of the odhcpd section in the dhcp topic as I was unable to set the leasefile and leasetrigger parameters while the last commit fixes possible segfault conditions in the odhcpd code (0b68cab).

Bye,
Hans


sbyx:
Thanks for the patch. However please note that alloca may not return NULL.

dedeckeh:
Thx Steven. Was not aware alloca may not return NULL

PR #4, opened 2013-12-10
Bugfixes

dedeckeh:
Hi Steven,

I have worked on 3 bugfixes for the odhcpd daemon; first patch (a946e2) fixes hop limit scanning in the ancillary data while the second patch (f587736) fixes router advertisement/solicitation option parsing as the option length is variable for each option. Last patch (3f4117e5) checks the sanity of the router advertisement/solicitation messages as some logic from the previous option parsing is moved to this patch and lines up with RFC4861 regarding message checking. If you have feedback/comments regarding the patches just let me know.

Bye,
Hans


sbyx:
Thanks.

PR #5, opened 2013-12-12
Fixes for option parsing

dedeckeh:
Hi Steven,

I've made two patches for fixing issues noticed during parsing of icmpv6 options (introduced by one of my previous commits); first patch (47ee3f8) fixes issue during icmpv6 option parsing when validating the icmpv6 message while the last patch (f40d224) fixes issue in the icmpv6 option iterator macro caused by lack of valid pointer casts.

Bye,
Hans

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -134,7 +134,7 @@ static int router_icmpv6_valid(struct sockaddr_in6 *source, uint8_t *data, size_
 {
 	struct icmp6_hdr *hdr = (struct icmp6_hdr *)data;
 	struct icmpv6_opt *opt;
-	size_t optlen = len - sizeof(*hdr);
+	size_t optlen;
 
 	/* Hoplimit is already checked in odhcpd_receive_packets */
 	if (len < sizeof(*hdr))
@@ -149,10 +149,12 @@ static int router_icmpv6_valid(struct sockaddr_in6 *source, uint8_t *data, size_
 			return 0;
 
 		opt = (struct icmpv6_opt *)((struct nd_router_advert *)data + 1);
+		optlen = len - sizeof(struct nd_router_advert);
 		break;
 
 	case ND_ROUTER_SOLICIT:
 		opt = (struct icmpv6_opt *)((struct nd_router_solicit *)data + 1);
+		optlen = len - sizeof(struct nd_router_solicit);
 		break;
 
 	default:
--- a/src/router.h
+++ b/src/router.h
@@ -37,8 +37,10 @@ struct nd_opt_recursive_dns {
 };
 
 #define icmpv6_for_each_option(opt, start, end)\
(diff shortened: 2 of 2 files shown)


dedeckeh:
Hi Steven,

I'm a bit confused the pull request is closed without a merge; correct me if I'm wrong. Currently the message validation of router advertisement messages in router_icmp6_valid will fail as the calculated option length is wrong (need to substract router advertisement header size and not icmp6_hdr size).
The macro icmpv6_for_each_option will fail in case the option data size is not 6 bytes; eg if recursive dns option is first in the router advertisement message and the source link address is next the parsing of the source link address option will fail as the macro will jump to the wrong location as I have seen on my test setup.

Bye,
Hans

sbyx:
Hi,

I've reverted some of your earlier stuff because it was causing problems (e.g. changing the ICMPv6 option struct and iterator macro). So most of this request should not be relevant any more.

Steven

dedeckeh:
Hi Steven,

A bit surprised by the revert of all my stuff; especially the router advertisement/solicitation sanity check as described in RFC4861 paragraph 6.1. This was brought to my attention when the Tahi IPv6 Core Test suite was ran against my OpenWRT device with odhcpd daemon enabled; tests C,D, E and F from v6LC.2.2.4 (https://www.ipv6ready.org/docs/Core_Conformance_Latest.pdf page 151) were failing.
Tahi Core Tests are a reference for CPE home devices; so fixing these failures could only improve the robustness of the odhcpd daemon. The previous fix regarding hoplimit checking was also detected by running the Tahi Core test suite.
My initial changes regarding the ICMPv6 option struct and iterator macro had indeed problems; my fault as I had to do better testing before launching the pull request (sorry for that). I know from own experience it can be annoying to take in faulty code but I'm quite happy to fix these issues if they're brought to my attention. Therefore I launched pull request 5 to solve these issues; so for 2014 I will set as one of my targets better testing so I can still contribute to the OpenWRT project in a positive manner :)

Hans

sbyx:
Hi,

I did revert the sanity checks for the reason that it was broken and could lead to a buffer overrun. This was before I read through this pull request in detail. I'm grateful for your patches anyway and should have read through them more thoroughly myself. Anyway: I just commited a slighly adapted version using the usual iterator macro to not have so much code duplication in any case.

I hope that resolves it.
Also regarding the icmpv6_for_each_option macro. I have that in several projects (e.g. also odhcp6c) so I don't really want to change it without having a good reason because then I would have to change it in all projects otherwise it would get messy.

Thanks,

Steven

dedeckeh:
Hi Steven,

Confirmation the above mentioned Tahi core tests are passing with the adapted version.

Thx,
Hans

sbyx:
Thank you very much for your work, also for the recent pull request.

PR #7, opened 2014-01-14
Fix interface ignore overwrite

dedeckeh:
Hi Steven,

Pull requests fixes overwrite of the interface ignore parameter; which was initially set when parsing the interface UCI, and later overwritten when resolving the interface ifindex.
If the interface ifindex cannot be resolved the interface is not set in use. By not setting the interface in use the interface will be closed and no dhcpv4/6 and router initialization will be done (which requires a valid ifindex anyway). 

Bye,
Hans

cannot be setup and needs to be closed.


sbyx:
Thanks.

PR #8, opened 2014-01-16
Don't rearm RS timer if callback is not set

dedeckeh:
Hi Steven,

Patch does not rearm the RS timer if callback is not set as it could lead to potential issues when interface is being closed due to invalid config.  

Bye,
Hans


sbyx:
Thanks again. I will push odhcpd in the coming days in OpenWrt after running some generic tests.

PR #9, opened 2014-01-23
Fix ra_management and ndproxy_routing config overwrite

dedeckeh:
Hi Steven,

During testing I observed the ra_management and ndproxy_routing UCI config was overwritten in odhcpd when ubus_apply_networkw as called as the attributes IFACE_ATTR_RA_MANAGEMENT and IFACE_ATTR_NDPROXY_ROUTING are not present in the blobmsg (and thus overwrite the initial values set by the UCI config). Solved by the issue by testing the overwrite parameter but I'm not 100% sure the parameter overwrite is intended for this purpose.
Otherwise everything seems to work fine; nice job !

Regarding 6rd I pushed a patch to the OpenWRT devel mailing list to make the "don't fragment" bit of 6rd tunnel interfaces configurable via UCI (one of my previous patches in netifd offers this support); do you see it as an usefull extension for OpenWRT as normally the "don't fragment" bit should be disabled according to the RFC while it's currently enabled.

Br,
Hans

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -454,7 +454,7 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 
 	if ((c = tb[IFACE_ATTR_RA_MANAGEMENT]))
 		iface->managed = blobmsg_get_u32(c);
-	else
+	else if (overwrite)
 		iface->managed = 1;
 
 	if ((c = tb[IFACE_ATTR_RA_OFFLINK]))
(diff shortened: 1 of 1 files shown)


sbyx:
Thanks for the fix. Let me have a look at your patch and see if i can merge it. If the RFC discourages setting DF then I wonder why someone set it as default.

dedeckeh:
Thx for applying the 6rd patch Hans

PR #10, opened 2014-02-25
Fix parsing of static ndp entries

dedeckeh:
Hi Steven,

Patch fixes parsing of static configured ndp entries. The strtok_r iteration was failing for multiple prefixes in the list as each prefix was null terminated in the static ndp parameter; additionally sscanf was failing for each entry in the list.

Hans

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -494,6 +494,9 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 			if (!iface->static_ndp)
 				goto err;
 
+			if (iface->static_ndp_len)
+				iface->static_ndp[iface->static_ndp_len - 1] = ' ';
+
 			memcpy(&iface->static_ndp[iface->static_ndp_len], blobmsg_get_string(cur), len);
 			iface->static_ndp_len += len;
 		}
--- a/src/ndp.c
+++ b/src/ndp.c
@@ -167,6 +167,7 @@ int setup_ndp_interface(struct interface *iface, bool enable)
 			memcpy(entry, iface->static_ndp, iface->static_ndp_len);
 
 			for (entry = strtok_r(entry, " ", &saveptr); entry; entry = strtok_r(NULL, " ", &saveptr)) {
+				char *sep;
 				struct ndp_neighbor *n = malloc(sizeof(*n));
 				if (!n) {
 					syslog(LOG_ERR, "Malloc failed for static NDP-prefix %s", entry);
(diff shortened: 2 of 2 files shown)


sbyx:
Cherry-picked the essential commit. Thanks.

dedeckeh:
Patch in config.c needs to be applied to solve the issue completely as each prefix is null terminated which makes strtok_r fail for multiple prefixes. Will do it in a cleaner way next time; sorry for that

dedeckeh:
Hi Steven,

I have a question related to the 5 netifd patches I submitted to the OpenWRT devel mailing list on Feb the 11th.
I noticed 2 out of the 5 patches were applied but none of the linksensing and bridge mtu patches.
The linksensing patches were initially applied by Felix Fietkau in netifd but later reverted as regression issues were seen (https://dev.openwrt.org/ticket/14590).
At that time the request was made to resubmit the netifd linksensing patches together with the solution.
In fact the regression issue was caused by an existing bug in netifd which is solved by the first netifd patch (https://lists.openwrt.org/pipermail/openwrt-devel/2014-February/023793.html) which was applied.
I tried to get feedback on the OpenWRT devel mailing list regarding the resubmitted linksensing patches but without success.
Was there any internal discussion between the OpenWRT developers regarding the relevance of the linksensing patches or is there any other reason why these patches were not applied ?
The primary use case implemented by the linksensing patches is restarting the dhcp client on a wan interface when the link flaps as intermediate devices in the access network clean up the relation between the IP address and the MAC address.
By restarting the DHCP client the connectivity is restored otherwise connectivity will only be restored when a rebind is transmitted which will depend on the leasetime given by the server as the udhcp client does not monitor the link state.

Thx,
Hans

sbyx:
No it hasn't really been discussed. I'll just try to bring it up again.

dedeckeh:
Need to check if the linksensing patches still apply after the latest
changes in netifd. Can you give me a sign if it has been discussed; if the
patches don't apply anymore I will resubmit them.

sbyx:
Pushed your link-sensing stuff again. Feel free to resubmit the bridge stuff but I think it needs some refactoring.

dedeckeh:
Thx for applying the linksensing patches.
Regarding the bridge MTU stuff what's the major potential problem you see ?

sbyx:
Nah I mainly mean it doesn't apply cleanly anymore.

dedeckeh:
Hi Steven,

I resubmitted the bridge MTU stuff together with two other netifd patches. Last two patches fix a wrong ubus interface status and a missing kernel route when a route is updated.

Br,
Hans

sbyx:
Felix just commited them. Thanks again.

PR #13, opened 2014-04-24
Bugfixes

dedeckeh:
Hi Steven,

The patches fix two issues regarding status code handling for decline messages and confirm messages without any address options which I noticed during testing.
Could you also have a look at the 2 netifd and the 6rd patches I just submitted on the mailing list ?
The 6rd patch allows the user to select the source IP address based on the tunlink parameter; basically the same logic is already present in the dslite script.
The first netifd patch fixes the removal of dynamic interfaces when a config reload is triggered; the dynamic node version marker is overwritten en interface_set_dynamic was done on an already freed pointer.
Last netifd patch only installs and cleans up the source policy route and other stuff when the prefix address can be installed.

Thx,
Hans

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -1191,7 +1191,8 @@ ssize_t dhcpv6_handle_ia(uint8_t *buf, size_t buflen, struct interface *iface,
 		response_len += ia_response_len;
 	}
 
-	if (hdr->msg_type == DHCPV6_MSG_RELEASE && response_len + 6 < buflen) {
+	if ((hdr->msg_type == DHCPV6_MSG_RELEASE  || hdr->msg_type == DHCPV6_MSG_DECLINE) &&
+			response_len + 6 < buflen) {
 		buf[0] = 0;
 		buf[1] = DHCPV6_OPT_STATUS;
 		buf[2] = 0;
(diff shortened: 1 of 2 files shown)


sbyx:
Thanks. I'll have a look.

dedeckeh:
Thx for checking/applying the netifd/6rd patches
Few weeks ago I submmited 3 other netifd patches (https://lists.openwrt.org/pipermail/openwrt-devel/2014-April/024637.html https://lists.openwrt.org/pipermail/openwrt-devel/2014-April/024638.html https://lists.openwrt.org/pipermail/openwrt-devel/2014-April/024639.html); don't know if they have been reviewed by you guys as I got the message from Felix the patches will be reviewed soon beginning last week.

dedeckeh:
Hi Steven,

I noticed your latest git commit (http://git.openwrt.org/?p=openwrt.git;a=commit;h=a5f5e5e258d9693aea8d1e62997e8451f5f185ff)  in the odhp6c package.
The issue you noticed is probably solved by one of the netifd patches  (https://lists.openwrt.org/pipermail/openwrt-devel/2014-May/025258.html but not yet applied) I submitted last wednesday.
I noticed the same issue with 6rd tunnels that after a dhcp renew (and thus a config update) the dynamic 6rd tunnel was brought down in netifd; fixed the issue with this patch and the 6rd tunnel stays up after a config update. Quite confident this patch will also solve the problems for dslite/map

Bye,
Hans

sbyx:
Ah nice. Thanks I will review the series later today.

PR #14, opened 2014-06-09
Fix DHCPv6 relay reply message in case raw DHCPv6 attributes are present

dedeckeh:
Hi Steven,

Patch fixes length of DHCPv6 relay reply in case raw DHCPv6 attributes are present.
Additional enums are used to index iov struct.

Bye,
Hans


sbyx:
Thanks

PR #17, opened 2014-07-01
leasefile: do not output expired dhcpv6 leases

mar-kolya:
Sometimes ip that is assigned to a host changes, old ip is still kept aroung as 'expired'.
This expired ip gets dumped into leasefile and is read by dnsmasq.
The result is that hotsname is resolved into expired ip and that confuses clients.
This patch prevents expired leases from being written into leasefile.


sbyx:
Thanks.

PR #18, opened 2014-07-01
Improved compliance with RFC3315

dedeckeh:
Hi Steven,

Added destination address check in DHCPv6 server as described in RFC3315; last patch adds inf_max_rt option in reply message in response to information request.

Bye,
Hans


sbyx:
Thanks.

PR #20, opened 2014-08-12
dhcpv4: implement router configuration option

mehlis:
RFC 2132 defines:

3.5. Router Option

   The router option specifies a list of IP addresses for routers on the
   client's subnet.  Routers SHOULD be listed in order of preference.

   The code for the router option is 3.  The minimum length for the
   router option is 4 octets, and the length MUST always be a multiple
   of 4.

```
Code   Len         Address 1               Address 2
```

   +-----+-----+-----+-----+-----+-----+-----+-----+--
   |  3  |  n  |  a1 |  a2 |  a3 |  a4 |  a1 |  a2 |  ...
   +-----+-----+-----+-----+-----+-----+-----+-----+--

### DIFF
--- a/README
+++ b/README
@@ -85,6 +85,8 @@ ndp		string	disabled		Neighbor Discovery Proxy
 
 dynamicdhcp	bool	1			dynamically create leases
 						for DHCPv4 and DHCPv6
+router          list    <local address>         Routers to announce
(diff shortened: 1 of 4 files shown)


sbyx:
Thanks.

PR #21, opened 2014-08-12
dhcpv4: offer a valid configuration with DHCP NAK

mehlis:
On network change the client requests a non valid
configuration. In this case we have to reply with
a DHCP NAK. But this NAK can include a valid
configuration like an ACK does.

With this change iDevices and Androids can finish
DHCP even with wrong configuration in first
REQUEST with two packets and without any additional
timeouts and round trips.


sbyx:
Thanks.

PR #22, opened 2014-08-12
dhcpv4: log requests and replies to syslog

mehlis:
This change let odhcpd print dhcpv4 events
to syslog.


mehlis:
this let's odhcpd behave like dnsmasq

sbyx:
Fine with me as well ;)

ZupoLlask:
How can we enable logging of DHCP queries to syslog? :-)

PR #23, opened 2014-08-18
dhcpv4: send Nak to broadcast address

mehlis:
this PR addresses https://github.com/sbyx/odhcpd/issues/16.
wireshark proves that destination IP is send to "new" IP, this is wrong. NAK to IP broadcast is right (inthe most cases).

PR is split in 3 commits.

please review, I'll appy comments ASAP! I'm sure my comments are not completely valid, so far it's a guess :) feel free to correct me!


sbyx:
They are fine ;)

PR #28, opened 2014-10-21
dhcpv4_assign: try 100 times to find a IP

mehlis:
This is what the comment on the loop is saying.
Also change the srand comment to match to the actual code.

Question: Am I right with this change? Or was it desired to loop only 10 times and the comment is wrong?


sbyx:
That is fine. I have bigger rewrite of the address / prefix assignment logic planned for the near future anyway including custom DHCP options etc. etc.

mehlis:
I'm working on a new implementation of `dhcpv4_assign` and I have some code I'll test tomorrow.
I'm going to open a PR for RFC tomorrow.

PR #29, opened 2014-10-22
dhcpv4_assign: rework the assignment logic

mehlis:
RFC: I just want to hear your opinion on the new code.

the code tries to:
1. assign the requested IP
2. assign the mapped IP
3. assign the mapped IP + 1 (and walk in the possible space with mod and stuff)

question: is the order in `iface->dhcpv4_assignments` relevant for other code parts?

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -502,50 +502,74 @@ static void handle_dhcpv4(void *addr, void *data, size_t len,
 			(struct sockaddr*)&dest, sizeof(dest));
 }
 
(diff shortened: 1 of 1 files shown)


mehlis:
@sbyx this code is heavily tested and runs in production on consumer devices. Do you think you have a better solution or are there any comments on this changeset?

sbyx:
Sorry must have been lost at some point. I'm currently working on a major overhaul of the address-assignment code unifying v6 and v4, but I'm not sure when I will finish. I can still merge this for now, thanks.

PR #30, opened 2014-10-30
Fix IPv6 DNS server adddress selection

dedeckeh:
Hi Steven,

Patch fixes the selection of IPv6 DNS address in DHCPv6 and RA overwrite as an address could be selected
with preferred lifetime zero while IPv6 addresses are in use with non zero preferred lifetimes.
Patch now tries to pick the IPv6 address with the longest preferred lifetime.
Fixes also the issue an IPv6 address with preferred lifetime zero could be returned in DHCPv6
DNS server option while an IPv6 address with non zero preferred lifetime is returned as DNS
recursive RA option for the same set of available IPv6 addresses.

Hans


sbyx:
Thanks

PR #33, opened 2015-01-01
Patches to add flexibility to odhcpd's Route Info Option handling

janakj:
Hi,

While working on a moderately complex IPv6 network setup, I ran into issues with odhcpd's handling of the RA Route Info Option. By default, odhcpd generates a Route Info Option in RAs based on the existence of unreachable routes on the router and this behavior cannot be changed. This pull request implements two new configuration options which make it possible to change odhcpd's behavior somewhat.

The option "ra_unreachable" makes it possible to disable the auto-generated Route Info Option. This is useful, for example, if the router contains multiple unreachable routes or if the route's prefix is different from what should be announced via RA.

The second option "ra_route" makes it possible to manually configure additional routes to be announced in /etc/config/dhcp. This is a per-interface option, different routes can be configured for different interfaces. It's a list option with the following value format:

```
<IPv6/bits [low|medium|high]>
```

The preference field is optional (medium by default). Here's an example:

```
list ra_route "fd00::/16 medium"
```

--Jan

### DIFF
--- a/README
+++ b/README
@@ -104,6 +104,9 @@ ra_preference	string	medium			Route(r) preference
 		[medium|high|low]
 ra_unreachable  bool    1                       Announce router's unreachable
                                                 routes via Route Info Option
+ra_route        list    <IPv6/bits [low|medium|high]> Announce manually
+                                                configured routes via the RA
+						Route Info Option
 ndproxy_routing	bool	1			Learn routes from NDP
 ndproxy_slave	bool	0			NDProxy external slave
 
--- a/README
+++ b/README
@@ -102,6 +102,8 @@ ra_management	integer	1			RA management mode
 ra_offlink	bool	0			Announce prefixes off-link
 ra_preference	string	medium			Route(r) preference
 		[medium|high|low]
+ra_unreachable  bool    1                       Announce router's unreachable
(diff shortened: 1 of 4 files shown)


peterwillcn:
ping @dedeckeh

treysis:
Why has this never been merged? This would implement RFC4191 (https://tools.ietf.org/html/rfc4191) from what I understand.

marc-h38:
Just FYI:
[ FS#3056 - odhcpd: "on-link" Router Information Options pollution in Router Advertisements](https://bugs.openwrt.org/index.php?do=details&task_id=3056)

sujoshua:
This is an useful feature, why not merge it for this long time?

peterwillcn:
any news?

systemcrash:
@Alphix I think at least statically defining some routes could be a useful feature here. We'd have to pick this in ourselves since the source repo is gone.

janakj:
> @Alphix I think at least statically defining some routes could be a useful feature here. We'd have to pick this in ourselves since the source repo is gone.

Unfortunately, I no longer have the source repository after such a long time. But you can still view the contents of the patches here on GitHub. It should not be hard to re-create the commits.

PR #63, opened 2015-10-21
Don't overflow router lifetime

willmo:
This also avoids advertising an incorrectly short router lifetime when
a prefix lifetime is greater than 2 *\* 16, due to the truncation to
16 bits.


willmo:
As the comment implies, if clamping to 9000 is undesirable for some reason, it should at least clamp to UINT16_MAX.

sbyx:
Thanks will clamp to 64K. Clients must accept any value annyway and we habe some phones who randomly drop some or all RAs in standby :/

willmo:
Meaning we want to advertise the longest possible lifetime? That makes sense.

I've updated this pull request, if you'd like to use it.

sbyx:
Ah nice. Thanks again.

PR #88, opened 2016-08-14
router: skip parse_routes when ra_default > 1

stintel:
It does not make sense to scan the entire routing table if the decision
will be overridden afterwards.

This makes it possible to use odhcpd on a router with full BGP table, by setting ra_default to 2. Without this change, odhcpd is constantly busy reading /proc/net/ipv6_route and doesn't respond to router sollications in time.


stintel:
https://git.lede-project.org/?p=project/odhcpd.git;a=commit;h=93ab25b2160ac8a2da6d5dc6ead003e9b8abf9c7

PR #89, opened 2016-09-23
Fix "ignore" option, and enable logging

karlp:
Also sent to openwrt-devel and lede-dev.  No idea which path is being used for this :)


stintel:
This can be closed after 08a93676f4d3d13fa417a1f109b40d6c577dfb76

PR #108, opened 2018-04-12
odhcpd: change ignore interface tag from ignore to ignore6

Menion2k:
There are network setup where it would be required to turn off dhcpv4
and have dhcpv6 (and all the address autoconfiguration) enabled
An example is a full static ipv4 dual stack network, also an ipv4
relayed bridge with ipv6 relay. Having the tag "ignore" used by both
dnsmasq and odhcpd as interface ignore tag prevents the configuration
of such networks

Signed-off-by: Antonio Silverio <menion@gmail.com>

### DIFF
--- a/README
+++ b/README
@@ -70,7 +70,7 @@ Option		Type	Default			Description
 interface	string	<name of UCI section>	logical OpenWrt interface
 ifname		string	<resolved from logical>	physical network interface
 networkid	string	same as ifname		compat. alias for ifname
-ignore		bool	0			ignore this interface
+ignore6		bool	0			ignore this interface
 master		bool	0			is a master interface
 						for relaying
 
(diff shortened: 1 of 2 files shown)


dedeckeh:
NAK; see https://git.openwrt.org/?p=project/odhcpd.git;a=commitdiff;h=08a93676f4d3d13fa417a1f109b40d6c577dfb76

Menion2k:
Note: the mentioned patch from openwrt git is not present in the github code

dedeckeh:
No but surely you can swap to the openwrt git code ?

Menion2k:
Sure, but my point was to the github code
Anyhow I guess sooner or later the github code will be aligned (or
deprecated), so I am fine
Bye

2018-04-13 13:31 GMT+02:00 Hans Dedecker <notifications@github.com>:

PR #112, opened 2018-05-21
dhcpv6: fix ivalid json

mislavn:
```
"ipv6-prefix": [
	{
		"<some ipv6 address>",
		"preferred-lifetime": 37979,
		"valid-lifetime": 48779,
		"prefix-length": 63
	}
],
```

Just wanted to point out that the ubus call "ubus call dhcp ipv6leases" does not return a valid JSON. The ipv6 address is missing the key name.

I'm not sure what the key name should be so for now I set it to "address".
Please can somebody take a look at this.

Signed-off-by: Mislav Novakovic <mislav.novakovic@sartura.hr>


dedeckeh:
Fixed in commit https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=373495ad610cfffaeac833283d1431e6328a109a; thx

PR #116, opened 2018-11-22
config: correctly break string for prefix filter

updateing:
Prefix filter does not work when the config seems absolutely right (e.g. `2001::/16`). After some digging I found this typo :-), and verified the fix worked.


dedeckeh:
Patch pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=ae6cf8008e50ba9838a056d78cb6aea405c55b7d); thx

PR #117, opened 2018-11-25
netlink: trigger address list refresh on preference change

updateing:
Currently we fire an address list change event in following cases:

1. address changed
2. preferred time changed to/from forever
3. valid period shortened
4. preferred period shortened

However there are common cases where addresses become deprecated and then preferred again. odhcpd will catch the former event (by case 4) and sets border->assigned to 0, but will not set it back to normal as it could not get notified about the latter event. Thus all PD allocations fail due to no prefix being available.

The cases can be as common as a redial on PPP interfaces, or a manual reconnection on the wan6 interface.

This commit adds another case for firing address list change event: address became preferred from deprecated, or the other way round. The former will solve the problem aforementioned, and the latter is a subset of existing case 4 so there should not be any regression.

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -160,6 +160,7 @@ static void refresh_iface_addr6(struct netevent_handler_info *event_info)
 	struct odhcpd_ipaddr *addr = NULL;
 	struct interface *iface = event_info->iface;
 	ssize_t len = netlink_get_interface_addrs(iface->ifindex, true, &addr);
+	time_t now = odhcpd_time();
 
 	if (len < 0)
 		return;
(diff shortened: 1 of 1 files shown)


updateing:
A simpler reproduction could be:

```
ip addr change 2001:xxx/58 dev br-lan preferred_lft 0 # No preferred address, PD breaks
ip addr change 2001:xxx/58 dev br-lan preferred_lft 3600 # The address becomes preferred again but PD still does not work
```

dedeckeh:
Not necessary to add an extra check line as the line before was intended to check this case; I fixed it in commit https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=d404c7e50bd487f9d8d3b13e5ff1c7fa878c7973
Thx for reporting

PR #123, opened 2019-02-09
Dhcp sntp

davefiddes:
This short series of changesets extends odhcpd to provide details of NTP servers for clients on the network. The configuration is very similar to DNS server addressing. I don't believe that it has any impact on routing or relaying.


dedeckeh:
I'm in doubt if it's a good idea to send sntp servers by default to clients if the clients do not request the sntp server option

davefiddes:
I think you are right and it doesn't seem to be the greatest idea. However it doesn't seem to be contrary to the DHCPv6 RFC and NetworkManager and LwIP stacks do not appear to tie the option requests to the responses.

I'll have a look at implementing that behaviour as long as you are happy with the general approach?

dedeckeh:
I'm fine with the approach although it would be better to have logic in place which allows to define a DHCP option in an opaque way as it's not scalable to add a new config option for every DHCP option

davefiddes:
I'm unlikely to get to these changes anytime soon. Closing this PR to reduce clutter.

PR #134, opened 2019-07-04
Support ra_lifetime=0

jplitza:
The value 0 has the special meaning of not being a default router. This is useful to only advertise prefixes or DNS settings without being a real router.

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -354,7 +354,7 @@ static uint32_t calc_ra_lifetime(struct interface *iface, uint32_t maxival)
 
 	if (iface->ra_lifetime >= 0) {
 		lifetime = iface->ra_lifetime;
-		if (lifetime < maxival)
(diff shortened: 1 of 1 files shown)


dedeckeh:
@jplitza  I like the idea however the patch is not signed off and should start with the prefix router: support ra_lifetime being 0
Please have a look at https://openwrt.org/submitting-patches#sign_your_work

jplitza:
Fixed

dedeckeh:
Patch pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=f6c2242c7af2de8f40f5b6b8e50c350a138bc183); thx

TCB13:
How likely is this fix to have something related to the issues here: https://github.com/openwrt/odhcpd/issues/133

Issue #135, opened 2019-07-10
Regression: Router Discovery broken after ifup lan


aertsk:
## Description

Following commit: 
https://github.com/openwrt/odhcpd/commit/c6dae8e6b2a82dca6e2e00e17b0bc968a504d7fc#diff-25d902c24283ab8cfbac54dfa101ad31

introduces an issue in odhcpd after 'ifup lan' resulting in the fact that Router Discovery is broken until odhcpd is restarted. Periodic Router Advertisement's are still send out.

Following description describes the reproduction on a TP-Link 1750:

Made use of rdisc6 as part of ndsic6 package (apt install ndisc6) to trigger a Router Solicitation from a host toward the gateway.
```host# rdisc6 -1 eth0```

## REPRODUCTION CONFIG

Used following config on a TP-Link 1750:
Config introduces second bridge 'lan2' as the test bridge next to the existing 'lan' bridge that is used to control the gateway.
Run odhcpd on top of lan2 to handle Router Solicitation/Advertisement
```
root@OpenWrt:~# cat /etc/config/network 

config interface 'loopback'
	option ifname 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option ula_prefix 'fd34:c595:be3b::/48'

config interface 'lan'
	option type 'bridge'
	option ifname 'eth1.1'
	option proto 'static'
	option ipaddr '192.168.1.1'
	option netmask '255.255.255.0'
	option ip6assign '60'

config interface 'wan'
	option ifname 'eth0.2'
	option proto 'dhcp'

config interface 'wan6'
	option ifname 'eth0.2'
	option proto 'dhcpv6'

config switch
	option name 'switch0'
	option reset '1'
	option enable_vlan '1'

config switch_vlan
	option device 'switch0'
	option vlan '1'
	option ports '2 3 0t'

config switch_vlan
	option device 'switch0'
	option vlan '2'
	option ports '1 6t'

config switch_vlan
	option device 'switch0'
	option vlan '3'
	option ports '4 5 0t'

config interface 'lan2'
	option type 'bridge'
	option ifname 'eth1.3'
	option proto 'static'
	option netmask '255.255.255.0'
	option ipaddr '192.168.2.1'
	option ip6assign '60'

root@OpenWrt:~# 

root@OpenWrt:~# cat /etc/config/dhcp 

config dnsmasq
	option domainneeded '1'
[...]

PR #136, opened 2019-08-07
router: fix router discovery not working after preforming ifup

meurisa:
Tracks interface device index, forces close and reinitialization if the if index has changed
also tracks multicast membership only run add/remove membership once per interface enable/disable

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -73,6 +73,7 @@ out:
 int router_setup_interface(struct interface *iface, bool enable)
 {
 	int ret = 0;
+	struct ipv6_mreq mreq;
 
 	enable = enable && (iface->ra != MODE_DISABLED);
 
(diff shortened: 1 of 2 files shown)


meurisa:
This commit fixes : https://github.com/openwrt/odhcpd/issues/135

src/router.c
```diff
@@ -81,12 +82,20 @@ int router_setup_interface(struct interface *iface, bool enable)
 		goto out;
 	}
 
+	if (iface->ifindex != iface->current_device_index &&  iface->router_event.uloop.fd >=0 ) {
```
dedeckeh, line 85: This should not be handled in router_setup_interface as there's a specific event NETEV_IFINDEX_CHANGE which can be used to trigger actions if the ifindex changes

dedeckeh:
Besides the hacky way the patch is implemented; it has formal issues like not signed-off and not respecting the existing coding style

dedeckeh:
Issue is fixed in commit https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=000182fe4f94a5a6ec139456a2b74f0cdea13b9c

PR #139, opened 2019-08-19
fix lingering uloop socket descriptor in close socket upon NETEV_IFINDEX_CHANGE code path

aertsk:
make sure the raw socket is removed from the uloop file descriptor list before the
    socket is closed. As introduced in https://github.com/openwrt/odhcpd/commit/000182fe4f94a5a6ec139456a2b74f0cdea13b9c
    
    Related to  https://github.com/openwrt/odhcpd/issues/135


aertsk:
Closing pull Request as this is not the proper way to do it.

PR #142, opened 2019-08-31
dhcpv6: retry failed PD assignments on addrlist change

updateing:
I found that odhcpd stopped distributing prefixes to clients when IPv6 addresses on the serving interface were removed and added again. This happens pretty frequently for me, as my IPv6 addresses come solely from PD by ISP and ISP kills the PPPoE WAN session every two days. Addresses on LAN will absolutely be cleared when WAN goes down.

This change fixes the issue for me but I'm not sure whether this is the best way to do it. I've put my analysis into the commit message, please have a look. Thanks!

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -700,7 +700,7 @@ static void handle_addrlist_change(struct netevent_handler_info *info)
 				c->managed_size)
 			continue;
 
-		if (c->length < 128 && c->assigned >= border->assigned && c != border)
+		if (c->length < 128 && (c->assigned == 0 || c->assigned >= border->assigned) && c != border)
 			list_move(&c->head, &reassign);
 		else if (c != border && (c->flags & OAF_BOUND))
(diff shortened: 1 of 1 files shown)


updateing:
A simpler reproduction:

1. Disable WAN, clear ULA prefix, set something larger than /64 on the LAN. Say 2001:db8:1000::/58.
2. Reboot the router to make sure everything is in position.
3. Launch a DHCPv6 client with IA_PD request from LAN. The IA_PD assignment should succeed.
4. Remove the address from LAN.
5. Retry the IA_PD request from the same client. No prefix could be obtained at this time.
6. Add the same address to LAN again.
7. Retry the IA_PD request from the same client. The request would still fail. The client is expected to obtain a prefix like in step 3.

dedeckeh:
Thx for the patch; pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=fd93e36727bc605f6f860759d7710a43bee5215d)

PR #143, opened 2019-09-08
ndp: answer global-addressed NS manually

sfan5:
The problem is also described in the first post of https://forum.archive.openwrt.org/viewtopic.php?id=40871:

The ISP router send an NS to the global address of a host, the kernel won't route it due to link-local source. The NS go unanswered until the ISP router eventually sends the same NS as multicast.
In my case this made the ISP router delay/drop packets.

### DIFF
--- a/src/ndp.c
+++ b/src/ndp.c
@@ -295,6 +295,33 @@ static void ping6(struct in6_addr *addr,
 	netlink_setup_route(addr, 128, iface->ifindex, NULL, 128, false);
 }
 
+/* Send a Neighbor Advertisement. */
+static void send_na(struct in6_addr *to_addr,
+		const struct interface *iface, struct in6_addr *for_addr,
+		const uint8_t *mac)
+{
+	struct sockaddr_in6 dest = { .sin6_family = AF_INET6, .sin6_addr = *to_addr };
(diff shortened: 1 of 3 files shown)


dedeckeh:
I understand the need to send a NA in reponse to an unicast NS but a t least it should be checked if the IPv6 address is known. In the case you're observing such unicast NS is the IPv6 address setup as a proxy IPv6 address by the NDP module ?

sfan5:
Yes, the unicast NS happens to/for addresses shown by `ip -6 n s proxy`.
How would I check that the addr is actually set up for proxying? Does odhcpd keep an internal list?

dedeckeh:
No odhcp does not keep a list; this can be checked by doing a netlink get on the neighbor proxy kernel table

dedeckeh:
Looks good; can you squash both patches into one commit ?
Can you also provide a SoB as described in https://openwrt.org/submitting-patches#sign_your_work ?

sfan5:
Done.

dedeckeh:
The SoB needs to be your full name; can you adapt this accordingly ?

sfan5:
Has been done.

dedeckeh:
Thx for the patch, pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=91a28e4566bdae6532d3801332bef9999f43605c)

PR #158, opened 2020-06-26
improve router advertisments

ukleinek:
This is actually only a PR for a single patch, it appears as more as the mirrors seems to be outdated.

I sent the patch to the openwrt-devel list, but this seems to be broken currently, so reproduce here for some visibility.

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -525,8 +525,6 @@ static void managed_handle_pd_data(struct ustream *s, _unused int bytes_new)
 			else
 				n->valid += now;
 
-			n->dprefix = 0;
-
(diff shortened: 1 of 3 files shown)


dedeckeh:
A more robust and future proof solution would be to cache per interface the relevant routes which can be populated via the netlink interface (see netlink.c and handle_rtm_route in particular) and use this route info to populate the route information options.
Via the netlink interface the valid/preferred lifetimes of the routes can be learned and would allow to get rid of scanning the /proc/net/ipv6_route file; which becomes a bottleneck once the routing table has a lot of entries; for every RA which is sent

dedeckeh:
Closing PR as I won't merge the changes in its current state

PR #160, opened 2020-08-10
Write leases then atomically rename lease file

vii:
### Problem

Consumers of this file need a consistent view.

For example, if /usr/lib/unbound/odhcpd.sh reads a partial file, it may remove many addresses from unbound DNS. Potentially it might also give an incomplete address entry to unbound.

### Solution

Write to a temporary file in the same directory (so hopefully same filesystem). Then use the POSIX rename function to atomically replace the lease file when done.

### Testing

This compiles, but not sure how to build an opkg to test. I have openwrt checked out, but I only build odhcpd-ipv6only for some reason and I need the full build to test on my network. Would love help!

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -306,10 +306,14 @@ void dhcpv6_ia_write_statefile(void)
 	md5_begin(&ctxt.md5);
 
 	if (config.dhcp_statefile) {
+		unsigned tmp_statefile_strlen = strlen(config.dhcp_statefile)+strlen(".tmp")+1;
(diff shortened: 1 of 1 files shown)


dedeckeh:
Did you notice the mentioned issues in combination with unbound ?

dedeckeh:
Before I look into this further I would like to know if you hit the mentioned issues in combination with unbound as in my previous comment

dedeckeh:
Closing as I pushed a slightly re-worked patch (https://git.openwrt.org/?p=project/odhcpd.git;a=commitdiff;h=fb55e80394c51d7502bb278f57520dec15a11355)

PR #162, opened 2020-11-01
dhcpv6: fix size_t fields in syslog format

mikma:
Signed-off-by: Mikael Magnusson <mikma@users.sourceforge.net>


dedeckeh:
Patch pushed; thx

PR #163, opened 2020-11-20
config: add an option to indicate the availability of dns server on interfaces

icpz:
In some cases the IPv6 dns server is not available on the LAN interface, thus we don't want ra/dhcpv6 server to announce the ipv6 address as dns to clients.

Currently, the ra server has 'ra_dns' option to disable the function, but there is no option for dhcpv6. I think a new option 'dnsvalid' to indicate the availability of dns server on specific interface would be better. (I'm not sure if the phrase 'dnsvalid' is suitable for this purpose.)

Thanks!

### DIFF
--- a/README
+++ b/README
@@ -103,6 +103,8 @@ router			list    <local address>		Routers to announce
 							accepts IPv4 only
 dns			list	<local address>		DNS servers to announce
 							accepts IPv4 and IPv6
+dns_service		bool	1			Announce the address of interface as DNS service
+							if the list of dns is empty
 domain			list	<local search domain>	Search domains to announce
 
 leasetime		string	12h			DHCPv4 address leasetime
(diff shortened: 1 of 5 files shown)


dedeckeh:
I would rather call the config option dns_service.
But the PR has formal issues as no SoB is present in the git commit desription (see https://openwrt.org/submitting-patches).
Also you should add the reason why you want this config option to the git commit description

icpz:
Thanks for the reply! I've updated the branch

dedeckeh:
General remark the config option should as well enable/disable sending of DNS server(s) via DHCPv4

icpz:
Updated, thanks!

The new option takes effect for DHCPv4 now.

dedeckeh:
Patch pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=694148a87fa5a0c964286fdaac1fa490a2fb530b); thx

PR #173, opened 2021-07-14
dhcpv6-ia: allow up to 64 bit wide hostid

mikma:
This PR implements up to 64 bit wide hostid for dynamic and static leases.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -35,6 +35,10 @@ struct config config = {.legacy = false, .main_dhcpv4 = false,
 #define START_DEFAULT	100
 #define LIMIT_DEFAULT	150
 
+#define HOSTID_LEN_MIN	12
+#define HOSTID_LEN_MAX	64
+#define HOSTID_LEN_DEFAULT HOSTID_LEN_MIN
+
 #define OAF_DHCPV6	(OAF_DHCPV6_NA | OAF_DHCPV6_PD)
 
 enum {
@@ -61,6 +65,7 @@ enum {
 	IFACE_ATTR_DHCPV6_ASSIGNALL,
 	IFACE_ATTR_DHCPV6_PD,
 	IFACE_ATTR_DHCPV6_NA,
+	IFACE_ATTR_DHCPV6_HOSTID_LEN,
 	IFACE_ATTR_RA_DEFAULT,
 	IFACE_ATTR_RA_MANAGEMENT,
 	IFACE_ATTR_RA_FLAGS,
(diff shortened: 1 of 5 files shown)


src/config.c
```diff
@@ -754,6 +757,17 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 	if ((c = tb[IFACE_ATTR_DHCPV6_NA]))
 		iface->dhcpv6_na = blobmsg_get_bool(c);
 
+	if ((c = tb[IFACE_ATTR_DHCPV6_HOSTID_LEN])) {
+		uint32_t hostid_len = blobmsg_get_u32(c);
+
+		if (hostid_len >= 12 && hostid_len <= 64)
```
dedeckeh: Use a define for the minimum hostid length 12 and the maximum hostid length 64

src/odhcpd.h
```diff
@@ -199,7 +199,7 @@ struct dhcp_assignment {
 	struct odhcpd_ref_ip *fr_ip;
 
 	uint32_t addr;
-	uint32_t assigned;
+	uint64_t assigned;
```
dedeckeh: Changing assigned into an uint64_t makes the code hard to understand and read in case of the prefix delegation logic as assigned is used to assign the subnet id

src/dhcpv6-ia.c
```diff
@@ -688,12 +689,32 @@ static bool assign_na(struct interface *iface, struct dhcp_assignment *a)
...
+		uint64_t try;
+
+		if (iface->dhcpv6_hostid_len > 32) {
+			uint32_t mask_high;
+
+			if (iface->dhcpv6_hostid_len >= 64)
+				mask_high = UINT32_MAX;
+			else
+				mask_high = (1 << (iface->dhcpv6_hostid_len - 32)) - 1;
+			do {
```
dedeckeh, line 740: Align with the existing coding style by inserting an empty line before the do statement

dedeckeh (changes requested):

src/config.c
mikma: Fixed.

src/dhcpv6-ia.c
mikma: Fixed.

src/odhcpd.h
mikma: I have changed `assigned` into an anonymous union with members `uint64_t assigned_host_id` and `uint32_t assigned_subnet_id`.

dedeckeh:
Can you squash all the commits into one commit as they're all related to the same feature

mikma:
I have squashed and rebased on master at git.openwrt.org.

I also added a check against the reserved IPv6 interface identifier ranges listed at http://www.iana.org/assignments/ipv6-interface-ids .

dedeckeh:
This fails to compile for me with the following error

src/dhcpv6-ia.c: In function 'dhcpv6_ia_write_statefile':
src/dhcpv6-ia.c:367:54: error: format '%zx' expects argument of type 'size_t', but argument 10 has type 'uint64_t' {aka 'long long unsigned int'} [-Werror=format=]
      ctxt.buf_idx = snprintf(ctxt.buf, ctxt.buf_len, "# %s %s %x %s%s %"PRId64" %zx %u ",
                                                      ^~~~~~~~~~~~~~~~~~~
src/dhcpv6-ia.c:374:9:
         (ctxt.c->flags & OAF_DHCPV6_NA) ? ctxt.c->assigned_host_id : ctxt.c->assigned_subnet_id,
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/dhcpv6-ia.c:367:83: note: format string is defined here
      ctxt.buf_idx = snprintf(ctxt.buf, ctxt.buf_len, "# %s %s %x %s%s %"PRId64" %zx %u ",
                                                                                 ~~^
                                                                                 %llx

As assigned_hostìd is an uint64_t the best option is to move this to two separate snprintf lines depending on the assignment type

mikma:
I have duplicated the snprintf statements and I use the PRIx64/32-macros in the format string now.

dedeckeh:
Patch pushed to master with a minor modification (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=16667699228101fda936068e3d11c0cf19270e34); thx

PR #176, opened 2021-08-15
dhcpv4: fix uninitialized hostname in some ubus events

mikma:
Ubus events contain broken hostnames unless a valid hostname is provided by the DHCP client. Android 10 and later is one example of devices which follow "Anonymity Profiles for DHCP Clients"
(https://datatracker.ietf.org/doc/html/rfc7844#section-3.7) and don't provide a hostname.

The bug can be reproduced with `ubus subscribe dhcp`.

    dhcpv4: fix uninitialized hostname in some ubus events
    
    The hostname buffer is uninitialized if the client doesn't provide
    DHCPV4_OPT_HOSTNAME. Use hostname from the assignment which is present if
    a static lease contains the hostname or if the client provides one, and
    the hostname is valid. It's also used in the ubus ipv4leases method.

### DIFF
--- a/README
+++ b/README
@@ -95,6 +95,8 @@ dhcpv6_assignall	bool	1			Assign all viable DHCPv6 addresses
 							in statefull mode; if disabled
 							only the DHCPv6 address having the
 							longest preferred lifetime is assigned
+dhcpv6_hostidlength	integer 12			Host ID length of dynamically created leases,
+							allowed values: 12 - 64 (bits).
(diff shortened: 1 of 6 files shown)


dedeckeh:
Patch pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=01b4e6046f10e21809c3f380f2d33bf3fe59698d); thx

PR #179, opened 2021-12-08
odhcpd: Support for Option NTP and SNTP

tekumallaa:
Support for DHCPv6 Option NTP (Option-56) and SNTP (Option-31) and DHCP Option NTP(Option-42) is implemented.
ntp list is supported for IPv4, IPv6 and FQDN.


src/dhcpv4.c
```diff
@@ -849,6 +849,11 @@ void dhcpv4_handle_msg(void *addr, void *data, size_t len,
 		dhcpv4_put(&reply, &cookie, DHCPV4_OPT_DNSSERVER,
 				4 * iface->dhcpv4_dns_cnt, iface->dhcpv4_dns);
 
+	if (iface->dhcpv4_ntp_cnt == 0)
+		dhcpv4_put(&reply, &cookie, DHCPV4_OPT_NTPSERVER, 4, &iface->dhcpv4_local);
```
dedeckeh: This changes the behavior as the NTP option is now sent by default; is there any strong reason to change the behavior as such ?

src/dhcpv6.c
```diff
@@ -362,21 +366,57 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+	uint8_t ntp_buf[IPV6_ADDR_LEN + 4], *ntp_ptr = iface->dhcpv6_ntp;
+	size_t ntp_cnt = iface->dhcpv6_ntp_cnt;
+
+	if ((dns_cnt == 0) || (sntp_cnt == 0)) {
+		odhcpd_get_interface_dns_addr(iface, &addr6);
+		if (dns_cnt == 0) {
+			dns_addr_ptr = &addr6;
+			dns_cnt = 1;
+		}
+		if (sntp_cnt == 0) {
```
dedeckeh: This changes the behavior as the SNTP option is now sent by default; is there any strong reason to change the behavior as such ?

tiagogaspar8:
Any news about this? Would be great if it was supported.

dedeckeh:
> Any news about this? Would be great if it was supported.
I'm waiting for a response on my questions before taking it further

tekumallaa:
> > Any news about this? Would be great if it was supported.
> > I'm waiting for a response on my questions before taking it further
May I Please know the questions. I do not see anything here.

dedeckeh:
> > > Any news about this? Would be great if it was supported.
> > > I'm waiting for a response on my questions before taking it further
> > > May I Please know the questions. I do not see anything here.

Sorry I forgot to push the finish review button

src/dhcpv4.c
tekumallaa: Yes. NTP option is set to local address by default if not explicitly specified. To be inline with DNS behavior.

src/dhcpv6.c
tekumallaa: Yes. SNTP option is set to local address by default if not explicitly specified. To be inline with DNS behavior.

src/dhcpv4.c
dedeckeh: DNS is a requirement to have IP connectivity while for NTP this is not really the case.
If a client requests for the NTP option via ORO and/or PRL we should send the NTP otpion but by default I have my doubts unless there're clients out there which fail if the NTP option is not sent by default

tekumallaa: Adapted to make sure system sends out NTP option only when clients requested

tiagogaspar8: Imagine I wanted to force DHCP to send even when client doesn't ask, can I?

dedeckeh: What's the use case to force send the NTP option if the DHCP(v6) client does not ask for it ?

tiagogaspar8: Fair is fair, sorry to bother, I was reading about it and I don't think it's worth forcing just for the client to ignore.

tekumallaa: Updated patch to send the NTP option only when the clients requests it from ORO.

src/dhcpv6.c
```diff
@@ -376,7 +380,51 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+
+	uint16_t otype, olen;
+	uint8_t *odata;
+	char reqopts[256];
+	size_t reqopts_len = 0;
+
+	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
+		if (otype == DHCPV6_OPT_ORO) {
+                        reqopts_len = olen;
+                        memcpy(reqopts, odata, reqopts_len);
```
dedeckeh: Potentially there's a buffer overflow here if the reqopts option is larget than 256 bytes.
A better option would be to keep a pointer to the ORO option data and remember the ORO option length
If this gets fixed I will merge the PR

tekumallaa: Handled Buffer overflow case for reqopts option.

src/dhcpv6.c
```diff
@@ -376,7 +380,50 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+
+	uint16_t otype, olen;
+	uint8_t *odata;
+	char *reqopts = NULL;
+	size_t reqopts_len = 0;
+
+	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
+		if (otype == DHCPV6_OPT_ORO) {
+			reqopts_len = olen;
+			memcpy(&reqopts, &odata, reqopts_len);
```
dedeckeh: Is this the way how you solve the buffer overflow ?
You're copying reqopts_len bytes from the address of odata to the address of reqopts; this is not the contents of the ORO option which is copied ...
This shows you're not testing your code changes.
All the time I've invested in reviewing this patch and giving advice I could already have implemented it myself

tekumallaa: Apologies. Updated the correct patch for copying the contents of ORO option.

src/dhcpv6.c
```diff
@@ -376,7 +380,55 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+
+	uint16_t otype, olen;
+	uint8_t *odata;
+	char *reqopts = NULL;
+	size_t reqopts_len = 0;
+
+	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
+		if (otype == DHCPV6_OPT_ORO) {
+			reqopts_len = olen;
+			reqopts = malloc(reqopts_len);
```
dedeckeh: Still not correct ...
reqopts is a char pointer which means if you index reqopts you will get the byte value on this entry while the requested ORO option is a 2 byte value (see §21.7 in RFC8415).
Once again you prove you don't test enough your code changes for the different cases (one or more requested options, host byte order vs network byte order) !
Next to that why do you need to malloc extra memory ? Why can't you keep a pointer reference to the data part of the ORO option ?

src/dhcpv6.c
```diff
@@ -376,7 +380,49 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
+		if (otype == DHCPV6_OPT_ORO) {
+			reqopts_len = olen;
+			reqopts = odata;
+		}
+	}
+
+	for(size_t opt = 0; opt < reqopts_len; opt++) {
+		if (iface->dhcpv6_sntp_cnt != 0 &&
+			reqopts[opt] == DHCPV6_OPT_SNTP_SERVERS) {
```
dedeckeh: Last saloon chance before I give up ...
Read §21.7 in RFC8415; you will read that requested options are encoded as **2 byte** values
Now the code compares **one byte** as you index a pointer char which is  **not correct** as you need to compare an **uint16_t**
Once again you're not testing your code changes despite the fact I've already said it a few times

dedeckeh:
Patch pushed (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=860ca900e41c5d0f98cc85e67b39977f6f2cb355); thx

PR #184, opened 2022-06-05
odhcpd: add support for dhcpv6_pd_min parameter

jtkohl:
the dhcpv6_pd_min option clamps the requested prefix delegation to be at
least as big as the option.  This allows a router to manage the size of
each downstream router's prefix delegation length independently from the
delegating interface's prefix length.

Signed-off-by: John Kohl <jtk.git@bostonpog.org>

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -39,6 +39,8 @@ struct config config = {.legacy = false, .main_dhcpv4 = false,
 #define HOSTID_LEN_MAX	64
 #define HOSTID_LEN_DEFAULT HOSTID_LEN_MIN
 
+#define PD_MIN_LEN_MAX (64-2) // must delegate at least 2 bits of prefix
+
 #define OAF_DHCPV6	(OAF_DHCPV6_NA | OAF_DHCPV6_PD)
 
 enum {
@@ -64,6 +66,7 @@ enum {
 	IFACE_ATTR_DHCPV6_RAW,
 	IFACE_ATTR_DHCPV6_ASSIGNALL,
 	IFACE_ATTR_DHCPV6_PD,
+	IFACE_ATTR_DHCPV6_PD_MIN_LEN,
 	IFACE_ATTR_DHCPV6_NA,
 	IFACE_ATTR_DHCPV6_HOSTID_LEN,
 	IFACE_ATTR_RA_DEFAULT,
@@ -116,6 +119,7 @@ static const struct blobmsg_policy iface_attrs[IFACE_ATTR_MAX] = {
 	[IFACE_ATTR_DHCPV6_RAW] = { .name = "dhcpv6_raw", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_DHCPV6_ASSIGNALL] = { .name ="dhcpv6_assignall", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_DHCPV6_PD] = { .name = "dhcpv6_pd", .type = BLOBMSG_TYPE_BOOL },
+	[IFACE_ATTR_DHCPV6_PD_MIN_LEN] = { .name = "dhcpv6_pd_min_len", .type = BLOBMSG_TYPE_INT32 },
 	[IFACE_ATTR_DHCPV6_NA] = { .name = "dhcpv6_na", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_DHCPV6_HOSTID_LEN] = { .name = "dhcpv6_hostidlength", .type = BLOBMSG_TYPE_INT32 },
 	[IFACE_ATTR_PD_MANAGER] = { .name = "pd_manager", .type = BLOBMSG_TYPE_STRING },
@@ -214,6 +218,7 @@ static void set_interface_defaults(struct interface *iface)
 	iface->dhcpv4_end.s_addr = htonl(START_DEFAULT + LIMIT_DEFAULT - 1);
 	iface->dhcpv6_assignall = true;
 	iface->dhcpv6_pd = true;
+	iface->dhcpv6_pd_min_len = 0;
 	iface->dhcpv6_na = true;
 	iface->dhcpv6_hostid_len = HOSTID_LEN_DEFAULT;
 	iface->dns_service = true;
@@ -850,6 +855,15 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 
 	if ((c = tb[IFACE_ATTR_DHCPV6_PD]))
 		iface->dhcpv6_pd = blobmsg_get_bool(c);
+
+	if ((c = tb[IFACE_ATTR_DHCPV6_PD_MIN_LEN])) {
+		uint32_t pd_min_len = blobmsg_get_u32(c);
+		if (pd_min_len != 0 && pd_min_len <= PD_MIN_LEN_MAX)
+			iface->dhcpv6_pd_min_len = pd_min_len;
+		else
+			syslog(LOG_ERR, "Invalid %s value configured for interface '%s'",
+			       iface_attrs[IFACE_ATTR_DHCPV6_PD_MIN_LEN].name, iface->name);
+	}
 
 	if ((c = tb[IFACE_ATTR_DHCPV6_NA]))
 		iface->dhcpv6_na = blobmsg_get_bool(c);
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -1332,6 +1332,31 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
 
 			if (reqlen > 64)
 				reqlen = 64;
+
+			/*
+			 * A requesting router can include a desired prefix length for its
+			 * delegation.  The delegating router (us) is not required to honor
+			 * the hint (RFC3633, section 11.2, we MAY choose to use the
+			 * information in the option; RFC8168, section 3.2 has several SHOULDs
+			 * about desired choices for selecting a prefix to delegate).
+			 *
+			 * We support a policy setting to conserve prefix space, which purposely
+			 * assigns prefixes that might not match the requesting router's hint.
+			 *
+			 * If the minimum prefix length is set in this interface's
+			 * configuration, we use it as a floor for the requested (hinted)
+			 * prefix length.  This allows us to conserve prefix space so that
+			 * any single router can't grab too much of it.  Consider if we have
+			 * an interface with a /56 prefix.  A requesting router could ask for
+			 * a /58 and take 1/4 of our total address space.  But if we set a
+			 * minimum of /60, we can limit each requesting router to get only
+			 * 1/16 of our total address space.
+			 */
+			if (iface->dhcpv6_pd_min_len && reqlen < iface->dhcpv6_pd_min_len) {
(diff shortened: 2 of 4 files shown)


jtkohl:
@dedeckeh

jtkohl:
> @dedeckeh

Hi, any chance to review this/comments/suggestions?

gtxaspec:
waiting on this to be merged too... looking forward to support for this!

jonathanunderwood:
@Ansuel @dedeckeh Any chance we can get this into 23.05?

src/dhcpv6-ia.c
```diff
@@ -1364,7 +1364,17 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
...
+							if (reqlen < iface->dhcpv6_pd_min) {
+								syslog(LOG_INFO, "clamping requested PD from %d to %d",
+								       reqlen, iface->dhcpv6_pd_min);
+								a->length = iface->dhcpv6_pd_min;
+							} else {
+								a->length = reqlen;
+							}
+						} else {
+							a->length = reqlen;
+						}
```
Ansuel: Can you restructure this to something better like

```suggestion
						a->length = reqlen;
						if (is_pd && reqlen < iface->dhcpv6_pd_min) {
							syslog(LOG_INFO, "clamping requested PD from %d to %d",
								    reqlen, iface->dhcpv6_pd_min);
							a->length = iface->dhcpv6_pd_min;
						}
```

src/config.c
```diff
@@ -832,6 +839,15 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
...
 
+	if ((c = tb[IFACE_ATTR_DHCPV6_PD_MIN])) {
+		uint32_t pd_min_len = blobmsg_get_u32(c);
+		if (pd_min_len >= PD_MIN_LEN_MIN && pd_min_len <= PD_MIN_LEN_MAX)
+			iface->dhcpv6_pd_min = pd_min_len;
+		else
+			syslog(LOG_ERR, "Invalid %s value configured for interface '%s'",
+			       iface_attrs[IFACE_ATTR_DHCPV6_PD_MIN].name, iface->name);
+	}
+
```
Ansuel: Considering the value might be not correct, is it right to set it to MIN? I would set it to MAX or add a bool to disable this feature and do nothing (tho using MAX should result in doing nothing)

Ansuel:
Also is this supported or do we have and RFC for this feature?

src/config.c
jtkohl: The default value for iface->dhcpv6_pd_min is set above (221) to PD_MIN_DEFAULT, which is defined as 32.  Reading RFC3769, I think that should change to 48.  (No need to support delegation for fewer than 48 bits of prefix.)

Here, if the configuration specifies a value, and it's not in the range [PD_MIN_LEN_MIN, PD_MIN_LEN_MAX], the specified value gets ignored and the dhcpv6_pd_min is left at PD_MIN_LEN_MIN (will be /48), which is applied to any IA_PD request as the minimum prefix length.  In practice that means the requested IA_PD prefix length is processed as it was before the existence of this option.

If the invalid configuration case sets the minimum prefix length here to PD_MIN_LEN_MAX, then the IA_PD request will be treated as a request for a /64, and nobody downstream will have a useful prefix delegated to them.

I think it's better to convert a bad value into old behavior rather than a useless delegation.  And I don't think we need a boolean to preserve the old behavior if the config file omits the setting.

jtkohl:
> Also is this supported or do we have and RFC for this feature?

I think we end up in the "implementer's option" spaces of RFC3633 and RFC8168, where the behavior can be chosen to meet our desires.

RFC8168 section 3.2 and 3.3 seems on point here. 3.2 starts:

>   [RFC3633] allows a client to include a prefix-length hint in the
   Solicit message to signal its preference to the server.  How the
   prefix-length hint should be handled by the server is unclear.  The
   client might want a different prefix length due to configuration
   changes or it might just want the same prefix again after reboot.
   The server should interpret these cases differently.
>   Many servers are configured to provide only prefixes of specific
   lengths to the client, for example, if the client requested for a /54
   but the server could only provide /30, /48, and /56.  How should
   these servers decide which prefix to give to the client based on the
   prefix-length hint?

RFC3633 section 11.2 describes the delegating router's behavior, in part:

>   If the requesting router includes an IA_PD Prefix option in the IA_PD
   option in its Solicit message, the delegating router MAY choose to
   use the information in that option to select the prefix(es) or prefix
   size to be delegated to the requesting router.
>The delegating router sends an Advertise message to the requesting
   router in the same way as described in section, "Creation and
   transmission of Advertise messages" of RFC 3315.  The delegating
   router MUST include an IA_PD option, identifying any prefix(es) that
   the delegating router will delegate to the requesting router.

I read this to indicate that the delegating router is free to make its own choices about how to respond to an IA_PD, including altering the delegated prefix length to differ from the requested prefix length hint

src/config.c
```diff
@@ -39,6 +39,10 @@ struct config config = {.legacy = false, .main_dhcpv4 = false,
 #define HOSTID_LEN_MAX	64
 #define HOSTID_LEN_DEFAULT HOSTID_LEN_MIN
 
+#define PD_MIN_DEFAULT 32 // will actually be clamped to the interface's prefix length + 1
```
jtkohl: I need to change this to 48 (and likewise the comment in README)

jtkohl: > I need to change this to 48 (and likewise the comment in README)

Sorry, I haven't looked at this code in a year, just re-learning it, and I have to rescind this statement.

I think I see why I chose 32: it's used elsewhere in the code when determining the reqhint (line 1307 or so of dhcpv6-ia.c), and the code will support delegating from a /32, so I think best to leave the minimum prefix at 32.

src/config.c
Ansuel: > The default value for iface->dhcpv6_pd_min is set above (221) to PD_MIN_DEFAULT, which is defined as 32. Reading RFC3769, I think that should change to 48. (No need to support delegation for fewer than 48 bits of prefix.)

Remember to document this in the commit message.



> Here, if the configuration specifies a value, and it's not in the range [PD_MIN_LEN_MIN, PD_MIN_LEN_MAX], the specified value gets ignored and the dhcpv6_pd_min is left at PD_MIN_LEN_MIN (will be /48), which is applied to any IA_PD request as the minimum prefix length. In practice that means the requested IA_PD prefix length is processed as it was before the existence of this option.

We need to be careful with this assumption and check code in dhcpdv6-ia.c
From what I can see reqlen is set to 62 (with is_pd) and is modified again if there is a prefix declared.
So the reqlen value is not always a static values.
This is why I think the default value should be set to the max possible (62?) to not cause side effects on other features.



> If the invalid configuration case sets the minimum prefix length here to PD_MIN_LEN_MAX, then the IA_PD request will be treated as a request for a /64, and nobody downstream will have a useful prefix delegated to them.

Your code sets the length to pd_min only if reqlen is < of pd_min. Setting the default value to MAX will effectively never change the length since it won't ever enter that condition.


> I think it's better to convert a bad value into old behavior rather than a useless delegation. And I don't think we need a boolean to preserve the old behavior if the config file omits the setting.

Yep this is why I'm discussing this, we need to make sure we don't cause regression in other place.

src/config.c
Ansuel: > Sorry, I haven't looked at this code in a year, just re-learning it, and I have to rescind this statement.

it's ok, np. Good that you were still active and interested in this feature. Thanks!

Ansuel:
> I think we end up in the "implementer's option" spaces of RFC3633 and RFC8168, where the behavior can be chosen to meet our desires.

I know it will be a PITA but I would ask you to add all these reference to justify this feature...

Anyway on to the real discussion of this. (thanks for the RFC reference so we know what this is)

This thing is really just another ""hint"" that can be communicated to the server. So what you actually get is quite different.
Because of this I would rename the variables to some thing like dhcpdv6_pd_hint_min and in commit description and in the comments stress that it's really an hint to the remote server and what you actually get can be different than what you requested. (as described in the RFC)

I would move the check condition in dhcpv6-ia at line 1324 and I would operate directly in the reqlen value (it's local and used only for this task)

Also I would drop the MIN_DEFAULT define and just use MAX directly (that from what I can see should be 62 and not 64) looking at the code in dhcpdv6-ia for reqlen.


Also about formal issue. I would ask you to force push and keep everything in one single commit and to drop any merge commit. I you need help with git commands feel free to ask but in short
- git rebase -i HEAD~3 (editor will open read the info on how to use it)
- git add . && git commit --amend (to add stuff to the previous commit)
- git push --force (to force push to this branch and overwrite the HEAD with your local one)

And yes I would ask you to add the info pointed in the RFC about this feature.

Again thanks for helping me in having this feature merged.

src/config.c
jtkohl: > Yep this is why I'm discussing this, we need to make sure we don't cause regression in other place.

I'll review/revise the code, with the goal of:
* if the setting is absent from config file, no behavior change
* if the setting is invalid in the config file, no behavior change (but log about the bad value)
* if the setting is valid, then enforce the minimum prefix length when responding to requests

Ansuel: Hope you also read the other comments with all the suggestion and request. The patch as is is not bad, just need some polish with some case, but I think the suggestion are enough to achieve what we need.

jtkohl:
> I know it will be a PITA but I would ask you to add all these reference to justify this feature...

Not a problem.


> 
> Anyway on to the real discussion of this. (thanks for the RFC reference so we know what this is)
>
> This thing is really just another ""hint"" that can be communicated to the server. So what you actually get is quite different.
> Because of this I would rename the variables to some thing like dhcpdv6_pd_hint_min and in commit description and in the comments stress that it's really an hint to the remote server and what you actually get can be different than what you requested. (as described in the RFC)

Keep in mind that odhcpd in this case is the delegating server.  This setting is not changing the hint that the requesting router sent--that router can send whatever it likes.  What the setting changes is what prefix length is assigned and delegated.  It's setting a policy that the delegating server (us) will only delegate prefix lengths >= the specified value.  I don't think pd_hint_min describes that.  Maybe dhcpdv6_pd_min_len ?

> I would move the check condition in dhcpv6-ia at line 1324 and I would operate directly in the reqlen value (it's local and used only for this task)
> 
> Also I would drop the MIN_DEFAULT define and just use MAX directly (that from what I can see should be 62 and not 64) looking at the code in dhcpdv6-ia for reqlen.

Let me re-read all the related code and then look at this suggestion.

> 
> Also about formal issue. I would ask you to force push and keep everything in one single commit and to drop any merge commit.

Sure, I'll do that at the end, once we have an agreed implementation.
(I come from the philosophy that intermediate development history is important to preserve.  I believe I can do that at least in my own repository on a branch, and do the equivalent of a squash commit to prepare the final pull request content.)

>And yes I would ask you to add the info pointed in the RFC about this feature.

Sure thing, in the code comments and in commit comments.

Ansuel:
@jtkohl Thanks a lot for the fixup, I fixed some formal issue and merged. I will not proceed with bumping the package in main and backport to 23.05

Ansuel:
@jtkohl also would really love if you could add support for this in luci...

gtxaspec:
+1 for luci support! =D

jtkohl:
Not sure when I will have time to get to LuCI support, but yes that's the right next step!

jtkohl:
LuCI PR: https://github.com/openwrt/luci/pull/6443

PR #189, opened 2022-10-14
fix null pointer dereference for INFORM messages

robekl:
The variable `a` was unset when an INFORM message was received, causing a segfault. This checks `a` before dereferencing.

Checked by running locally and sending a DHCP INFORM message and verifying that the process does not segfault.


robekl:
@dedeckeh Please review

dedeckeh:
Patch pushed to master; thx

PR #190, opened 2022-10-21
dhcpv6-ia.c: make tmp lease file hidden

ldir-EDB0:
Use a hidden . prefixed temporary lease file instead of appending '.tmp'.  Dnsmasq is capable of scanning files/directories using inotify to receive file change notifications and updating its view of hostname ip address mapping without being SIGHUPped.  Until dnsmasq v2.88 this mechanism allows additions to hostnames, no deletions.  dnsmasq v2.88 when released will understand how to remove mappings.

Unfortunately without this change dnsmasq sees odhcpd's temporary lease file via inotify and it also sees the change when odhcpd atomically renames the file from '.tmp' to the correct name.

dnsmasq excludes hidden '.' files from it's inotify scans, thus changing odhcpd to use a hidden temporary lease file reduces load and makes sense.

Signed-off-by: Kevin Darbyshire-Bryant <ldir@darbyshire-bryant.me.uk>

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <libgen.h>
 #include <stdbool.h>
 #include <arpa/inet.h>
 #include <sys/timerfd.h>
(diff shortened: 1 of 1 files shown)


dedeckeh:
Thx for the patch; merged as https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=a92c0a73d018cd6453dcf253d9617f97311becab

PR #194, opened 2023-01-16
dhcpv4: improve error when a prefix is too long

rvandegrift:
If a user tries to enable dhcpv4 on an interface with a /29, odhcp won't work. The logs will only contain a message that doesn't help identify the problem. It'd be idea to support any prefix with a valid pool, but at least this would point a confused user in the right direction.

Helps improve https://github.com/openwrt/openwrt/issues/11813

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -37,6 +37,7 @@
 
 #define PACKET_SIZE(start, end) (((uint8_t *)end - (uint8_t *)start) < DHCPV4_MIN_PACKET_SIZE ? \
 				 DHCPV4_MIN_PACKET_SIZE : (uint8_t *)end - (uint8_t *)start)
+#define MAX_PREFIX_LEN 28
 
 static void dhcpv4_netevent_cb(unsigned long event, struct netevent_handler_info *info);
 static int setup_dhcpv4_addresses(struct interface *iface);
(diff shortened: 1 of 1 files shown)


Ansuel:
@rvandegrift Are you still interested in this feature? Can you help me with that? 

I posted some comments to this and I would need you to add a SoB tag with your real name surname and email (hope it's not a problem)

src/dhcpv4.c
```diff
@@ -250,7 +250,7 @@ static int setup_dhcpv4_addresses(struct interface *iface)
 
 	/* Don't allocate IP range for subnets bigger than 28 */
 	if (iface->addr4[0].prefix > 28) {
-		syslog(LOG_WARNING, "Auto allocation of DHCP range fails on %s", iface->name);
+		syslog(LOG_WARNING, "Auto allocation of DHCP range fails on %s (prefix length must be < 29).", iface->name);
```
Ansuel: Since you need to repush this can you by chance improve this by
- Creating a define right under `#define PACKET_SIZE`. Call it something like MAX_SUBNET_SIZE and assign value 28
- Change that 28 to this new define
- Change the syslog to a %d and reference the value  MAX_SUBNET_SIZE  + 1 (or use <= and use MAX_SUBNET_SIZE)

rvandegrift: done - also fixed the wording in the comment (smaller subnets have bigger masks)

rvandegrift:
> @rvandegrift Are you still interested in this feature? Can you help me with that?
> 
> I posted some comments to this and I would need you to add a SoB tag with your real name surname and email (hope it's not a problem)

No problem at all - done.

PR #195, opened 2023-02-10
dhcpv4: detect noarp interfaces

mikma:
Don't add ARP entries to interfaces with IFF_NOARP, it causes problems with for example WireGuard interfaces (which requires this change to be usable with DHCPv4-over-DHCPv6).

Signed-off-by: Mikael Magnusson <mikma@users.sourceforge.net>


dedeckeh:
Patch pushed to master (https://git.openwrt.org/?p=project/odhcpd.git;a=commit;h=dfab0fad85c8b705c483343c99907c229413f7b2); thx

PR #207, opened 2023-11-07
dhcpv6-ia: add some noise to the T1 and T2 periods

ldir-EDB0:
Without this, all the clients get the same timeouts and try to renew addresses at the same time.  Reduce the T1 & T2 by a pseudo random 0 to 127 seconds, thus the renew attempts are likely to be spread out a bit in time.

What do we think?  Sensible?  Stupid?

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -449,14 +449,20 @@ struct nd_opt_capt_portal {
 };
 
 /* IPv6 RA PIOs */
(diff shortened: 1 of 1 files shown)


systemcrash:
Sensible - helps guard against update avalanches.

schuettecarsten:
Nice idea, patch needs to be updated to work with latest odhcpd codebase

systemcrash:
ping @ldir-EDB0 to update this PR

PR #210, opened 2023-12-09
router: skip warning when configured lifetime is 0

rpigott:
The configured lifetime is overridden to be 0 in certain cases where advertising a positive lifetime is not sensible. A log warning is only warranted if the user was not already aware of this condition and erroneously set the lifetime to a positive value.

If the user acknowledges these conditions by manually configuring a 0 lifetime, we should not warn them of the override.

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -638,7 +638,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 
 	if (default_route && valid_prefix) {
 		adv.h.nd_ra_router_lifetime = htons(lifetime < UINT16_MAX ? lifetime : UINT16_MAX);
-	} else {
+	} else if (lifetime) {
 		adv.h.nd_ra_router_lifetime = 0;
 
 		if (default_route) {


src/router.c
```diff
@@ -638,7 +638,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 
 	if (default_route && valid_prefix) {
 		adv.h.nd_ra_router_lifetime = htons(lifetime < UINT16_MAX ? lifetime : UINT16_MAX);
-	} else {
+	} else if (lifetime) {
```
systemcrash, line 641: Doesn't this change the logic flow here? The first if is only entered if both `default_route && valid_prefix` evaluate to true. The second `else` is always entered in all other cases. With this change, it is now only entered if `lifetime` is true.

rpigott: Yes? That is the point of this patch. Can you clarify what is the issue?

I'm trying to remove this superfluous warning about overriding the ra_lifetime to 0. If the lifetime is already 0, there is no need to override it, and certainly no need to warn about overriding it.

systemcrash: What about the 'other' cases? 
Before change: if a and b... else *everything else* end if
After change: if a and b... else if c... end if.

*everything else* is no longer handled. You might be right, I just haven't evaluated what *everything else* might be here.

rpigott: Well the else block only does the override and the warning. If we don't want to do these things (because lifetime is already 0) I reckon there is nothing more to do.

Noltari:
According to my own observations this warning is no longer showing up as it did after recent changes.
Feel free to update and reopen if this still applies.

PR #211, opened 2023-12-22
router: implement RFC 7084 errata 7699

alinnastac:
RFC 7084 L-3 requires router advertisement daemon to send RIO for every prefix delegation that allocates a prefix on the interface. There is one special case though where PIO advertise the on-link prefix route that conflicts with the RIO prefix.

Example:
 - CE-Router receives IA-PD 2001:db8:1234::/64
 - this prefix delegation gets assigned to the lan interface
 - odhcpd advertise on-link PIO with prefix 2001:db8:1234::/64

If original RFC 7084 L-3 would be followed, RA will also contain a RIO with prefix 2001:db8:1234::/64 which will require lan hosts to add the routes
   2001:db8:1234::/64 dev br-lan # the on-link prefix route
   2001:db8:1234::/64 dev br-lan nexthop fe80::1 # RIO route

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -26,6 +26,7 @@
 #include <netlink/attr.h>
 
 #include <arpa/inet.h>
+#include <net/route.h>
 #include <libubox/list.h>
 
 #include "odhcpd.h"
@@ -54,8 +55,15 @@ static struct event_socket rtnl_event = {
 	.sock_bufsize = 133120,
 };
 
-int netlink_init(void)
+static FILE *fp_ipv6_route = NULL;
+
+int netlink_init(bool ipv6)
 {
+	if (ipv6 && !(fp_ipv6_route = fopen("/proc/net/ipv6_route", "r"))) {
+		syslog(LOG_ERR, "fopen(/proc/net/ipv6_route): %m");
+		goto err;
+	}
+
 	rtnl_socket = create_socket(NETLINK_ROUTE);
 	if (!rtnl_socket) {
 		syslog(LOG_ERR, "Unable to open nl socket: %m");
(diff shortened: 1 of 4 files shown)


src/netlink.c
```diff
@@ -212,6 +278,20 @@ static void refresh_iface_addr6(int ifindex)
 			}
 
 			if (change) {
+				/*
+				 * Remove invalid prefixes that were added back to the interface.
```
systemcrash: `				 * Remove invalid prefixes that were added to the interface.`

systemcrash:
@alinnastac I'm running odhcpd patched with this - is there a trivial way to verify it?

Please rebase your changes on current master.

Please revise your commit (and first message) message to (I think it should read like this - did I get that correct?) the following:
```
 router: implement RFC 7084 errata 7699

RFC 7084 L-3 requires router advertisement daemon to send RIO for
every prefix delegation (PD) that allocates a prefix on the interface.
There is one edge case where odhcpd advertises the on-link
PIO prefix route that conflicts with the RIO prefix.

 - PIO = Prefix Information Option
 - RIO = Route Information Option

Example:
 - CE-Router receives IA-PD 2001:db8:1234::/64
 - this prefix gets assigned to the lan interface
 - odhcpd still advertises on-link PIO with prefix 2001:db8:1234::/64

RFC 7084 L-3 mandates that RA also contain a RIO with prefix
2001:db8:1234::/64 which require lan hosts to add the routes

   2001:db8:1234::/64 dev br-lan # the on-link prefix route
   2001:db8:1234::/64 dev br-lan nexthop fe80::1 # RIO route

```

alinnastac:
> is there a trivial way to verify it?

Pretty trivial to check the behavior of odhcpd, but upstream must provide a IA-PD DHCPv6 option with a prefix length set to 64 in order to trigger the change implemented in the patch.

> Please rebase your changes on current master.

I don't think openwrt is interested in this fix, otherwise they would had merged it already years ago.

systemcrash:
I figured it needs upstream input. Just wanted to be sure. 

This repo doesn’t get much love, admittedly - but rebase and make it ready. That I’m testing it is progress.

alinnastac:
You could also modify odhcp6c script to hardcode wan6 prefix delegation length to 64.

src/netlink.c
alinnastac: Hmm... I think "added back" reflects the scenario I cover here.

The scenario is this:
a) prefix X is added to the interface
b) prefix X gets deprecated and Y takes its place
c) prefix Y is deprecated and prefix X is reinstated as preferred prefix

systemcrash: That’s a much clearer explanation. Why not add that?

alinnastac: I thought I did explained clearly by using syntagm "added back".

Besides, it is already explained in the commit description:
> further precaution has been taken in netlink.c to prevent possible
    overlapping between current list of interface IPv6 addresses and the
    list of invalid addresses

systemcrash: It raises more questions than it answers. Better to be explicit. Commit descriptions are not available in the code (without extra steps like `blame`) which is often the first thing someone after us finds.

alinnastac: IMO source code is explaining itself.  This pseudo code explains what it does:
>for a in intf.addresses
>>for b in intf.invalid_addresses
>>>if a include b then remove b from intf.invalid_addresses
    
What comment would you use to explain this?

systemcrash: Your first reply. ( since the code is not immediately obvious )

alinnastac: That is not a suitable comment. I told you the usual scenario because you seem not to understand what this code does, but it is not limited to only this scenario.

This code lines up the invalid address state with the valid address state. RAs mustn't contain same prefix advertised twice, both as valid and invalid prefix at the same time.

systemcrash: Case in point: an elucidating comment would evidently help summarise why this block does what it does, if I do not seem to understand what this code does.

alinnastac: I am willing to change this (since obviously I'm not the best person to explain what code does) but you don't give me a better explanation. Please explain in your words what I explained that the code does.

Further information about odhcpd design: In iface->invalid_addr6 odhcpd keeps track of previous addresses interface had. Elements in this list are added automatically when address dissapears from the interface and will be removed after element get advertised 3 times. New addresses might coincide with entries from invalid_addr6, so precaution must be taken to avoid such case.

systemcrash:
```c
				/*
				 * Remove invalid prefixes that were reinstated.
				 * a) prefix X is added to the interface
				 * b) prefix X gets deprecated and Y takes its place
				 * c) prefix Y is deprecated and prefix X is reinstated as preferred prefix
				 * New addresses might coincide with entries from invalid_addr6,
				 * so precaution must be taken to avoid such case.
				 */
```
?

alinnastac:
@systemcrash I've rebased my commits over the current master branch, but I haven't tested the new commits in any way (not even compiled the code) as I no longer work with openwrt distro.

I will no longer contribute to this project, I'm tired of being ignored.

systemcrash:
Hi @alinnastac - can you bring this up to current please? There have been a number of changes recently requiring rebase.

systemcrash:
@Noltari this can (unfortunately) be closed - the errata were rejected :/

GoetzGoerisch:
Nevertheless 7084bis is in discussion.

https://datatracker.ietf.org/doc/draft-ietf-v6ops-rfc7084bis/

PR #213, opened 2024-02-10
odhcpd: add DNR (RFC 9463) support

Alphix:
Discovery of Network-designated Resolvers (DNR) allows devices on the network
to discover encrypted DNS resolvers, which has so far required either manual
configuration or other approaches (like `systemd-resolved`'s "opportunistic"
mode).

To enable DNR, a new uci parameter has been added, which needs to contain at
the very least, the priority (1-65535, lower = higher priority) and the server
hostname (Authentication Domain Name, ADN, to use the wording of RFC9463):

```
config dhcp 'lan'
        …
        list dnr '100 foobar.example.com'
```

Optionally (and preferably), a comma-separated list of IP addresses and SvcParams can also be specified, like this (line wrapping added):

```
config dhcp 'lan'
        …
        list dnr '100 resolver1.example.com
                    fda7:ab54:69fb::1,fda7:ab54:69fb::2,10.0.0.1
                    alpn=dot port=853'
        list dnr '200 resolver2.example.com
                    fda7:ab54:69fb::2,10.0.1.1,10.0.1.2
                    alpn=dot port=853'
```

Client support is on it's way (e.g. in systemd PR [#30952](https://github.com/systemd/systemd/pull/30952) and in [Windows](https://techcommunity.microsoft.com/t5/networking-blog/introducing-dnr-support-for-windows-insiders/ba-p/3959895)).

~~Draft for now, needs more testing (I've only tested the RA alternative so far) and some small changes to how the options are handled~~.


Alphix:
@rpigott this might interest you, given this PR: https://github.com/systemd/systemd/pull/30952

rpigott:
Yes! Thanks for the ping.

Alphix:
Ok, changed the options handling a bit (it's now `<prio> <ADN> <IPs> <SVC>`) and added some fixes/changes (e.g. for unaligned struct access).

Alphix:
And now I've done more testing, both on x86/qemu and on real hardware (Turris Omnia, running stock OpenWRT 23.05.2). Using a patched version of `rdisc6` (patches sent upstream already) and eyeballing the generated DHCPv4/DHCPv6 packets using `tcpdump`. Looks good to me. Going to convert this to a PR for more review.

One thing I'm less certain about is the configuration format. I've just started looking at `luci`, and these multivalue strings look like they could make life difficult....

systemcrash:
> One thing I'm less certain about is the configuration format. I've just started looking at `luci`, and these multivalue strings look like they could make life difficult....

Hard choices need to be made. A single string is the cleanest copy/paste way to get data in, while chopping fields up makes error checking generally simpler. I recommend staying with long strings.

systemcrash:
@Alphix There is a bit in the RFC about lifetime:
https://www.rfc-editor.org/rfc/rfc9463.html#name-option-format-3

A value of zero means that this ADN MUST no longer be used.

Is it possible (does it make sense to?) to add lifetime as a parameter or flag, for those cases where it can be set to zero and force invalidate a specific resolver?  (i.e. have a config entry with lifetime 0 as 'reminder' that some older resolver must no longer be used). I'm debating a bit with myself here since some other settings co-opt zero to mean 0xffffffff i.e. infinity. 

It might not be immediately convenient to add. My last reading of the source is that lifetime is internally managed.

Also, does this handle the ND(P) DNR?

Alphix:
> Is it possible (does it make sense to?) to add lifetime as a parameter or flag, for those cases where it can be set to zero and force invalidate a specific resolver? (i.e. have a config entry with lifetime 0 as 'reminder' that some older resolver must no longer be used). I'm debating a bit with myself here since some other settings co-opt zero to mean 0xffffffff i.e. infinity.
> 
> It might not be immediately convenient to add. My last reading of the source is that lifetime is internally managed.

@systemcrash It's certainly possible...and you're right that it's currently internally managed (i.e. hardcoded in `src/router.c`, line 807 to use `lifetime` which corresponds to `3 * MaxRtrAdvInterval`, as per [§6.1 of RFC9463](https://www.rfc-editor.org/rfc/rfc9463.html#section-6.1)).

The config string could be extended to also include a lifetime. If we use a fake SVC parameter (e.g. `_lifetime=XYZ`), it would be easy to make it optional (so that the automagic `3 * MaxRtrAdvInterval` can be used when the lifetime is not explicitly configured).

> Also, does this handle the ND(P) DNR?

Yeah, all possibilities (sections 4, 5 and 6 of RFC9463) are covered:
* DHCPv4 (though only in `odhcp`, which is of course not traditionally used as the DHCPv4 server in OpenWRT)
* DHCPv6
* IPv6 RA (NDP)

Alphix:
> Is it possible (does it make sense to?) to add lifetime as a parameter or flag, for those cases where it can be set to zero and force invalidate a specific resolver? (i.e. have a config entry with lifetime 0 as 'reminder' that some older resolver must no longer be used). I'm debating a bit with myself here since some other settings co-opt zero to mean 0xffffffff i.e. infinity.

@systemcrash Sorry for the long delay.

I've rebased the PR and added support for a fake SVC `_lifetime=` parameter, which allows the lifetime to be set manually. I've also tested the result and made sure that the advertised lifetime changes when configured manually...

I hope you can have another look at this PR now...

src/dhcpv6.c
```diff
@@ -393,35 +393,112 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
 	uint8_t *odata;
-	size_t reqopts_len = 0;
+	uint16_t *reqopts = NULL;
+	size_t reqopts_cnt = 0;
 
+	/* FIXME: this should be merged with the second loop further down */
 	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
 		if (otype == DHCPV6_OPT_ORO) {
-			reqopts_len = olen;
+			reqopts_cnt = olen / 2;
```
systemcrash: cognitive load: so the count is always half the length?

Alphix: Yes, it's an array of `uint16_t`s ([source](https://datatracker.ietf.org/doc/html/rfc8415#section-21.7)) and the patch is just moving the divide by 2 up from the deleted `for` loop in line 414 ([this line](https://github.com/openwrt/odhcpd/pull/213/files#diff-4de64a0e599dc48a59ad9d60710b226c7cea18f494814037ff08cb26a2a218ecL414)).

systemcrash: OK. I find a peppering of code comments referencing important bits of the RFC like this very helpful. The code kind of self-justifies, in this way.

Alphix: I wouldn't mind doing that (and I also find replacing magic constants like `/ 2` with `/ sizeof(uint16_t)` to be much clearer)....but I wanted to stick to the existing style as much as possible....I can certainly add more stuff like that in a separate PR later...

systemcrash: Prefer `/ sizeof(uint16_t)` then. Use it. I find that the code here while functional and well written, it was a sort of 'add this bit here and there' to support the 95th percentile of standards needed at home, and some in the enterprise. Who knows, you sometimes find bugs this way. I did.

Alphix: Ok, I've pushed an updated version which adds both a comment referencing the relevant RFC and using `sizeof(uint16_t)`

dedeckeh:
Patch merged, thx

PR #224, opened 2024-10-12
router: add option to include custom routes in RA

### LINKED ISSUE #74: Support adding static router advertisments
Basically I want to be able to replace this `radvd` configuration snippet:

```
interface br-lan
{
        AdvSendAdvert on;
        AdvManagedFlag off;
        AdvOtherConfigFlag off;

        route fc00::/8
        {
                AdvRoutePreference medium;
        };
};
```

It statically advertises a route for the `fc00::/8` address space on all LAN interfaces, nothing else; no assigning addresses, no extra subnet routes, just handing out that single route. This this possible with `odhcpd` right now? Is it hard to implement?

dbnk0:
Introduces new configuration option `ra_staticroute` for adding custom route(s) in route advertisement (RA) messages when no default route is advertised.

This is useful when, for example, end devices are given only ULA IPv6 addresses (e.g. from `fc00:0:100::/48`), but one still needs inter subnet communication to e.g. `fc00:0:200::/48`. Currently, this wouldn't be possible without static route on each end device as the odhcpd server doesn't present itself as default router and no route information is given except for local ULA subnet (`fc00:0:100::/48`).

New configuration option `ra_staticroute` makes it possible to advertise arbitrary routes to the end devices and thus making correct routing possible even when no default route (or public IPv6 prefix) is available.

Option `ra_staticroute` exists under both `odhcpd` section and interfaces' sections. Latter sets up interface-specific routes. Routes configured under `odhcpd` are global and advertised for all interfaces (including those with interface-specific static routes).

Config example:

```
config odhcpd 'odhcpd'
	...
	list ra_staticroute 'fc00:10::/32'
	list ra_staticroute 'fc00:20::/32'

config dhcp 'lan'
	...
	list ra_staticroute 'fd50:60:70::/48'
```

Fixes: https://github.com/openwrt/odhcpd/issues/74 and possibly helps with https://github.com/openwrt/odhcpd/issues/152

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -31,7 +31,8 @@ struct vlist_tree leases = VLIST_TREE_INIT(leases, lease_cmp, lease_update, true
 AVL_TREE(interfaces, avl_strcmp, false, NULL);
 struct config config = {.legacy = false, .main_dhcpv4 = false,
 			.dhcp_cb = NULL, .dhcp_statefile = NULL, .dhcp_hostsfile = NULL,
-			.log_level = LOG_WARNING};
+			.log_level = LOG_WARNING, .ra_static_routes = NULL,
+			.ra_static_routes_cnt = 0};
 
 #define START_DEFAULT	100
 #define LIMIT_DEFAULT	150
@@ -103,6 +104,7 @@ enum {
 	IFACE_ATTR_RA_HOPLIMIT,
 	IFACE_ATTR_RA_MTU,
 	IFACE_ATTR_RA_DNS,
+	IFACE_ATTR_RA_STATICROUTE,
 	IFACE_ATTR_RA_PREF64,
 	IFACE_ATTR_PD_MANAGER,
 	IFACE_ATTR_PD_CER,
@@ -159,6 +161,7 @@ static const struct blobmsg_policy iface_attrs[IFACE_ATTR_MAX] = {
 	[IFACE_ATTR_RA_HOPLIMIT] = { .name = "ra_hoplimit", .type = BLOBMSG_TYPE_INT32 },
 	[IFACE_ATTR_RA_MTU] = { .name = "ra_mtu", .type = BLOBMSG_TYPE_INT32 },
 	[IFACE_ATTR_RA_DNS] = { .name = "ra_dns", .type = BLOBMSG_TYPE_BOOL },
+	[IFACE_ATTR_RA_STATICROUTE] = { .name = "ra_staticroute", .type = BLOBMSG_TYPE_ARRAY },
 	[IFACE_ATTR_RA_PREF64] = { .name = "ra_pref64", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_NDPROXY_ROUTING] = { .name = "ndproxy_routing", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_NDPROXY_SLAVE] = { .name = "ndproxy_slave", .type = BLOBMSG_TYPE_BOOL },
@@ -200,6 +203,7 @@ enum {
 	ODHCPD_ATTR_LEASETRIGGER,
 	ODHCPD_ATTR_LOGLEVEL,
 	ODHCPD_ATTR_HOSTSFILE,
+	ODHCPD_ATTR_RA_STATICROUTE,
 	ODHCPD_ATTR_MAX
 };
 
@@ -210,6 +214,7 @@ static const struct blobmsg_policy odhcpd_attrs[ODHCPD_ATTR_MAX] = {
 	[ODHCPD_ATTR_LEASETRIGGER] = { .name = "leasetrigger", .type = BLOBMSG_TYPE_STRING },
 	[ODHCPD_ATTR_LOGLEVEL] = { .name = "loglevel", .type = BLOBMSG_TYPE_INT32 },
 	[ODHCPD_ATTR_HOSTSFILE] = { .name = "hostsfile", .type = BLOBMSG_TYPE_STRING },
+	[ODHCPD_ATTR_RA_STATICROUTE] = { .name = "ra_staticroute", .type = BLOBMSG_TYPE_ARRAY },
 };
 
 const struct uci_blob_param_list odhcpd_attr_list = {
@@ -285,6 +290,7 @@ static void clean_interface(struct interface *iface)
 	free(iface->dhcpv4_router);
 	free(iface->dhcpv4_dns);
 	free(iface->dhcpv6_raw);
+	free(iface->ra_static_routes);
 	free(iface->filter_class);
 	free(iface->dhcpv4_ntp);
 	free(iface->dhcpv6_ntp);
@@ -364,6 +370,135 @@ static int parse_ra_flags(uint8_t *flags, struct blob_attr *attr)
 	return 0;
 }
 
+/*
+ * Parses static route from configuration attribute to `route`
+ *
+ * Returns: 0 on success, -1 on failure
+ */
+static int parse_ra_staticroute(struct blob_attr *attr,
+				struct odhcpd_ip6prefix *route)
+{
+	if (blobmsg_type(attr) != BLOBMSG_TYPE_STRING ||
+	    !blobmsg_check_attr(attr, false)) {
+		return -1;
+	}
+
+	const char *str = blobmsg_get_string(attr);
+
+	/* Parse the address and prefix length */
+	if (odhcpd_parse_addr6_prefix(str, &(route->addr), &(route->len)) < 0) {
+		return -1;
+	}
+
+	/*
+	 * Validate prefix length
+	 * Default route (::/0) is managed elsewhere.
+	 */
+	if (route->len == 0 || route->len > 128) {
+		return -1;
+	}
+
+	/* Get network address (zero out host bits) */
+	if (!odhcpd_addr6_to_network(&(route->addr), route->len, &(route->addr))) {
+		/* Should never fail */
+		return -1;
+	}
+
+	/*
+	 * Validate address
+	 * Zero-address (::) specifies default route and is managed elsewhere,
+	 * loopback and link-local addresses are not routable.
+	 */
+	if (IN6_IS_ADDR_UNSPECIFIED(&(route->addr)) ||
+	    IN6_IS_ADDR_LOOPBACK(&(route->addr)) ||
+	    IN6_IS_ADDR_LINKLOCAL(&(route->addr))) {
+		return -1;
+	}
+
+	return 0;
+}
+
+/*
+ * Appends route to the list of static routes
+ *
+ * Returns: 0 on success, -1 on failure (`routes` and `routes_cnt` are
+ * unmodified in that case).
+ */
+static int append_ra_staticroute(struct odhcpd_ip6prefix *new_route,
+				 struct odhcpd_ip6prefix **routes,
+				 size_t *routes_cnt)
+{
+	/* Attempt resize and reallocation */
+	struct odhcpd_ip6prefix *tmp = realloc(*routes, (*routes_cnt + 1) *
+					       sizeof(**routes));
+	if (!tmp) {
+		/* Allocation failed */
+		return -1;
+	}
+
+	*routes = tmp;
+	(*routes)[*routes_cnt] = *new_route;
+	++(*routes_cnt);
+
+	return 0;
+}
+
+/*
+ * Processes static route attribute from configuration
+ *
+ * Attribute value is parsed and added to the list of static routes.
+ * Processing success or failure is logged.
+ *
+ * Returns: 0 on success, -1 on recoverable error, -2 on fatal error
+ * (memory allocation failure)
+ */
+static int process_ra_staticroute(struct blob_attr *attr,
+	struct odhcpd_ip6prefix **routes, size_t *routes_cnt,
+	const char *attr_name, const char *ifname)
+{
+	struct odhcpd_ip6prefix new_route = {};
+
+	/* Parse the route */
+	if (parse_ra_staticroute(attr, &new_route) != 0) {
+		/* Invalid route */
+		char default_str[] = "?";
+		char *str = default_str;
+		if (blobmsg_type(attr) == BLOBMSG_TYPE_STRING &&
+		    blobmsg_check_attr(attr, false)) {
+			str = blobmsg_get_string(attr);
+		}
+
+		if (ifname) {
+			syslog(LOG_WARNING, "Invalid %s value (%s) for "
+			       "interface '%s', ignoring this route", attr_name,
+			       str, ifname);
+		} else {
+			syslog(LOG_WARNING, "Invalid %s value (%s), ignoring "
+			       "this route", attr_name, str);
+		}
+		return -1;
+	}
+
+	/* Append the route to the list */
+	if (append_ra_staticroute(&new_route, routes, routes_cnt) != 0) {
+		syslog(LOG_ERR, "Appending route failed for %s",
+		       attr_name);
+		return -2;
+	}
+
+	/* Print the route */
+	char addr_str[INET6_ADDRSTRLEN];
+	inet_ntop(AF_INET6, &new_route.addr, addr_str, sizeof(addr_str));
+	if (ifname) {
+		syslog(LOG_INFO, "Adding static route %s/%u for interface '%s'",
+		       addr_str, new_route.len, ifname);
+	} else {
+		syslog(LOG_INFO, "Adding static route %s/%u for all interfaces",
+		       addr_str, new_route.len);
+	}
+	return 0;
+}
+
 static void set_config(struct uci_section *s)
 {
 	struct blob_attr *tb[ODHCPD_ATTR_MAX], *c;
@@ -399,6 +534,31 @@ static void set_config(struct uci_section *s)
 		if (config.log_level != log_level) {
 			config.log_level = log_level;
 			setlogmask(LOG_UPTO(config.log_level));
+		}
+	}
+
+	if ((c = tb[ODHCPD_ATTR_RA_STATICROUTE])) {
+		free(config.ra_static_routes);
+		config.ra_static_routes = NULL;
+		config.ra_static_routes_cnt = 0;
+
+		struct blob_attr *cur;
+		unsigned rem;
+
+		syslog(LOG_INFO, "Processing global static routes");
+
+		blobmsg_for_each_attr(cur, c, rem) {
+			int ec = process_ra_staticroute(cur,
+				&config.ra_static_routes,
+				&config.ra_static_routes_cnt,
+				odhcpd_attrs[ODHCPD_ATTR_RA_STATICROUTE].name,
+				NULL);
+			if (ec == -1) {
+				continue;
+			} else if (ec == -2) {
+				/* Fatal error */
+				break;
+			}
 		}
 	}
 }
@@ -1316,6 +1476,28 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 
 	if ((c = tb[IFACE_ATTR_RA_DNS]))
 		iface->ra_dns = blobmsg_get_bool(c);
+
+	if ((c = tb[IFACE_ATTR_RA_STATICROUTE])) {
+		struct blob_attr *cur;
+		unsigned rem;
+
+		syslog(LOG_INFO, "Processing static routes for interface '%s'",
+		       iface->name);
+
+		blobmsg_for_each_attr(cur, c, rem) {
+			int ec = process_ra_staticroute(cur,
+				&(iface->ra_static_routes),
+				&(iface->ra_static_routes_cnt),
+				iface_attrs[IFACE_ATTR_RA_STATICROUTE].name,
+				iface->name);
+			if (ec == -1) {
+				continue;
+			} else if (ec == -2) {
+				/* Fatal error */
+				break;
+			}
+		}
+	}
 
 	if ((c = tb[IFACE_ATTR_DNR])) {
 		struct blob_attr *cur;
--- a/src/config.c
+++ b/src/config.c
@@ -1853,6 +1853,8 @@ void odhcpd_reload(void)
 	if (!uci)
 		return;
 
+	syslog(LOG_NOTICE, "Reloading...");
+
 	vlist_update(&leases);
 	avl_for_each_element(&interfaces, i, avl)
 		clean_interface(i);
--- a/README
+++ b/README
@@ -67,6 +67,8 @@ leasefile	string				DHCP/v6 lease/hostfile
 leasetrigger	string				Lease trigger script
 hostsfile	string				DHCP/v6 hostfile
 loglevel	integer 6			Syslog level priority (0-7)
+ra_staticroute	list	string			Announced static routes for all interfaces
+		[IPv6 prefix]			(used if RA doesn't include default router info already)
 
 
 Sections of type dhcp (configure DHCP / DHCPv6 / RA / NDP service)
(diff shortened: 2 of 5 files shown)


systemcrash:
I think this is a good option to have. 

@DavidB137 please fix the conflicts.

dbnk0:
Sorry, I messed up quite a bit when resolving the conflicts. At first, I used spaces instead of tabs for indentation and then tried to fix it (unsuccessfully). So I instead synced my fork with upstream and reapplied the changes which should also make the history cleaner as there's no merge commit. Once again, sorry for the confusion.

@systemcrash Hopefully you can merge now.

systemcrash:
I wonder whether it'd be better to call them `list ra_customroute` or `list ra_staticroute` instead? Add is a verb which might suggest something else than what they are for. What do you think @dbnk0 ?

dbnk0:
I completely agree with you, `list ra_staticroute` seems like the best option to me.

I renamed everything, did some refactoring, removed unnecessary dynamic memory allocation and improved error handling.
In case you want to review rename and everything else individually, these are the commits: f1ee9c6, 4fec3c8, 8b8b1cc, af20724.

I squashed everything together now.

@systemcrash Let me know if you want to change anything else.

systemcrash:
This still looks great - more readable now. Could you update your initial post example please?

dbnk0:
Updated

dbnk0:
@dedeckeh, @Ansuel Could you take a look, please?

src/config.c
```diff
@@ -364,6 +370,63 @@ static int parse_ra_flags(uint8_t *flags, struct blob_attr *attr)
 	return 0;
 }
 
+/* Returns 0 on success, -1 on invalid value, -2 on memory error
+(`routes` is unmodified in that case) */
+static int parse_append_ra_staticroute(struct blob_attr *attr,
```
systemcrash: Perhaps we could split this function up? It does two things. We might not know that the parse was in error or the append operation was in error. One function, one task. ( Parse could call append, but its implied naming `parse_ra_staticroute` would belie its purpose ). Suggested: parse returns 0 if all went well, and the calling function body calls an append function.

dbnk0: Fixed, see [this diff](https://github.com/openwrt/odhcpd/compare/6b0671fec3706a52b1eec4727c4733bdd9539513..105a8be850094fe763ca1797a4a7fd7f4aac9db2).

I split `parse_append_ra_staticroute` into `parse_ra_staticroute` and `append_ra_staticroute`, and also added a new function `process_ra_staticroute` to orchestrate them and deduplicate code.

`process_ra_staticroute` also provides logging, both for successfully (`INFO` level) and unsuccessfully (`WARNING` level) processed values.

dbnk0:
As mentioned in the review comment above, I added **more logging** for static route processing.
The objective is to provide user a better feedback on which entries have (not) been pushed to the list of advertised routes.

To reduce confusion of repeated config parsing, I decided to add a `NOTICE` level message "Reloading..." into `odhcpd_reload()` function. That forms a separate commit (76021b313934679be4327c2f24e488442a394b2b).

---

Example of possible syslog output from odhcpd on OpenWrt router during process startup:

```
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Processing global static routes
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Adding static route fe90::/16 for all interfaces
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (invalidtext123), ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (invalidtext245), ignoring this route
Thu Apr 10 19:20:58 2025 daemon.notice odhcpd[20873]: Reloading...
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Processing global static routes
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Adding static route fe90::/16 for all interfaces
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (invalidtext123), ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (invalidtext245), ignoring this route
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Processing static routes for interface 'lan'
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Adding static route fd20:1:2::/48 for interface 'lan'
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (invalidtext) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (::/0) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (::/1) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (2000::/129) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (::1/128) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.warn odhcpd[20873]: Invalid ra_staticroute value (fe80::110/127) for interface 'lan', ignoring this route
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Processing static routes for interface 'lan_iot'
Thu Apr 10 19:20:58 2025 daemon.info odhcpd[20873]: Adding static route fd40:1:2::/48 for interface 'lan_iot'
Thu Apr 10 19:20:59 2025 daemon.warn odhcpd[20873]: No default route present, overriding ra_lifetime to 0!
Thu Apr 10 19:20:59 2025 daemon.notice odhcpd[20873]: Sending a RA on lan
Thu Apr 10 19:20:59 2025 daemon.warn odhcpd[20873]: No default route present, overriding ra_lifetime to 0!
Thu Apr 10 19:20:59 2025 daemon.notice odhcpd[20873]: Sending a RA on lan_guest
Thu Apr 10 19:20:59 2025 daemon.warn odhcpd[20873]: No default route present, overriding ra_lifetime to 0!
Thu Apr 10 19:20:59 2025 daemon.notice odhcpd[20873]: Sending a RA on lan_infra
Thu Apr 10 19:20:59 2025 daemon.warn odhcpd[20873]: No default route present, overriding ra_lifetime to 0!
Thu Apr 10 19:20:59 2025 daemon.notice odhcpd[20873]: Sending a RA on lan_iot
```

systemcrash:
OK - memory (re)allocation and cleanup looks good. Separate functions. I think everything looks sufficient here. You've added `append_router_advert_routes` - so I assume the cleanup for those is already present at config(re)load?

dbnk0:
Dynamic memory allocation is performed at two places:

1. [`append_router_advert_routes()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/router.c#L463): Called from [`send_router_advert()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/router.c#L867-L891) on local variable `routes` which is freed at the [end of the function body](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/router.c#L974).
2. [`append_ra_staticroute()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L432): Called via [`process_ra_staticroute()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L483) from:
   - [`set_config()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L551): `config.ra_static_routes` is [freed by `set_config()` itself](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L541) just like everything else in `config`.
   - [`config_parse_interface()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L1486): `iface->ra_static_routes` is [freed by `clean_interface()`](https://github.com/openwrt/odhcpd/blob/76021b313934679be4327c2f24e488442a394b2b/src/config.c#L293) function just like everything else in that structure.

dbnk0:
Ad small patch above: I extended error code validation for `process_ra_staticroute()`. The functionality didn't change at all (`blobmsg_for_each_attr` would have `continue`d in either case), new conditions are there just to make things explicit and visible.

dwardor:
Hello,

I got here looking for a solution to advertize [non default] static routes on Openwrt 24.10.0 

I was doing this using radvd on ddwrt and searching for a way to do it on openwrt since migrating.

Here is how my network works:
My ISP :
* does IPV4 CGNAT & IPV6
* does not do prefix delegation just simple dhcpv6 
* only gives me x:y:z:1::/64 (ISP's router's lan side) and defines x:y:z:2::/64 as dmz and provides an associated routing table entry

ISP's router is x:y:z:1::1 and i configure route to x:y:z:2::0/64 (dmz) via x:y:z:1::2 (or the wan fe80:: link local ipv6 of OPENWRT router)
OPENWRT router is x:y:z:1::2 on wan side (dhcpv6 client) and x:y:z:2::1/64 on lan side (static) and runs dhcpv6 (range 111-210) &SLAAC + a wireguard server.

So my OPENWRT router only has a /64 at disposal which meens that to provide my wireguard clients with a public ipv6 (to force all their traffic through the tunnel) I use the x:y:z:2::1:0/112 sub network (which should not overlap the dhcpv6 handouts)

And I therefore need openwrt to advertise that it is the gateway to both :
* ::/0 --> the default route
* x:y:z:2::1:0/112 --> my wireguard subnetwork

My undestanding is that your proposal could make this possible but for now decides not to as the default route is advertized.

Using radvd on ddwrt (for the past 4 years)  all i would do is launch radvd with the following configuration (while dnsmasq was doing the default route ra) :
interface br0 {
   AdvSendAdvert on;
   AdvManagedFlag on;
   AdvOtherConfigFlag on;
   AdvReachableTime 1800;
   MaxRtrAdvInterval 180;
   route x:y:z:2::1:0/112 { };
};

Looking forward to your consideration of my use case and request, (or, if an alternative [which I missed] without ipv6 masquerading exists, your suggestions). 

Regards

P.S. I tried to use uradvd but it does not accept prefix lengths greater than 64 :-(

Edited to fix an error :-(

systemcrash:
@dbnk0 you can add my Tested-by signature. 

`Tested-By: Paul Donald <newtwen+github@gmail.com>`

I've compiled and run-tested this patch on snapshot. Works great ❤️ 

Make sure PR is editable by maintainers. 

ping @robimarko 




@dwardor I tested it with /96 - works too.

```
ICMPv6 Option (Route Information : Medium fc00:10::/96)
    Type: Route Information (24)
    Length: 3 (24 bytes)
    Prefix Length: 96
    Flag: 0x00, Route Preference: Medium
    Route Lifetime: 1800 (30 minutes)
    Prefix: fc00:10::
```

dwardor:
Learned quilt and applied the 9 patches to get ra_staticroute

It's a no go for me... 

I played around and could only get the ra_staticroutes advertized from lan only if both these condtions were met:
*  I'm not advertizing a public prefix (i.e. static lan definition does not contain a public ipv6)
*  option ra_default '0'

Hence my request to allow this in more cases than what is planned for as of now

systemcrash:
@dwardor In what case would you want to advertise a static route when you have a public prefix? Just... in addition to public prefixes?

dwardor:
My x:y:z:2::/64 clients need a route via x:y:z:2::1 to x:y:z:2::1:0/112 where my wireguard peers are.

Without it when they try to reach x:y:z:2::1:0/112 they follow the "normal" x:y:z:2::/64 route... which is just one hop out their "network dev" and therefore doesn't work...

_Edited for typo_

systemcrash:
I see. That could be useful. Perhaps if we ask nicely, @dbnk0 could modify this patch to accommodate this use case - it might be out of scope, however. 

This patch code currently has:
```c
	if (!default_route || !valid_prefix) {
```

So the behaviour you've observed is thus expected.

dwardor:
@systemcrash I can confirm I'm getting what I need :1st_place_medal: with this simple test patch:
```
--- a/src/router.c
+++ b/src/router.c
@@ -864,7 +864,7 @@ static int send_router_advert(struct int
         */
 
        /* If router is not default */
-       if (!default_route || !valid_prefix) {
+       //if (!default_route || !valid_prefix) {
                /* Global static RA routes */
                if (config.ra_static_routes_cnt > 0) {
                        int ec = append_router_advert_routes(&routes, &routes_cnt,
@@ -888,7 +888,7 @@ static int send_router_advert(struct int
                                       "static RA routes on %s", iface->name);
                        }
                }
-       }
+       //}
 
        for (ssize_t i = 0; i < valid_addr_cnt; ++i) {
                struct odhcpd_ipaddr *addr = &addrs[i];
```

@dbnk0, I'm not an ipv6 expert, so I don't understand what the arguments are for limiting to `if (!default_route || !valid_prefix) {`. If there aren't any could the limitation be removed ?

Regards

systemcrash:
Good to hear. I think those conditions could be combined with a flag (yay, more flags). E.g.
```c
	if ((!default_route || !valid_prefix) && !ignore_custom_routes) {
```

But whether or not it should be like this might require an exhaustive reading of the IPv6 RFC. IIRC there's no limitation regarding static routes - just mandated behaviours when a PD is assigned.

dbnk0:
Paul, I added your `Tested-By` signature.

@dwardor If I understand correctly, you are basically allocating `/64` prefix to one network (LAN) while also taking `/112` prefix out of that allocation and reallocating it to another network.

Hosts on your LAN of course mark `/64` as on-link route, and thus don't route the packets destined to `/112` via the default gateway. You need to force them to add a more specific routing table entry for `/112`, so that routing works as you need to.

Are you sure you want to do that? 😄

In any case, I can add `ra_forcestaticroutes` flag to accomodate your use case.

---

However, I also came accross [RFC 4191](https://datatracker.ietf.org/doc/html/rfc4191) which, despite addressing a different issue, can fit this use case as well. I briefly read through the whole document and it seems we should *always* allow custom route advertisements even if the default route is advertised, because custom combinations of default and non-default routes may be necessary in a multi-router situations. Moreover, we should make preference configurable for each route.

With that, what do you think is a better solution for configuration syntax? @systemcrash

1. Optionally specifying preference (with `option ra_preference` value as a default) after the route itself like:
   ```ini
   list ra_staticroute 'fc00:10::/32'      # uses `ra_preference`
   list ra_staticroute 'fc00:20::/32 high'
   ```
2. Adding a new section type `ra_route` in `/etc/config/dhcp`:
   ```ini
   config ra_route
       list interface 'lan'
       option target 'fc00:10::/32'
       # no `option preference` specified: uses `ra_preference`

   config ra_route
       list interface 'lan'
       option target 'fc00:20::/32'
       option preference 'high'
   ```

The first one will be a lot more straightforward to implement and easier to use.
The second solution is more difficult, but future-proof, because it can be easily extended with more options (route lifetime?).

I generally prefer future-proof solutions, but will need to take a look into the codebase and see how much hassle is it going to incur.

dwardor:
> @dwardor If I understand correctly, you are basically allocating `/64` prefix to one network (LAN) while also taking `/112` prefix out of that allocation and reallocating it to another network.
> 
> Hosts on your LAN of course mark `/64` as on-link route, and thus don't route the packets destined to `/112` via the default gateway. You need to force them to add a more specific routing table entry for `/112`, so that routing works as you need to.
> 
> Are you sure you want to do that? 😄

My ISP only gives me one /64 so that's the only solution I have identified to provide public ipv6 to the wireguard peers. 

The only other solution not requiring a static route, but without public ipv6 for the peers, would be using a dedicated /64 ULA on the wireguard tunnel, and putting the wireguard interface in its own firewall zone, and masquerading on wan interface what comes from wireguard interface) [i.e. using an ipv4 strategy...]

You hav'ent asked me... but... although 1. serves my needs and could be available "fast", 2. seems the better solution.

dbnk0:
Yeah, yeah, sorry. I asked Paul, because he did a review and contributed to odhcpd before.

NAT seems like a more reasonable solution than allocating the same address space twice.
Maybe another question, do you really need a public IP for Wireguard clients? Wouldn't be private subnet sufficient? If you only need a VPN to access local network remotely, isn't access to IPv4 internet + routing IPv6 within local network enough? This way, you don't even need IPv6 NAT.

systemcrash:
RFC 4191 is problematic. I *actually* chased up on *exactly* that RFC to the stewards of the header files for Linux ( to  Florian Weimer @ RedHat, in 2023 ) regarding exactly the flags outlined in that RFC:

https://sourceware.org/pipermail/libc-alpha/2014-February/048714.html
https://sourceware.org/pipermail/libc-alpha/2014-February/048854.html

```
* Paul D.:

> What was the outcome?
>
> I find myself in a position where it's time to rely on some ICMP types
> and flags which have been documented in RFCs for more than a decade,
> but we're using code which hasn't budged since about 20 years ago...

Some of these values are from RFC 4191, but it did not properly update
RFC 2461, so the updates were lost in RFC 4861.  So maybe this is not
the direction in which neighbor discovery evolved after all?

I haven't looked at the other definitions.

Thanks,
Florian

```

So it's possible that even if we implement that RFC here, other platforms won't understand it. So it looks like we must go rely on the 'default' `ra_preference`.

I think option 1 is most likely to survive reviews and merge.

systemcrash:
> > @dwardor If I understand correctly, you are basically allocating `/64` prefix to one network (LAN) while also taking `/112` prefix out of that allocation and reallocating it to another network.
> > Hosts on your LAN of course mark `/64` as on-link route, and thus don't route the packets destined to `/112` via the default gateway. You need to force them to add a more specific routing table entry for `/112`, so that routing works as you need to.
> > Are you sure you want to do that? 😄
> 
> My ISP only gives me one /64 so that's the only solution I have identified to provide public ipv6 to the wireguard peers.
> 
> The only other solution not requiring a static route, but without public ipv6 for the peers, would be using a dedicated /64 ULA on the wireguard tunnel, and putting the wireguard interface in its own firewall zone, and masquerading on wan interface what comes from wireguard interface) [i.e. using an ipv4 strategy...]
> 
> You hav'ent asked me... but... although 1. serves my needs and could be available "fast", 2. seems the better solution.

I find it pointless to assign a public prefix to any of the wireguard clients, especially if it's a prefix the ISP can reassign - I just generate some fake ULA subnet within the ULA space, and assign each one a ULA based address like `fd00:a::1/128`, and have in `/etc/config/network`

```
config rule6
	option src 'fd00::/8'
	option lookup 'main'
```

This way when my public prefix changes, I never have to update my config. It just works. It also has the benefit that everyone in the net can always reach the wg clients. ULA were originally designed to be global, so as long as remote nets have a static route for your local ULA, it works. Then you have IPv4 :)

systemcrash:
> In any case, I can add `ra_forcestaticroutes` flag to accomodate your use case.

Simplest is best for now.

dwardor:
> I find it pointless to assign a public prefix to any of the wireguard clients, especially if it's a prefix the ISP can reassign - I just generate some fake ULA subnet within the ULA space, and assign each one a ULA based address like `fd00:a::1/128`, and have in `/etc/config/network`
> 
> ```
> config rule6
> 	option src 'fd00::/8'
> 	option lookup 'main'
> ```
> 
> This way when my public prefix changes, I never have to update my config. It just works. It also has the benefit that everyone in the net can always reach the wg clients. ULA were originally designed to be global, so as long as remote nets have a static route for your local ULA, it works. Then you have IPv4 :)

I have a fixed prefix :-) so the simplicity of the advertized static route solution (with radvd at the time) was very tempting as it provided full ipv4 and ipv6 !

dbnk0:
So to summarize.

There was [RFC 2461](https://datatracker.ietf.org/doc/html/rfc2461) defining neighbor discovery in IPv6. It was extended by [RFC 4191](https://datatracker.ietf.org/doc/html/rfc4191) to make route advertisements and specifying preference possible. RFC 2461 was then replaced by [RFC 4861](https://datatracker.ietf.org/doc/html/rfc4861). However, it doesn't take fully into account RFC 4191 (though RFC 4191 is mentioned [here](https://datatracker.ietf.org/doc/html/rfc4861#ref-RTSEL)).

The issue you are pointing out is that the RFC 4191 support isn't that universally widespread.
Am I correct, @systemcrash?

But RFC 4191 AFAIK doesn't break anything from RFC 4861. Legacy clients may ignore "Route information" options in RAs.

I ran some tests locally. I've tried running two routers -- one modified odhcpd and one radvd -- both advertising the route `fd99::/48` to the same network, but one of them with `LOW`, the other one with `HIGH` preference, and under different lifetimes. I checked the resulting routing tables on Ubuntu 20.04, Windows 11 and Android 15 devices. Ubuntu and Windows correctly picked up all changes (discovery of both routers, disappearance of one of them, start of router with higher preference, etc.) and routed packets to the right router. Android device seemed to stick to one of routers and ignore higher preference altoghether. It also ignored advertisements for `/16` prefix for some reason. So I would assume Linux and Windows-based clients implement and support RFC 4191, while Android mobile devices have some issues (at least based on this quick test).

In any case, I don't see a problem with properly implementing RFC 4191 and making a route-specific preference and lifetime possible. radvd [has these options](https://github.com/radvd-project/radvd/blob/b9c1ce7640e49a2d5feb8c58f30e5af622eca484/test/test1.conf#L30-L40).

> > In any case, I can add `ra_forcestaticroutes` flag to accomodate your use case.
> 
> Simplest is best for now.

`ra_forcestaticroutes` flag was meant as a "workaround" while keeping the original "static routes only if the router is not default" philosophy, but that shouldn't be the case according to RFC 4191.

I would much rather properly implement static route advertisements now than have it reimplemented a couple of years later with migrations and/or backwards-compatible changes.

systemcrash:
> So to summarize.
> 
> There was [RFC 2461](https://datatracker.ietf.org/doc/html/rfc2461) defining neighbor discovery in IPv6. It was extended by [RFC 4191](https://datatracker.ietf.org/doc/html/rfc4191) to make route advertisements and specifying preference possible. RFC 2461 was then replaced by [RFC 4861](https://datatracker.ietf.org/doc/html/rfc4861). However, it doesn't take fully into account RFC 4191 (though RFC 4191 is mentioned [here](https://datatracker.ietf.org/doc/html/rfc4861#ref-RTSEL)).
> 
> The issue you are pointing out is that the RFC 4191 support isn't that universally widespread. Am I correct, @systemcrash?

Correct - either by a miss or by some kind of collective "oh, it's microsoft trying to embrace, extend, extinguish" again. The Linux header files never adopted the necessary definitions, but the values are laid out in the RFC so implementing it isn't hard - there's just no guarantee that a Linux based client would understand it, as you observed in your tests, I think.

You're welcome to implement it - it should be simple enough. It might cause confusion, however.

Do you think the preference difference is attributable to the routers own preference flag, or that included with the RI?

dbnk0:
> Do you think the preference difference is attributable to the routers own preference flag, or that included with the RI?

I'm not sure I understand your question, but based on my tests, I would assume the route preference on Ubuntu and Windows clients corresponds to preference in RI. On Android, I am not certain -- it could correspond to preference in RI, but also to the router preference embedded in RA itself.

systemcrash:
Does implementing 4191 mean a new PR, or integrating into this one?

dbnk0:
I will close this one and create a new PR later. There will be a different config interface and semantics, so I think it's better to separate the new implementation and keep old one for reference in it's original form.

PR #225, opened 2024-12-29
config: clamp configurable values to within sensible ranges

systemcrash:
Clamp values read from config to RFC mandated sane values instead of just
complaining.  Fixed also a bug in MTU handling.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -431,18 +431,14 @@ static double parse_leasetime(struct blob_attr *c) {
 	double time = strcmp(val, "infinite") ? strtod(val, &endptr) : UINT32_MAX;
 
 	if (time && endptr && endptr[0]) {
-		if (endptr[0] == 's')
-			time *= 1;
-		else if (endptr[0] == 'm')
-			time *= 60;
-		else if (endptr[0] == 'h')
-			time *= 3600;
-		else if (endptr[0] == 'd')
-			time *= 24 * 3600;
-		else if (endptr[0] == 'w')
-			time *= 7 * 24 * 3600;
-		else
-			goto err;
+		switch(endptr[0]) {
+			case 's': break; /* seconds */
+			case 'm': time *= 60; break; /* minutes */
+			case 'h': time *= 3600; break; /* hours */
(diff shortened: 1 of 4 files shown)


systemcrash:
ping @Ansuel

systemcrash:
These have been tested on v22 and v23.

Ansuel reviewed v1 of the patchset. Could I get a review, please ? @ldir-EDB0 or @dangowrt ?

systemcrash:
friendly ping @Ansuel  :)

systemcrash:
Ready for review @Noltari @Alphix. 

The only bit I'm still thinking about is the third commit which sets:
`#define RouterLifetime					5400` 
(and therefore a codeblock caps lifetimes to this as a max value). But this is anyway RFC mandated so in line with recent changes and RFC based philosophy.

Noltari (approved):
LGTM, @Alphix any objections?

Alphix:
LGTM, and these things are not like complicated restructuring of code...if a limit turns out to be wrong (the RFCs are a bit difficult to follow since many of them have been altered by so many later RFCs) or cause issues in the wild, it'll be very easy to change the limits with a one-line

systemcrash:
Late thought: maybe those LOG_ERROR classes can be revised to LOG_WARN. Since the init code tries to help by clamping to sane values instead of just dumping an error (and continuing). The original code also assumed that at least the value read in would have been the correct type (e.g. int for a numeric value and not some random char value which warrants an error), so I think we can de-escalate from ERROR.

Noltari:
> Late thought: maybe those LOG_ERROR classes can be revised to LOG_WARN. Since the init code tries to help by clamping to sane values instead of just dumping an error (and continuing). The original code also assumed that at least the value read in would have been the correct type (e.g. int for a numeric value and not some random char value which warrants an error), so I think we can de-escalate from ERROR.

@systemcrash I agree. Please take care of that and ping me when the PR is ready for merge.

systemcrash:
LGTM

Noltari:
Merged, thanks @systemcrash

PR #231, opened 2025-03-28
relay: forward message if dns rewriting is impossible

mkresin:
If the DHCPV6_OPT_AUTH option is set in a DHCPv6 server response, we can't rewrite the DNS server within the response as it would result in a checksum mismatch.

Instead of silently dropping the server response, forward the unmodified response instead.

### DIFF
--- a/src/dhcpv6.c
+++ b/src/dhcpv6.c
@@ -761,6 +761,31 @@ static void handle_dhcpv6(void *addr, void *data, size_t len,
 	}
 }
 
+static void rewrite_dns(struct interface *iface, struct in6_addr *dns_ptr, size_t dns_count)
+{
+	if (!dns_ptr || dns_count == 0)
+		return;
+
+	const struct in6_addr *rewrite = iface->dns;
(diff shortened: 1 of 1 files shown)


src/dhcpv6.c
```diff
@@ -761,6 +761,31 @@ static void handle_dhcpv6(void *addr, void *data, size_t len,
...
+	if (!dns_ptr || dns_count == 0)
+		return;
+
+	const struct in6_addr *rewrite = iface->dns;
+	struct in6_addr addr;
+	size_t rewrite_cnt = iface->dns_cnt;
+
+	if (rewrite_cnt == 0) {
+		if (odhcpd_get_interface_dns_addr(iface, &addr)) {
+			syslog(LOG_ERR, "Unable to get interface address");
```
systemcrash, line 775: Perhaps some more context in the error message. e.g.

`			syslog(LOG_ERR, "Unable to get interface address while rewriting DNS servers during relay of server response ");`

PR #232, opened 2025-03-28
relay: respect dhcpv6_[na|pd] config option

mkresin:
Strip the DHCPV6_OPT_IA_NA or DHCPV6_OPT_IA_PD prior to forwarding the client request.

### DIFF
--- a/src/dhcpv6.c
+++ b/src/dhcpv6.c
@@ -872,6 +872,48 @@ static struct odhcpd_ipaddr *relay_link_address(struct interface *iface)
 	return addr;
 }
 
+static bool filter_client_request(struct interface *iface, void *data, size_t *len) {
+	struct dhcpv6_client_header *h = data;
+	uint8_t *start = data + sizeof(*h);
+	uint8_t *odata, *end = data + *len;
+	uint16_t otype, olen;
+	uint8_t *tmp, *p_tmp;
+
+	if (h->msg_type != DHCPV6_MSG_SOLICIT)
+		return false;
+
+	if (iface->dhcpv6_na && iface->dhcpv6_pd)
+		return false;
+
+	tmp = p_tmp = malloc(end - start);
+	if (!tmp)
+		return false;
+
+	dhcpv6_for_each_option(start, end, otype, olen, odata) {
+		/* point to the start of the option and not the option data */
+		odata -= sizeof(otype) + sizeof(olen);
+		olen += sizeof(otype) + sizeof(olen);
+
+		if (otype == DHCPV6_OPT_IA_NA && !iface->dhcpv6_na) {
+			*len -= olen;
+			continue;
+		}
+
+		if (otype == DHCPV6_OPT_IA_PD && !iface->dhcpv6_pd) {
+			*len -= olen;
+			continue;
+		}
+
+		memcpy(p_tmp, odata, olen);
+		p_tmp += olen;
+	}
+
+	memcpy(start, tmp, p_tmp - tmp);
+	free(tmp);
+
+	return true;
+}
+
 /* Relay client request (regular DHCPv6-relay) */
 static void relay_client_request(struct sockaddr_in6 *source,
 		const void *data, size_t len, struct interface *iface)
@@ -904,6 +946,11 @@ static void relay_client_request(struct sockaddr_in6 *source,
 			return; /* Invalid hop count */
 
 		hdr.hop_count = h->hop_count + 1;
+	}
+
+	if (filter_client_request(iface, (void *)data, &len)) {
+		iov[1].iov_len = len;
+		hdr.relay_message_len = htons(len);
 	}
 
 	/* use memcpy here as the destination fields are unaligned */


systemcrash:
Can you detail the reasoning for this? What is the effect of before and after the patch?

mkresin:
> What is the effect of before and after the patch?

Huh? Isn't it described in the commit message?

odhcpd is used as DHCPv6 relay. odhcpd receives a solicit from a client with the request options DHCPV6_OPT_IA_NA and/or DHCPV6_OPT_IA_PD set. In case the patch is applied, the already existing uci config options dhcpv6_na and dhcpv6_pd are evaluated.

If `dhcpv6_na = 0`, the DHCPV6_OPT_IA_NA part is stripped from the solicit prior to forwarding the request to the upstream DHCPv6 server.

If `dhcpv6_pd = 0`, the DHCPV6_OPT_IA_PD part is stripped from the solicit prior to forwarding the request to the upstream DHCPv6 server.

Without the patch, the solicit is forwarded as is to the upstream DHCPv6 server.

> Can you detail the reasoning for this?

Fair question!

I do have a test setup which replicates my ISPs setup. For the sake of having GUAs, the test setup is part of my LAN.

The OpenWrt "test client" should only get a prefix delegated but no IPv6 address via DHCPv6. It replicates the behaviour of my ISP. The DHCPv6 server finally handling the requests is my main LAN DHCPv6 server.

One way to archiving the goal would be to configure my main odhcpd DHCPv6 server this way. But the DHCPv6 server serves my LAN as well, where DHCPv6 based addressing is required.

Another way would be, to simply set `dhcpv6_na = 0` on the OpenWrt "test client" to prevent asking for a IPv6 address via DHCPv6. But that would not exactly match the configuration of my main OpenWrt router, as my OpenWrt main router asks for an IPv6 address + delegated prefix via DHCPv6 but only gets a prefix assigned and no IPv6 address from my ISP.

Hence this proposal to strip the DHCPV6_OPT_IA_NA request option prior to relaying the solicit to the upstream DHCPv6 server.

systemcrash:
Oh :) Amend your commit message to detail that functionality at a higher level. Extremely useful later on down the road. This only affects forwarded requests, yes?

mkresin:
> This only affects forwarded requests, yes?

By no means I want to be unpolite, but have you had a look at the changes? I mean, the altered function is called `relay_client_request()` and has the comment

> Relay client request (regular DHCPv6-relay)

So yes, this change only affects the DHCPv6 relay mode where messages are forwarded to another DHCPv6 server. Or at least that is my intention.

systemcrash:
In that case, please revise the commit message to:
```
relay: remove relayed DHCPV6_OPT_IA_[NA|PD] when dhcpv6_[na|pd] = 0

Respect the uci config options dhcpv6_na and dhcpv6_pd in DHCPv6
relay mode as well.

If dhcpv6_na = 0, odhcpd removes the DHCPV6_OPT_IA_NA from the
client solicit message prior to forwarding the message to the upstream
DHCPv6 server.

If dhcpv6_pd = 0, odhcpd removes the DHCPV6_OPT_IA_PD from the
client solicit message prior to forwarding the message to the upstream
DHCPv6 server.

This patch enables control over what kind of addresses are
returned to the requesting client. Similar to the existing
feature "replace DNS server sent by upstream DHCPv6 server".

Signed-off-by: Mathias Kresin <dev@kresin.me>
```

I did not find the 'similar' feature - which one were you referring to?

src/dhcpv6.c
```diff
@@ -872,6 +872,48 @@ static struct odhcpd_ipaddr *relay_link_address(struct interface *iface)
...
+	uint8_t *start = data + sizeof(*h);
+	uint8_t *odata, *end = data + *len;
+	uint16_t otype, olen;
+	uint8_t *tmp, *p_tmp;
+
+	if (h->msg_type != DHCPV6_MSG_SOLICIT)
+		return false;
+
+	if (iface->dhcpv6_na && iface->dhcpv6_pd)
+		return false;
```
systemcrash, line 886: These can be checked before we enter this function already.

systemcrash:
OK - this makes sense. @Alphix ?

Alphix:
Hmm....my operational knowledge of DHCPv6 relaying is a bit hazy, but what about [RFC8987, §4.1](https://datatracker.ietf.org/doc/html/rfc8987#section-4.1), requirement G-1?

systemcrash:
Yeah, that’s a big must which rules this out, were it to be a static behaviour, but since this behaviour is configurable, I figure it’s a nice to have, so we’re no worse off.

Alphix:
> Yeah, that’s a big must which rules this out, were it to be a static behaviour, but since this behaviour is configurable, I figure it’s a nice to have, so we’re no worse off.

From what I've read so far...this seems like an very niche use-case...I don't think this should go in. From my understanding, `dhcpv6_na` and `dhcpv6_pd` were never meant to cover the behaviour in relaying...so it would be a change in behaviour...

Alphix:
@systemcrash @Noltari I think this is a case of `wontfix`...it's a niche case that's contrary to the relevant RFCs

PR #233, opened 2025-04-01
feat: support multiple mac addresses per static lease

### LINKED ISSUE #221: Support `list mac` in /etc/config/dhcp
Please add support for the new 23.05.x syntax. Static leases stopped working with recent OpenWRT releases. Associated forum post: https://forum.openwrt.org/t/dhcp-static-leases-stopped-working-properly/197380/3

### LINKED ISSUE #9598: odhcpd: Static leases with multiple MAC addresses get skipped when configuring odhcpd
In https://openwrt.org/docs/guide-user/base-system/dhcp#static_leases, the _mac_ option of a DHCP _Static Lease_ is defined as:
```
mac | string | no | (none) | The hardware address(es) of this host, separated by spaces.
```

This explicitly allows to define _leases_ with _multiple MAC addresses_. However, _odhcpd_ skips such _leases_ during configuration.

Here's the code fragment which handles the MAC address(es) of static leases in `set_lease(..)` (found in `config.c`):
``` c
	if ((c = tb[LEASE_ATTR_MAC]))
		if (!ether_aton_r(blobmsg_get_string(c), &l->mac))
			goto err;
```
As can be seen, the code doesn't have any special handling for the multiple MAC case. So `ether_aton_r(..)` returns `NULL` which triggers `goto err;` and the lease doesn't make it into _leases_.

YodaDaCoda:
fixes openwrt/odhcpd#221
fixes openwrt/openwrt#9598

Very keen for some feedback on this - my C is pretty rusty but I think the adjustments I've made make sense.

Haven't been able to test this beyond compiling it and giving it a config file to parse (confirmed it works with both the old `option mac` and the new `list mac` formats) but I don't think there's any other changes required to support this.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -181,7 +181,7 @@ const struct uci_blob_param_list interface_attr_list = {
 
 const struct blobmsg_policy lease_attrs[LEASE_ATTR_MAX] = {
 	[LEASE_ATTR_IP] = { .name = "ip", .type = BLOBMSG_TYPE_STRING },
-	[LEASE_ATTR_MAC] = { .name = "mac", .type = BLOBMSG_TYPE_STRING },
+	[LEASE_ATTR_MAC] = { .name = "mac", .type = BLOBMSG_TYPE_ARRAY },
 	[LEASE_ATTR_DUID] = { .name = "duid", .type = BLOBMSG_TYPE_STRING },
 	[LEASE_ATTR_HOSTID] = { .name = "hostid", .type = BLOBMSG_TYPE_STRING },
 	[LEASE_ATTR_LEASETIME] = { .name = "leasetime", .type = BLOBMSG_TYPE_STRING },
@@ -434,6 +434,7 @@ err:
 static void free_lease(struct lease *l)
 {
 	free(l->hostname);
+	free(l->mac_arr);
 	free(l);
 }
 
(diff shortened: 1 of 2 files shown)


systemcrash:
This would be a welcome addition. This is driven by the 'host' format which can have multiple MAC addresses. The list format is handled transparently by dnsmasq.init script, but it would appear that this daemon cannot yet.

schuettecarsten (approved):

fda77:
If MAC-1 gets IP X, and then MAC-2 requestes also one, does both MAC get the same IP X? Or only if lease of MAC-1 is timed out etc?
How is /etc/ethers handled which can not have multiple MACs for the same entry?
Will there be a patch for all the server software like dnsmasq to be able to handle multiple  MAC for 1 IP?

systemcrash:
> Will there be a patch for all the server software like dnsmasq to be able to handle multiple MAC for 1 IP?

dnsmasq handles it natively. https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

But only one MAC can have the IP concurrently. Its use case is outlined:
```
As a special case, in DHCPv4, it is possible to include more than one hardware address. 
eg: --dhcp-host=11:22:33:44:55:66,12:34:56:78:90:12,192.168.0.2. This allows an IP address
to be associated with multiple hardware addresses, and gives dnsmasq permission to abandon
a DHCP lease to one of the hardware addresses when another one asks for a lease. Beware that
this is a dangerous thing to do: it will only work reliably if only one of the hardware addresses
is active at any time and there is no way for dnsmasq to enforce this. It is, for instance, useful
to allocate a stable IP address to a laptop which has both wired and wireless interfaces. 
```

fda77:
Thx, this is a bad hack of dnsmasq, which could not be read from /etc/ethers too...
I still think its not a good idea and could cause much trouble

thess:
@dedeckeh - Applied and tested this fix on Linksys E8450. Solves the problem with LuCI making MAC addresses in DHCP UCI `host` config `list` instead of `option` entries. I'm embarrassed to say it took me more than a day to figure out why previously configured static leases suddenly stopped working after a power outage that rebooted the router.

I recommend merging this patch either as-is or with the code of your choice. It is a somewhat critical necessity due to the LuCI/DNSMASQ changes for multiple MAC addresses which affects even a single MAC entry. (That said, I do find the feature sketchy at best)

thess (approved):

CtrlC-Root:
Found this PR after encountering the same breaking issue in the latest OpenWRT 24.10.2 release. What's left to do before merging this PR?

Alphix:
No disrespect to @YodaDaCoda, but I have submitted an alternative implementation in PR https://github.com/openwrt/odhcpd/pull/251

thess:
Likewise, no disrespect to @YodaDaCoda for this proposal. This fix was much needed in spite of the controversial side-effects.

Closing this for now in favor of #251. Thanks again.

PR #234, opened 2025-04-02
config: support ip=ignore in static leases config

### LINKED ISSUE #198: ip=ignore is ignored
I'm trying to configure odhcpd to ignore a specific machine MAC address and I think there is a bug somewhere.

The [static leases documentation](https://openwrt.org/docs/guide-user/base-system/dhcp#static_leases) states:

> ip | the IP address to be used for this host, or `ignore` to ignore any DHCP request from this host

But the [odhcpd host section](https://openwrt.org/docs/techref/odhcpd#host_section) does not state `ignore` is supported, which is somewhat odd?

I've configured a static lease like:

```bash
# see https://openwrt.org/docs/guide-user/base-system/dhcp#static_leases
# see https://openwrt.org/docs/techref/odhcpd#host_section
# see /etc/config/dhcp
# see /tmp/hosts/odhcpd
ssh root@192.168.1.254
#uci delete dhcp.@host[-1]
id="$(uci add dhcp host)"
uci set "dhcp.$id.mac=08:00:27:00:00:46"
# XXX even with this configuration, the host is not being ignored :-(
uci set "dhcp.$id.ip=ignore"
uci set "dhcp.$id.name=t1"
uci changes dhcp
uci commit dhcp
uci show dhcp
service odhcpd reload
exit
```

But I can still see DHCP traffic, and odhcpd is still assigning/replying to the configured MAC address.

Can you help understand why odhcpd is not ignoring the defined MAC address?

Am I doing something wrong?

Kasoo:
Allows hosts to be ignored from DHCP and makes odhcpd compatible with the configuration as documented here:

https://openwrt.org/docs/guide-user/base-system/dhcp#static_leases

fixes #198

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -1110,13 +1110,14 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 	struct dhcp_assignment *a = find_assignment_by_hwaddr(iface, mac);
 	struct lease *l = config_find_lease_by_mac(mac);
 	time_t now = odhcpd_time();
+	bool ignored = (l && l->ignore);
 
 	/*
 	 * If we found a static lease cfg, but no old assignment for this
 	 * hwaddr, we need to clear out any old assignments given to other
 	 * hwaddrs in order to take over the IP address.
 	 */
-	if (l && !a && (msg == DHCPV4_MSG_DISCOVER || msg == DHCPV4_MSG_REQUEST)) {
+	if (l && !ignored && !a && (msg == DHCPV4_MSG_DISCOVER || msg == DHCPV4_MSG_REQUEST)) {
 		struct dhcp_assignment *c, *tmp;
 
 		list_for_each_entry_safe(c, tmp, &l->assignments, lease_list) {
@@ -1125,7 +1126,7 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 		}
 	}
 
-	if (l && a && a->lease != l) {
+	if (l && a && (a->lease != l || ignored)) {
 		free_assignment(a);
 		a = NULL;
 	}
@@ -1139,7 +1140,7 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 		bool assigned = !!a;
 
 		if (!a) {
-			if (!iface->no_dynamic_dhcp || l) {
+			if ((!iface->no_dynamic_dhcp || l) && !ignored) {
 				/* Create new binding */
 				a = alloc_assignment(0);
 				if (!a) {
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ and may also receive information from ubus
 ### Sections of type host (static leases)
 | Option		| Type	|Default| Description |
 | :-------------------- | :---- | :---- | :---------- |
-| ip			|string	|(none) | IPv4 host address |
+| ip			|string	|(none) | IPv4 host address or `ignore` to ignore any DHCP request from this host |
 | mac			|list\|string|(none) | HexadecimalMACaddress(es) |
 | duid			|string	|(none)	| HexadecimalDUID |
 | hostid		|string	|(none)	| IPv6hostidentifier |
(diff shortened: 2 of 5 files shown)


Kasoo:
Note: I've not been able to test this yet as I can't find any documentation on how to build odhcpd for use with openwrt.

Would appreciate if someone could point me in the right direction.

Prototik:
@Kasoo [you can point](https://openwrt.org/docs/guide-developer/packages#config_src_tree_override) git-src symlink in the openwrt tree to your local git copy of odhcpd to use that sources.

Kasoo:
Thanks for the pointer @Prototik, I've now been able to build and test the code, looks good to go as far as I'm concerned

systemcrash:
Please rebase

systemcrash:
ping @Kasoo

Kasoo:
Rebased on latest `master`

systemcrash:
@Noltari @Alphix

Noltari (approved):
LGTM

@Alphix any objections?

src/dhcpv4.c
```diff
@@ -1110,13 +1110,14 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 	struct dhcp_assignment *a = find_assignment_by_hwaddr(iface, mac);
 	struct lease *l = config_find_lease_by_mac(mac);
 	time_t now = odhcpd_time();
+	bool ignored = (l && l->ignore);
```
Alphix, line 1113: Hm, maybe I'm missing something, but right now `ignored` is checked three times later in this function, and in the end the state file is updated, etc, even though nothing should change...and then the function returns `NULL`.

Can't we just return `NULL` here straight away and skip all the other checks?

And the same for `dhcpv6_ia_handle_IAs()` but right after the `config_find_lease_*()` calls...

Kasoo: I believe my logic to try and handle that case that a client is marked as ignored, but we still have an existing assignment for them, trying to only handle the ignored logic on the initial request but still communicating with clients that have an assignment even if they are ignored, so we could handle things like `DHCPV4_MSG_RELEASE` properly.

But re-reading the logic now I think the logic around line 1175-1231 would presumably mean that that an ignored client with an outstanding assignment could just keep refreshing that assignment forever and would never be ignored. Perhaps that isn't the best expected behaviour.

If we were to change to early return null, we'll still need to clean up any potentially existing assignment before returning, but I can do that.

Alphix: I'm not sure that we *really* need to handle the case of `ignore` + outstanding lease...

One reason being that it complicates the code for what I assume is a rare corner-case.

The other reason being that I'm working on adding explicit support to `odhcpd` for revoking an existing lease via `ubus` calls (and adding a button to LuCI for that purpose), so that would make it easy to do the `ignore` + `revoke` dance if the administrator wishes to do so.

Alphix:
An on a random note...I wonder if we should add support in the future (in a different PR, not suggesting you implement it here) for `reject` and `ignore` as two separate options (one leading to a NACK, the other meaning the client will be ignored completely and get no replies at all). OTOH, it might have little practical value 😄

src/dhcpv4.c
Kasoo: Makes sense, I'll refactor to simplify

systemcrash: Does it really make sense to handle ignore for existing leases on the wire? odhcpd would have to re-read its config to know about ignore while it has an existing lease, and it would forget the existing lease at reload...?

Alphix: > Does it really make sense to handle ignore for existing leases on the wire? odhcpd would have to re-read its config to know about ignore while it has an existing lease, and it would forget the existing lease at reload...?

I'm working changing the code which writes leases to `/tmp` so they'll be read at startup (i.e. survive a `odhcpd` restart). But anyway, we seem to agree that existing leases can be ignored for now...

systemcrash:
OK. LGTM?

Alphix:
> OK. LGTM?

@systemcrash I think the current state is that @Kasoo said "I'll refactor to simplify"?

systemcrash:
> > OK. LGTM?
> 
> @systemcrash I think the current state is that @Kasoo said "I'll refactor to simplify"?

Hmm, could have sworn it was this PR I saw a refresh. 🤷

Noltari:
Superseeded by https://github.com/openwrt/odhcpd/pull/303

PR #235, opened 2025-04-04
Add support for RFC8925 IPv6-Only Preferred Option for DHCPv4

Kasoo:
Adds a new interface configuration variable `dhcpv4_ipv6only` which enables support for [RFC8925 IPv6-Only Preferred mode](https://datatracker.ietf.org/doc/rfc8925/), along with `dhcpv4_ipv6only_wait` which configures the time clients should wait before checking DHCPv4 again.
    
This means that IPv6-mostly networks that provide still provide an IPv4 DHCP server for legacy clients can allows clients that support IPv6-only to opt-out of having an IPv4 address assigned to them.
    
Notably MacOS 12, iOS 15 and Android 12 all support RFC8925 currently.


Kasoo:
Note: This adds new [ /etc/config/dhcp configuration](https://openwrt.org/docs/guide-user/base-system/dhcp) that's not currently mirrored  by dnsmasq.

I'm not sure what the policy is regarding maintaining parity between odhcp and dnsmasq when it comes to ` /etc/config/dhcp`, it seems like it is intended that most configuration it usable by both, but there are other that are only in one or the other.

Its possible to partially support RFC8925 in dnsmasq using `list dhcp_option '108,0000i'`, however i opted to pull this out as a separate feature here because technically doing that is not strictly conforming to the RFC. Using that flag with dnsmasq would still have a IPv4 address allocated to the IPv6-only client, only for it to be ignored, whereas the implementation in this PR handles that correctly.

However if the configuration divergence is a problem it would be trivial to make `dhcpv4_ipv6only` map into the ` --dhcp-option` flag for dnsmasq if that is desired.

I also opted against of supporting `dhcp_option` fully in odhcpd, but that's also technically possible, but then I'd still have the question about whether using that would be RFC compliant.

Finally, I added two separate configurations `dhcpv4_ipv6only` and `dhcpv4_ipv6only_wait` when only one is really needed. Again I chose to do this because the RFC suggests that implementations SHOULD have both. Additionally if we only had the `dhcpv4_ipv6only_wait` time (and used `0` to signal "not enabled"), its not clear how to communicate that the default timeout is suggested to be 1800 to potential users.

systemcrash:
Preferable that everything is found in the dhcp config file. The wiki could do with a column for each property which denotes which daemons depend on it.

src/config.c
```diff
@@ -1452,6 +1456,21 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 		}
 	}
 
+	if ((c = tb[IFACE_ATTR_IPV6ONLY])) {
+		if (blobmsg_get_bool(c)) {
+			iface->ipv6only_wait = 1800;
```
systemcrash: Can't these static values be made into a #DEFINE so that they're easier to find? Setting static values is fine: edit: they're defined: https://www.rfc-editor.org/rfc/rfc8925.html#section-3.4

Kasoo: Happy to do so, I didn't previously because all the other literals in the same file don't have defines.

systemcrash: Perhaps in `odhcpd.h` since it's already included?

Alphix:
Can't the two config options be combined into a single one? Other cfg options don't have separate enable/disable and value settings...?

systemcrash:
Like `ipv6only_wait > 0` then it's enabled.

systemcrash:
ping @Kasoo - please revise your commit message:

```
Added a new interface configuration variable `dhcpv4_ipv6only` which
enables support for RFC8925 IPv6-Only Preferred mode, along with
`dhcpv4_ipv6only_wait` which configures the time clients should wait
before checking DHCPv4 again.

This means that IPv6-mostly networks that still provide an IPv4
DHCP server for legacy clients allows clients that support IPv6-only
to opt-out of having an IPv4 address assigned to them.

Notably MacOS 12, iOS 15 and Android 12 all support RFC8925 currently.
```

src/config.c
Kasoo: Will update to add `#define` in `odhcpd.h`.

Kasoo:
> Like `ipv6only_wait > 0` then it's enabled.

I had considered this, and happy to do it. The main reason I didn't do it originally is that the RFC says:

```
V6ONLY_WAIT:
    The time for which the client SHOULD stop the DHCPv4 configuration process. The value MUST NOT be less than MIN_V6ONLY_WAIT seconds. Default: 1800 seconds

MIN_V6ONLY_WAIT:
    The lower boundary for V6ONLY_WAIT. Value: 300 seconds
```

We want to have the default setting to be disabled, but then if it is enabled the default setting should be 1800 and this could be configured to a different value.

Having only a `ipv6only_wait` configuration value makes it harder to communicate that the recommended default is 1800.

Perhaps is it possible to have a single value that takes `boolean|number`? Where setting it to `1` actually enables the default behaviour (1800) and setting it to a real value above 300 sets it to that time?

Kasoo:
I've rebased on latest master, and condensed down to a single configuration option `ipv6only_wait` which supports the values `0: disabled, 1: default timeout (1800s), 300+: custom timeout(s).`

Also added `#define` for the configuration literals (`DHCPV4_MIN_V6ONLY_WAIT`, `DHCPV4_DEFAULT_V6ONLY_WAIT`)

systemcrash:
@Alphix @Noltari

Noltari (approved):
LGTM, but I will wait for @Alphix feedback since he reviewed this one too.

src/odhcpd.h
```diff
@@ -375,6 +375,7 @@ struct interface {
 	struct in_addr *dhcpv4_dns;
 	size_t dhcpv4_dns_cnt;
 	bool dhcpv4_forcereconf;
+	uint32_t ipv6only_wait;
```
Alphix, line 378: `struct interface` is already 2+ pages long when printed out (yes, I have a printout of it on my desk right now, doesn't everyone?), I'd suggest renaming this to `dhcpv4_ipv6only_wait` to make it clear that it's related to the DHCPv4 server cfg even though the variable name contains a reference to IPv6 (the bool in `dhcpv4_handle_msg()` doesn't need a longer name)

systemcrash:
ping @Kasoo

Alphix:
> ping @Kasoo

@systemcrash maybe you could hold a bit on this one? I have a big bunch of changes to `dhcpv4.c` in the pipeline, and it would be much easier to rework this patch after that PR lands (it's waiting for the no-syslog stuff first)....I could create an alternative version of this PR after that (no disrespect to @Kasoo)

systemcrash:
@Alphix where are we on this? Did you want to make a PR for this?

Alphix:
> @Alphix where are we on this? Did you want to make a PR for this?

I have a bunch of options (`clientid`, `timezone`, `sip`, custom options, `ipv6-only-pref`, `ignore`, etc) lined up for DHCPv4...but they're waiting for more refactoring...specifically PR #286....and one more that I haven't submitted yet (because it also needs to come after #286) that renames the misleading `struct lease` into something better...and splits `struct dhcp_assignment` into `struct dhcpv4_lease` and `struct dhcpv6_lease`

PR #240, opened 2025-05-26
ndp: Loopback is allowed when iface is the master interface.

CalunVier:
### Introduction

This pull request introduces a small but significant change to the Neighbor Discovery Protocol (NDP) handling logic in the `src/ndp.c` file. The update ensures that the function `handle_solicit` checks if the interface is a master before determining if a packet is looped back.

### Key Change:
* [`src/ndp.c`](diffhunk://#diff-c44a33cae021d6841c57de750969fb66e5aed8540f7f6b813f7d9525951bf2d4L357-R357): Updated the condition in `handle_solicit` to include a check for `iface->master` to handle packets from the master iface itself.

### Why This Change Is Necessary

When IPv6 operates in relay mode, `odhcp` typically handles Neighbor Solicitation (NS) requests from the upstream network (master interface) correctly. It uses `ping6` to discover devices within the local subnet (slaved interface), ensuring that devices on the local network are properly discoverable.

However, this ideal scenario only works if an NS packet is actually sent from the upstream network. If no such NS packet is received, devices on the local subnet may remain undiscovered, and **any packet destined for those addresses will fail to be routed properly**. This can occur in the following situations:

1. If the upstream router had previously sent an NS request and cached the IP-to-MAC mapping (e.g., IP6 to OpenWrt's MAC) in its neighbor table, and that entry hasn't expired, while the corresponding neighbor entry on OpenWrt has already expired, then incoming packets from the Internet will be forwarded directly to OpenWrt without triggering a new NS. Since OpenWrt no longer has a valid neighbor entry, it will attempt to resolve the destination via the WAN (master) interface—because the route for that IPv6 prefix exists only there. However, the actual device resides on the LAN, and **the packet cannot be correctly forwarded to the intended device**.
2. If no external NS request was ever received, and OpenWrt tries to reach a long-idle local device (whose neighbor entry has expired), the NS request will be sent out via the WAN interface. But since the device is on the LAN, the NS packet is sent in the wrong direction, and the connection cannot be established.

This issue has impacted many users. I found numerous discussions about it online, which motivated me to attempt this fix.

### Fix Strategy

The fix is simple: allow OpenWrt to also process NS packets originating from the local system and sent to the WAN (master) interface. This behavior was previously blocked by a condition at line 357 in `ndp.c`. I’ve added an exception to permit this case.

### Pre-Merge Warning

Due to the complexity of the OpenWrt build environment, I haven’t been able to compile or test this patch. However, I’ve thoroughly reviewed the logic and “mentally compiled” it. I believe the change is correct. I hope maintainers can help validate and test the patch before merging.

### DIFF
--- a/src/ndp.c
+++ b/src/ndp.c
@@ -332,6 +332,7 @@ static void handle_solicit(void *addr, void *data, size_t len,
 	struct interface *c;
 	char ipbuf[INET6_ADDRSTRLEN];
 	uint8_t mac[6];
+	bool is_self_sent;
 
 	/* Solicitation is for duplicate address detection */
 	bool ns_is_dad = IN6_IS_ADDR_UNSPECIFIED(&ip6->ip6_src);
@@ -354,7 +355,8 @@ static void handle_solicit(void *addr, void *data, size_t len,
 	syslog(LOG_DEBUG, "Got a NS for %s on %s", ipbuf, iface->name);
 
 	odhcpd_get_mac(iface, mac);
-	if (!memcmp(ll->sll_addr, mac, sizeof(mac)))
+	is_self_sent = !memcmp(ll->sll_addr, mac, sizeof(mac));
+	if (is_self_sent && !iface->master)
 		return; /* Looped back */
 
 	avl_for_each_element(&interfaces, c, avl) {
@@ -366,7 +368,7 @@ static void handle_solicit(void *addr, void *data, size_t len,
 	/* Catch global-addressed NS and answer them manually.
 	 * The kernel won't answer these and cannot route them either. */
 	if (!IN6_IS_ADDR_MULTICAST(&ip6->ip6_dst) &&
-			IN6_IS_ADDR_LINKLOCAL(&ip6->ip6_src)) {
+			IN6_IS_ADDR_LINKLOCAL(&ip6->ip6_src) && !is_self_sent) {
 		bool is_proxy_neigh = netlink_get_interface_proxy_neigh(iface->ifindex,
 				&req->nd_ns_target) == 1;
 


systemcrash:
@CalunVier please add your SOB line, and describe the commits in the commit messages.

CalunVier:
> @CalunVier please add your SOB line, and describe the commits in the commit messages.

I have added the SOB line and refined the commit message.

Noltari:
@CalunVier according to the OpenWrt rules you must use your real name in the SoB line:
https://openwrt.org/submitting-patches#sign_your_work

CalunVier:
> @CalunVier according to the OpenWrt rules you must use your real name in the SoB line: https://openwrt.org/submitting-patches#sign_your_work

Please excuse me, this is my first time submitting code to a serious infrastructure repository. I’ve updated the SoB line and used my legal name. The latest commit references a different GitHub account, which represents my legal identity. I'm simply used to surfing the web under the name CalunVier.

Noltari:
@systemcrash if you have no objections I can merge this

systemcrash:
@CalunVier can you include (some more of) the "Why This Change Is Necessary" in the commit message? That's a great explanation.

CalunVier:
> @CalunVier can you include (some more of) the "Why This Change Is Necessary" in the commit message? That's a great explanation.

I added some background information to the commit message. I hope it explains “Why This Change Is Necessary.”

systemcrash:
@Noltari ready (unless you want to reformat the commit message wrap)

Noltari:
> @Noltari ready (unless you want to reformat the commit message wrap)

Yes, but I can do that myself, no worries 😉

Noltari:
Merged https://github.com/openwrt/odhcpd/commit/f0d855358b86a36efbfeb5a9de5a1d2a4d9d80fe, thanks @CalunVier!

BTW, next time try to wrap the commit summary at 70-75 characters and also create a git branch instead of committing directly to your master branch.

CalunVier:
> Merged [f0d8553](https://github.com/openwrt/odhcpd/commit/f0d855358b86a36efbfeb5a9de5a1d2a4d9d80fe), thanks @CalunVier!
> 
> BTW, next time try to wrap the commit summary at 70-75 characters and also create a git branch instead of committing directly to your master branch.

Thank you, I've learned a lot.

Issue #202, opened 2023-07-31
odhcpd IPv6 NDP + macOS don't play together


golddranks:
I'm just cross-posting this issue here, because it's more relevant here. I feel like it has been languishing in the main OpenWRT repo for years.

https://github.com/openwrt/openwrt/issues/7561

The gist is that odhcpd seems to use ULA to send neighbor solicitation messages, when ULA is set up. MacOS doesn't respond to these messages, and fails to communicate its IPv6 address to the router. (In case of SLAAC). Everything works when ULA is disabled; MacOS seems to respond to messages that have LLA as the source address.

It would be better, if there were an option to force LLA. Currently having an ULA set up in the router settings basically prevents one from using MacOS in the network.

systemcrash:
Can this be closed? @golddranks please test with current odhcpd master

Noltari:
Fixed with d402cdae431668f55f9d82b7072b0afa3b8090df.
Feel free to reopen if the issue still persists.

PR #242, opened 2025-06-11
ndp: fix macOS IPv6 compatibility by using link-local source addresses

stephengroat:
macOS ignores NDP packets that don't originate from link-local addresses,
causing IPv6 connectivity issues with odhcpd. This change ensures NDP
packets (Neighbor Advertisements and ICMP Echo Requests) are sent using
link-local source addresses for RFC 4861 compliance.

Changes:
- Add ndp_from_link_local configuration flag
- Add odhcpd_send_with_src() to allow explicit source address control
- Add odhcpd_get_interface_linklocal_addr() to find link-local addresses
- Update send_na() and ping6() to prefer link-local source addresses when flag is enabled
- Add RFC 4861 section 4.2 comments explaining the mandated behavior
- Maintain backward compatibility with fallback behavior

Fixes: https://github.com/openwrt/openwrt/issues/7561 https://github.com/openwrt/odhcpd/issues/202

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -117,6 +117,7 @@ enum {
 	IFACE_ATTR_PD_CER,
 	IFACE_ATTR_NDPROXY_ROUTING,
 	IFACE_ATTR_NDPROXY_SLAVE,
+	IFACE_ATTR_NDP_FROM_LINK_LOCAL,
 	IFACE_ATTR_PREFIX_FILTER,
 	IFACE_ATTR_MAX_PREFERRED_LIFETIME,
 	IFACE_ATTR_MAX_VALID_LIFETIME,
@@ -171,6 +172,7 @@ static const struct blobmsg_policy iface_attrs[IFACE_ATTR_MAX] = {
 	[IFACE_ATTR_RA_PREF64] = { .name = "ra_pref64", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_NDPROXY_ROUTING] = { .name = "ndproxy_routing", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_NDPROXY_SLAVE] = { .name = "ndproxy_slave", .type = BLOBMSG_TYPE_BOOL },
+	[IFACE_ATTR_NDP_FROM_LINK_LOCAL] = { .name = "ndp_from_link_local", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_PREFIX_FILTER] = { .name = "prefix_filter", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_MAX_PREFERRED_LIFETIME] = { .name = "max_preferred_lifetime", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_MAX_VALID_LIFETIME] = { .name = "max_valid_lifetime", .type = BLOBMSG_TYPE_STRING },
@@ -271,6 +273,8 @@ static void set_interface_defaults(struct interface *iface)
 	iface->ra = MODE_DISABLED;
 	iface->ndp = MODE_DISABLED;
 	iface->learn_routes = 1;
+	iface->ndp_from_link_local = true;
+	iface->cached_linklocal_valid = false;
 	iface->dhcp_leasetime = 43200;
 	iface->max_preferred_lifetime = ND_PREFERRED_LIMIT;
 	iface->max_valid_lifetime = ND_VALID_LIMIT;
@@ -1450,6 +1454,9 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 
 	if ((c = tb[IFACE_ATTR_NDPROXY_SLAVE]))
 		iface->external = blobmsg_get_bool(c);
+
+	if ((c = tb[IFACE_ATTR_NDP_FROM_LINK_LOCAL]))
+		iface->ndp_from_link_local = blobmsg_get_bool(c);
 
 	if ((c = tb[IFACE_ATTR_PREFIX_FILTER]))
 		odhcpd_parse_addr6_prefix(blobmsg_get_string(c),
--- a/src/ndp.c
+++ b/src/ndp.c
@@ -280,7 +280,7 @@ static void ndp_netevent_cb(unsigned long event, struct netevent_handler_info *i
 /* Send an ICMP-ECHO. This is less for actually pinging but for the
  * neighbor cache to be kept up-to-date. */
 static void ping6(struct in6_addr *addr,
-		const struct interface *iface)
+		struct interface *iface)
 {
 	struct sockaddr_in6 dest = { .sin6_family = AF_INET6, .sin6_addr = *addr , };
 	struct icmp6_hdr echo = { .icmp6_type = ICMP6_ECHO_REQUEST };
@@ -291,13 +291,16 @@ static void ping6(struct in6_addr *addr,
 	syslog(LOG_DEBUG, "Pinging for %s on %s", ipbuf, iface->name);
 
 	netlink_setup_route(addr, 128, iface->ifindex, NULL, 128, true);
-	odhcpd_send(iface->ndp_ping_fd, &dest, &iov, 1, iface);
+
+	/* Use link-local address as source for RFC 4861 compliance and macOS compatibility */
+	odhcpd_try_send_with_src(iface->ndp_ping_fd, &dest, &iov, 1, iface);
+
 	netlink_setup_route(addr, 128, iface->ifindex, NULL, 128, false);
 }
 
 /* Send a Neighbor Advertisement. */
 static void send_na(struct in6_addr *to_addr,
-		const struct interface *iface, struct in6_addr *for_addr,
+		struct interface *iface, struct in6_addr *for_addr,
 		const uint8_t *mac)
 {
 	struct sockaddr_in6 dest = { .sin6_family = AF_INET6, .sin6_addr = *to_addr };
@@ -319,7 +322,8 @@ static void send_na(struct in6_addr *to_addr,
 	inet_ntop(AF_INET6, to_addr, ipbuf, sizeof(ipbuf));
 	syslog(LOG_DEBUG, "Answering NS to %s on %s", ipbuf, iface->ifname);
 
-	odhcpd_send(iface->ndp_ping_fd, &dest, &iov, 1, iface);
+	/* Use link-local address as source for RFC 4861 compliance and macOS compatibility */
+	odhcpd_try_send_with_src(iface->ndp_ping_fd, &dest, &iov, 1, iface);
 }
 
 /* Handle solicitations */
--- a/src/odhcpd.c
+++ b/src/odhcpd.c
@@ -206,10 +206,10 @@ int odhcpd_get_flags(const struct interface *iface)
 }
 
 
-/* Forwards a packet on a specific interface */
-ssize_t odhcpd_send(int socket, struct sockaddr_in6 *dest,
+/* Forwards a packet on a specific interface with optional source address */
+ssize_t odhcpd_send_with_src(int socket, struct sockaddr_in6 *dest,
 		struct iovec *iov, size_t iov_len,
-		const struct interface *iface)
+		const struct interface *iface, const struct in6_addr *src_addr)
 {
 	/* Construct headers */
 	uint8_t cmsg_buf[CMSG_SPACE(sizeof(struct in6_pktinfo))] = {0};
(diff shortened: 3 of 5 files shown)


systemcrash:
I think this is a good addition, but before it goes in I'd like to see its behaviour behind a preference flag, that defaults to the current behaviour. e.g.

```c
	if (odhcpd_get_interface_linklocal_addr(iface, &src_addr) == 0 && iface->ndp_from_link_local) {
		odhcpd_send_with_src(iface->ndp_ping_fd, &dest, &iov, 1, iface, &src_addr);
	} else {
		/* Fall back to default behavior if no link-local address is available */
		odhcpd_send(iface->ndp_ping_fd, &dest, &iov, 1, iface);
	}
```

Please also add some comments which call out to the RFC paragraphs about mandated behaviour. e.g.
```
https://www.rfc-editor.org/rfc/rfc4861#section-4.2

      Source Address
                     MUST be the link-local address assigned to the
                     interface from which this message is sent.
```

stephengroat:
Thanks for the comment, makes sense to flag this new paradigm to defaultly maintain current behavior.

I tried to address all comments, please let me know if there's anything else needed

systemcrash:
Given that the behaviour is RFC mandated, I think we can set the default to true, and now we have the flag, anyone experiencing problems (unlikely) can manually set the flag to false. 

Perhaps a brief review from @Alphix and @Noltari ? I think this is ready.

stephengroat:
default is now set to `true`

src/config.c
```diff
@@ -162,6 +163,7 @@ static const struct blobmsg_policy iface_attrs[IFACE_ATTR_MAX] = {
 	[IFACE_ATTR_RA_PREF64] = { .name = "ra_pref64", .type = BLOBMSG_TYPE_STRING },
 	[IFACE_ATTR_NDPROXY_ROUTING] = { .name = "ndproxy_routing", .type = BLOBMSG_TYPE_BOOL },
 	[IFACE_ATTR_NDPROXY_SLAVE] = { .name = "ndproxy_slave", .type = BLOBMSG_TYPE_BOOL },
+	[IFACE_ATTR_NDP_FROM_LINK_LOCAL] = { .name = "ndp_from_link_local", .type = BLOBMSG_TYPE_BOOL },
```
Alphix, line 175: This new option should be documented in the `README`

src/odhcpd.c
```diff
@@ -254,6 +266,21 @@ static int odhcpd_get_linklocal_interface_address(int ifindex, struct in6_addr *
 	return ret;
 }
 
+/* Get link-local address for NDP operations - prefer this over ULA for macOS compatibility */
```
Alphix: ...and RFC4861 compliance...but I don't think this comment is necessary in `odhcpd.c`, it's just a function for getting a LL address...and some other part of the code might use it for other purposes than RFC4861/macOS compliance...

src/odhcpd.c
```diff
@@ -189,6 +189,14 @@ int odhcpd_get_flags(const struct interface *iface)
...
 		struct iovec *iov, size_t iov_len,
 		const struct interface *iface)
+{
+	return odhcpd_send_with_src(socket, dest, iov, iov_len, iface, NULL);
+}
+
+/* Forwards a packet on a specific interface with optional source address */
+ssize_t odhcpd_send_with_src(int socket, struct sockaddr_in6 *dest,
+		struct iovec *iov, size_t iov_len,
+		const struct interface *iface, const struct in6_addr *src_addr)
```
Alphix, line 212: Nitpick: I'd suggest reversing the order of these two functions, so the specific comes first and the generic (which calls the specific) comes second.

src/odhcpd.h
```diff
@@ -437,8 +438,13 @@ void odhcpd_process(struct odhcpd_event *event);
 ssize_t odhcpd_send(int socket, struct sockaddr_in6 *dest,
 		struct iovec *iov, size_t iov_len,
 		const struct interface *iface);
+ssize_t odhcpd_send_with_src(int socket, struct sockaddr_in6 *dest,
+		struct iovec *iov, size_t iov_len,
+		const struct interface *iface, const struct in6_addr *src_addr);
```
Alphix: And reverse the order here as well

src/odhcpd.c
```diff
@@ -254,6 +266,21 @@ static int odhcpd_get_linklocal_interface_address(int ifindex, struct in6_addr *
...
+		if (IN6_IS_ADDR_LINKLOCAL(&iface->addr6[i].addr.in6)) {
+			*addr = iface->addr6[i].addr.in6;
+			return 0;
+		}
+	}
+
+	/* Fallback to the existing method */
+	return odhcpd_get_linklocal_interface_address(iface->ifindex, addr);
+}
+
```
Alphix: Question: (and this is based on a very quick reading of the code, so I might have missed something) should we cache the result of `odhcpd_get_linklocal_interface_address()` in `struct interface` so that the latter function only gets called once per interface?

`odhcpd_get_linklocal_interface_address()` does a whole `socket()`/`connect()`/`close()` dance, and I don't think we want that to be done every time a NDP operation is performed (do we expect the LL address obtained via that method to really change, ever?)

src/ndp.c
```diff
@@ -319,7 +329,15 @@ static void send_na(struct in6_addr *to_addr,
...
-	odhcpd_send(iface->ndp_ping_fd, &dest, &iov, 1, iface);
+	/* Use link-local address as source for RFC 4861 compliance and macOS compatibility.
+	 * RFC 4861 section 4.2 mandates that Neighbor Advertisement source address MUST be
+	 * the link-local address assigned to the interface from which this message is sent. */
+	if (iface->ndp_from_link_local && odhcpd_get_interface_linklocal_addr(iface, &src_addr) == 0) {
+		odhcpd_send_with_src(iface->ndp_ping_fd, &dest, &iov, 1, iface, &src_addr);
+	} else {
+		/* Fall back to default behavior if no link-local address is available or flag is disabled */
+		odhcpd_send(iface->ndp_ping_fd, &dest, &iov, 1, iface);
+	}
```
Alphix: Question: this is duplicated from `ping6()`. Is it worth moving this into yet another function? Like `odhcp_try_send_with_src()` and moving the detailed comment about the rationale into that function...then this and the corresponding change to `ping6()` will be reduced to a single line and we have just one place where the behaviour can be explained in detail instead of repeating the rationale...?

(and the reference to RFC4861 would be even better if it was a reference to RFC4861, §4.2 - digging around in 100-page long RFCs can be frustrating)

src/odhcpd.c
```diff
@@ -254,6 +266,21 @@ static int odhcpd_get_linklocal_interface_address(int ifindex, struct in6_addr *
 	return ret;
 }
 
+/* Get link-local address for NDP operations - prefer this over ULA for macOS compatibility */
+int odhcpd_get_interface_linklocal_addr(const struct interface *iface, struct in6_addr *addr)
```
Alphix: With this change, we'd have two functions:
`odhcpd_get_interface_linklocal_addr()`
`odhcpd_get_linklocal_interface_address()`

I think that might be pretty confusing. `odhcpd_get_linklocal_interface_address()` currently has a single user (`odhcpd_get_interface_dns_addr()` in `odhcpd.c`), couldn't the two functions be merged into one? (i.e. inline `odhcpd_get_linklocal_interface_address()` into `odhcpd_get_interface_linklocal_addr()` and change `odhcpd_get_interface_dns_addr()` to use it). Or would that break something?

Alphix (changes requested):

Alphix:
> Given that the behaviour is RFC mandated, I think we can set the default to true, and now we have the flag, anyone experiencing problems (unlikely) can manually set the flag to false.

I agree. And once this behaviour has been part of an OpenWrt release, assuming it causes no issues, I guess the flag might be removed and this made mandatory...

src/odhcpd.c
systemcrash: Reasonable.

src/odhcpd.c
systemcrash: Caching would seem to be the better route, despite its evils. IIUC, LL addresses are not required to be SLAAC generated, so it seems it won't change frequently.

stephengroat:
I tried to address all the comments, please let me know if i'm missing anything

I left the comment unresolved so that @Alphix can resolve as they see fit.

src/odhcpd.c
```diff
@@ -228,32 +232,79 @@ ssize_t odhcpd_send(int socket, struct sockaddr_in6 *dest,
...
+ssize_t odhcpd_send(int socket, struct sockaddr_in6 *dest,
+		struct iovec *iov, size_t iov_len,
+		const struct interface *iface)
+{
+	return odhcpd_send_with_src(socket, dest, iov, iov_len, iface, NULL);
+}
 
-static int odhcpd_get_linklocal_interface_address(int ifindex, struct in6_addr *lladdr)
+
+int odhcpd_get_interface_linklocal_addr(const struct interface *iface, struct in6_addr *addr)
```
Alphix: Right now the function signature is:

```C
int odhcpd_get_interface_linklocal_addr(const struct interface *iface, struct in6_addr *addr)
```

Which isn't true because the `struct *iface` is modified by the function, so change it to
```C
int odhcpd_get_interface_linklocal_addr(struct interface *iface, struct in6_addr *addr)
```

And then you can remove all casts later, like these:
```C
((struct interface *)iface)->cached_linklocal_addr = *addr;
((struct interface *)iface)->cached_linklocal_valid = true;
```

Alphix (changes requested):

src/odhcpd.c
```diff
@@ -228,32 +232,79 @@ ssize_t odhcpd_send(int socket, struct sockaddr_in6 *dest,
...
+		}
+		close(sock);
 	}
 
-	close(sock);
+	if (ret == 0) {
+		/* Cache the result for future use */
+		((struct interface *)iface)->cached_linklocal_addr = *addr;
+		((struct interface *)iface)->cached_linklocal_valid = true;
+	}
```
Alphix: If you move this into the previous `if` statement, you can drop the `ret` variable completely and just return `0` or `-1` in the right places.

Alphix (changes requested):

Alphix:
> I tried to address all the comments, please let me know if i'm missing anything

Thank you 👍🏻 

> I left the comment unresolved so that @Alphix can resolve as they see fit.

I don't have write access so I can't resolve comments. Consider everything but the last two nits I highlighted a few minutes ago as resolved.

Noltari:
@Alphix I marked all the comments as resolved except the last two.

@stephengroat you're missing the SoB line on your commit with your real name according to the OpenWrt rules:
https://openwrt.org/submitting-patches#sign_your_work

systemcrash:
ping @stephengroat for your SOB line please

systemcrash:
Oh!

systemcrash:
You can re-open it since it's so close to the finish line @stephengroat

stephengroat:
sorry for the close, bad push when trying to merge

systemcrash:
no merge commits - just rebase :)

stephengroat:
sorry, i know, not enough sleep (newborn), i tried again

stephengroat:
appreciate the help & thoughtful comment and sorry for the issues, i think i've addressed everything

systemcrash:
OK! @Noltari  ?

Alphix:
> sorry, i know, not enough sleep (newborn), i tried again

Been there, know that. Patch looks good to me

Noltari:
@stephengroat I've got a combination of a 1 year old daughter and a really bad time at work, so I'm getting home at 7-8pm. I feel your pain...
@Alphix @systemcrash please forgive me if I can't follow up by merging everything but it's being a really bad month for me at work...

Alphix:
> @Alphix @systemcrash please forgive me if I can't follow up by merging everything but it's being a really bad month for me at work...

No need to apologise...you've done a great job...take your time, focus on what's important

systemcrash:
No problem - just hoping to take advantage of momentum. Focus on the fam.

Noltari (approved):

Noltari:
Merged, thanks @stephengroat

willmo:
I notice that RFC 4861 only _mandates_ this for RAs and Redirects (right?), and this PR doesn't change `send_router_advert()`. I haven't checked the actual behavior with RAs now but is that an oversight?

Alphix:
> I notice that RFC 4861 only _mandates_ this for RAs and Redirects (right?), and this PR doesn't change `send_router_advert()`. I haven't checked the actual behavior with RAs now but is that an oversight?

@willmo I think it's an oversight, though more of a correctness issue. The router solicitations will usually originate from link-local addresses (or the unspecified IPv6 address) so the reply will also be from the link-local address. But if the RS source addr is not link-local, it does look like `odhcpd` will reply with a non-link-local address as the src address...I'll create a PR for that...(and for the proxy case as well)

PR #209, opened 2023-12-01
odhcpd: improve RFC 9096 compliance

skbeh:
and allow configuring upper limit for preferred and valid lifetime.

PR #245, opened 2025-08-07
[RFC9096] Improve section 3.5 SLAAC compliance

Noltari:
[RFC9096 § 3.5](https://datatracker.ietf.org/doc/html/rfc9096#section-3.5)

When a CE router provides LAN-side address-configuration information
via SLAAC:

*  A CE router sending RAs that advertise prefixes belonging to a
   dynamically learned prefix (e.g., via DHCPv6-PD) SHOULD record, on
   stable storage, the list of prefixes being advertised via PIOs on
   each network segment and the state of the "A" and "L" flags of the
   corresponding PIOs.

*  Upon changes to the advertised prefixes, and after bootstrapping,
   the CE router advertising prefix information via SLAAC proceeds as
   follows:

   -  Any prefixes that were previously advertised by the CE router
      via PIOs in RA messages, but that have now become stale, MUST
      be advertised with PIOs that have the "Valid Lifetime" and the
      "Preferred Lifetime" set to 0 and the "A" and "L" bits
      unchanged.

   -  The aforementioned advertisements MUST be performed for at
      least the "Valid Lifetime" previously employed for such
      prefixes.

---

The current implementation of this PR should be set with a `piofile` in `/tmp` by default (for example `/tmp/odhcpd-rfc9096`) and then let the users decide if they want to use a file in the flash to comply with RFC9096 on reboots.
Right now, there's a flash write every 12h per interface with IPv6 RA enabled. This is due to the fact that even though the prefix delegated has a validity of 24h, it's renewed at half of that validity.
Different interfaces renewals are spread in time and therefore can't be grouped into a single write with the current implementation:

```
Wed Sep  3 17:51:39 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43193 -> 86393)
Wed Sep  3 17:51:44 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Wed Sep  3 17:55:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (42971 -> 86171)
Wed Sep  3 17:55:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

```
Thu Sep  4 05:51:55 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43177 -> 86380)
Thu Sep  4 05:52:00 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Thu Sep  4 05:54:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (43030 -> 86233)
Thu Sep  4 05:54:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

```
Thu Sep  4 17:52:03 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43172 -> 86375)
Thu Sep  4 17:52:08 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Thu Sep  4 17:59:23 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (42732 -> 85935)
Thu Sep  4 17:59:28 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

```
Fri Sep  5 05:52:12 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43165 -> 86367)
Fri Sep  5 05:52:17 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Fri Sep  5 05:54:09 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (43049 -> 86251)
Fri Sep  5 05:54:16 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -4,13 +4,16 @@
 #include <arpa/inet.h>
 #include <unistd.h>
 #include <libgen.h>
+#include <linux/magic.h>
 #include <net/if.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/vfs.h>
 #include <syslog.h>
 
 #include <uci.h>
 #include <uci_blob.h>
+#include <json-c/json.h>
 #include <libubox/utils.h>
 #include <libubox/avl.h>
 #include <libubox/avl-cmp.h>
@@ -31,7 +34,10 @@ struct vlist_tree leases = VLIST_TREE_INIT(leases, lease_cmp, lease_update, true
 AVL_TREE(interfaces, avl_strcmp, false, NULL);
 struct config config = {.legacy = false, .main_dhcpv4 = false,
 			.dhcp_cb = NULL, .dhcp_statefile = NULL, .dhcp_hostsfile = NULL,
-			.log_level = LOG_WARNING};
+			.ra_piofile = NULL, .log_level = LOG_WARNING};
+
+bool ra_pio_sync = false;
+bool ra_pio_update = false;
 
 #define START_DEFAULT	100
 #define LIMIT_DEFAULT	150
@@ -200,6 +206,7 @@ enum {
 	ODHCPD_ATTR_LEASETRIGGER,
 	ODHCPD_ATTR_LOGLEVEL,
 	ODHCPD_ATTR_HOSTSFILE,
+	ODHCPD_ATTR_PIOFILE,
 	ODHCPD_ATTR_MAX
 };
 
@@ -210,6 +217,7 @@ static const struct blobmsg_policy odhcpd_attrs[ODHCPD_ATTR_MAX] = {
 	[ODHCPD_ATTR_LEASETRIGGER] = { .name = "leasetrigger", .type = BLOBMSG_TYPE_STRING },
 	[ODHCPD_ATTR_LOGLEVEL] = { .name = "loglevel", .type = BLOBMSG_TYPE_INT32 },
 	[ODHCPD_ATTR_HOSTSFILE] = { .name = "hostsfile", .type = BLOBMSG_TYPE_STRING },
+	[ODHCPD_ATTR_PIOFILE] = { .name = "piofile", .type = BLOBMSG_TYPE_STRING },
 };
 
 const struct uci_blob_param_list odhcpd_attr_list = {
@@ -317,7 +325,7 @@ static void close_interface(struct interface *iface)
 	clean_interface(iface);
 	free(iface->addr4);
 	free(iface->addr6);
-	free(iface->invalid_addr6);
+	free(iface->pio);
 	free(iface->ifname);
 	free(iface);
 }
@@ -386,6 +394,16 @@ static void set_config(struct uci_section *s)
 	if ((c = tb[ODHCPD_ATTR_HOSTSFILE])) {
 		free(config.dhcp_hostsfile);
 		config.dhcp_hostsfile = strdup(blobmsg_get_string(c));
+	}
+
+	if ((c = tb[ODHCPD_ATTR_PIOFILE])) {
+		struct statfs ra_piofile_info;
+
+		free(config.ra_piofile);
+		config.ra_piofile = strdup(blobmsg_get_string(c));
+
+		statfs(config.ra_piofile, &ra_piofile_info);
+		ra_pio_sync = (ra_piofile_info.f_type == TMPFS_MAGIC);
 	}
 
 	if ((c = tb[ODHCPD_ATTR_LEASETRIGGER])) {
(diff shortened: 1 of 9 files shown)


Noltari:
See also: https://github.com/openwrt/netifd/pull/53

Noltari:
Converted to draft since I tested it and it doesn’t work as expected (apparrently stale address are present on the valid array instead of the invalid array).

Noltari:
Should be fixed now, although I still don't fully get why stale addresses aren't present on `invalid_addr6`...

Noltari:
After some additional changes this should be now fully compliant with RFC9096 section 3.5 only for SLAAC.
I'm now testing it on my own home network, and it looks good so far.

Pending:
- Logging has to be reduced and/or demoted.
- Split into smaller patches?

Noltari:
Updated with ubus compatibility:
```
root@rpi5:~# ubus call dhcp ipv6slaac
{
        "interfaces": {
                "br-lan": [
                        {
                                "lifetime": 44068,
                                "prefix": "2a02:9142:XXXX:8d00::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 44068,
                                "prefix": "2a02:9142:XXXX:8f00::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 62107,
                                "prefix": "2a02:9142:XXXX:9000::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 64853,
                                "prefix": "2a02:9142:XXXX:9100::/64",
                                "stale": false
                        }
                ],
                "eth1.10": [
                        {
                                "lifetime": 44068,
                                "prefix": "2a02:9142:XXXX:8d10::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 44068,
                                "prefix": "2a02:9142:XXXX:8f10::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 62107,
                                "prefix": "2a02:9142:XXXX:9010::/64",
                                "stale": true
                        },
                        {
                                "lifetime": 64853,
                                "prefix": "2a02:9142:XXXX:9110::/64",
                                "stale": false
                        }
                ]
        }
}
```

Noltari:
Multiple improvements and optimizations added which reduce the number of file writes as much as possible.
The debug logging has been removed and now only PIO add, load, remove, renew and file writes are logged as warnings.

Therefore, this should be now ready for review. Feel free to suggest changes and improvements :)

```
root@rpi5:~# ubus call dhcp ipv6ra
{
        "interfaces": {
                "br-lan": [
                        {
                                "lifetime": 59512,
                                "prefix": "2a02:9142:XXXX:9100::/64",
                                "stale": true
                        },
                        {
                                "prefix": "fd91::/64",
                                "stale": false
                        },
                        {
                                "lifetime": 85680,
                                "prefix": "2a02:9142:XXXX:a200::/64",
                                "stale": false
                        }
                ],
                "eth1.10": [
                        {
                                "lifetime": 59512,
                                "prefix": "2a02:9142:XXXX:9110::/64",
                                "stale": true
                        },
                        {
                                "prefix": "fd91:0:0:10::/64",
                                "stale": false
                        },
                        {
                                "lifetime": 85680,
                                "prefix": "2a02:9142:XXXX:a210::/64",
                                "stale": false
                        }
                ]
        }
}
```

src/router.c
```diff
@@ -888,9 +1014,12 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 
 	syslog(LOG_NOTICE, "Sending a RA on %s", iface->name);
 
-	if (odhcpd_send(iface->router_event.uloop.fd, &dest, iov, ARRAY_SIZE(iov), iface) > 0)
+	if (odhcpd_send(iface->router_event.uloop.fd, &dest, iov, ARRAY_SIZE(iov), iface) > 0) {
 		iface->ra_sent++;
 
+		uloop_timeout_set(&ra_pio_timeout, 5000);
```
Noltari: With this `uloop_timeout` I intended to avoid writing the RA PIO file once per interface, but it looks like it's just delaying the write by 5s.
Also, without this `uloop_timeout`, the RA was sent immediately for each interface, but now it seems to be delayed by 90s...

Noltari: Nevermind, it's just the kernel not refreshing the interfaces at the same time...

```
Wed Sep  3 05:51:38 2025 daemon.warn odhcpd[19680]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43194 -> 86394)
Wed Sep  3 05:51:43 2025 daemon.warn odhcpd[19680]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Wed Sep  3 05:53:08 2025 daemon.warn odhcpd[19680]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (43104 -> 86304)
Wed Sep  3 05:53:13 2025 daemon.warn odhcpd[19680]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

```
Wed Sep  3 17:51:39 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43193 -> 86393)
Wed Sep  3 17:51:44 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Wed Sep  3 17:55:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (42971 -> 86171)
Wed Sep  3 17:55:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

```
Thu Sep  4 05:51:55 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2a02:9142:XXXX:a200::/64 (43177 -> 86380)
Thu Sep  4 05:52:00 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
Thu Sep  4 05:54:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2a02:9142:XXXX:a210::/64 (43030 -> 86233)
Thu Sep  4 05:54:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-rfc9096
```

Therefore, the current implementation implies 1 flash write per interface every 12h.

Noltari:
Ping @dedeckeh @Ansuel

src/config.c
```diff
@@ -1663,6 +1683,227 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+{
+	json_object *json = json_object_from_file(config.ra_piofile);
+
+	if (json) {
+		json_object *time_json = json_object_object_get(json, JSON_TIME);
+		time_t time_ref_json = (time_t) json_object_get_int64(time_json);
+		json_object *interfaces_json = json_object_object_get(json, JSON_INTERFACES);
+		time_t time_ref, now;
+		uint8_t new_md5[16];
+		md5_ctx_t md5;
```
efahl: Is there a missing `md5_begin` call here?  There's one in the function just below, so looks odd that none here.

efahl:
LGTM (but I didn't test it at all, just read through the code).

It would be nice if the commit message mentioned 
 1) the tmp file name with a one-liner of what it contains,
 2) the new ubus call, and 
 3) had at least a cursory mention about flash usage.

(So we don't have to chase through the code or back to this PR to find that out.)

src/config.c
Noltari: Nice catch!

Noltari: Fixed in 123877f4128e133deb823f850c59845e008eac24

Noltari:
> LGTM (but I didn't test it at all, just read through the code).
> 
> It would be nice if the commit message mentioned
> 
> 1. the tmp file name with a one-liner of what it contains,
> 2. the new ubus call, and
> 3. had at least a cursory mention about flash usage.
> 
> (So we don't have to chase through the code or back to this PR to find that out.)

Added in 123877f4128e133deb823f850c59845e008eac24

GoetzGoerisch:
@Noltari Thank you for doing this! Really much appreciated.

One question which I could not figure out reviewing this.

commit description explains as default this should be set:
`uci set dhcp.odhcpd.piofile="/tmp/odhcpd-rfc9096"`

Does this needs to be added as a default file when generating the initial installation? That this is setup on new installs. Further more I would like to suggest naming is `odhcpd-piofile`.

May I suggest to change the prefixs in the commit message to the documentation prefix?

Noltari:
> @Noltari Thank you for doing this! Really much appreciated.
> 
> One question which I could not figure out reviewing this.
> 
> commit description explains as default this should be set: `uci set dhcp.odhcpd.piofile="/tmp/odhcpd-rfc9096"`
> 
> Does this needs to be added as a default file when generating the initial installation? That this is setup on new installs. Further more I would like to suggest naming is `odhcpd-piofile`.

If this is merged and odhcpd package is updated on the openwrt repo, it will be added to the default configuration with any value, it doesn't have to be the same as the one suggested in this commit description since it's just an example.
However, I'll change the commit description to `uci set dhcp.odhcpd.piofile="/tmp/odhcpd-piofile"`.

> May I suggest to change the prefixs in the commit message to the documentation prefix?

What do you mean?

GoetzGoerisch:
> If this is merged and odhcpd package is updated on the openwrt repo, it will be added to the default configuration with any value, it doesn't have to be the same as the one suggested in this commit description since it's just an example. However, I'll change the commit description to `uci set dhcp.odhcpd.piofile="/tmp/odhcpd-piofile"`.

Thank you.


> > May I suggest to change the prefixs in the commit message to the documentation prefix?
> 
> What do you mean?

RFC3849 defines an IPv6 prefix for documentation. This commit message is documentation from my PoV.

Therefore my suggestion is:

```text
When a CE router provides LAN-side address-configuration information
via SLAAC:

*  A CE router sending RAs that advertise prefixes belonging to a
   dynamically learned prefix (e.g., via DHCPv6-PD) SHOULD record, on
   stable storage, the list of prefixes being advertised via PIOs on
   each network segment and the state of the "A" and "L" flags of the
   corresponding PIOs.

*  Upon changes to the advertised prefixes, and after bootstrapping,
   the CE router advertising prefix information via SLAAC proceeds as
   follows:

   -  Any prefixes that were previously advertised by the CE router
      via PIOs in RA messages, but that have now become stale, MUST
      be advertised with PIOs that have the "Valid Lifetime" and the
      "Preferred Lifetime" set to 0 and the "A" and "L" bits
      unchanged.

   -  The aforementioned advertisements MUST be performed for at
      least the "Valid Lifetime" previously employed for such
      prefixes.

This should be enabled by default with a file in "/tmp" to avoid forcing
flash writes for all the users and at least handle the stale PIOs on PPPoe
reconnections. For example:
  uci set dhcp.odhcpd.piofile="/tmp/odhcpd-piofile"
  uci commit dhcp

There's a new ubus call which allows getting all the IPv6 RA PIOs, with
information about their lifetime and stale status:
  ubus call dhcp ipv6ra
  {
    "interfaces": {
      "br-lan": [
        {
          "lifetime": 40123,
          "prefix": "2001:db8:9a00::/64",
          "stale": true
        },
        {
          "prefix": "fd91::/64",
          "stale": false
        },
        {
          "lifetime": 83866,
          "prefix": "2001:db8:9b00::/64",
          "stale": false
        }
      ],
      "eth1.10": [
        {
          "lifetime": 40123,
          "prefix": "2001:db8:9a10::/64",
          "stale": true
        },
        {
          "prefix": "fd91:0:0:10::/64",
          "stale": false
        },
        {
          "lifetime": 83866,
          "prefix": "2001:db8:9b10::/64",
          "stale": false
        }
      ]
    }
  }

The current implementation implies a flash write every 12h per interface
with IPv6 RA enabled. This is due to the fact that even though the
delegated prefix has a validity of 24h, it's renewed at half of that:
  Wed Sep  3 17:51:39 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2001:db8:9a00::/64 (43193 -> 86393)
  Wed Sep  3 17:51:44 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-piofile
  Wed Sep  3 17:55:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2001:db8:9a10::/64 (42971 -> 86171)
  Wed Sep  3 17:55:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-piofile

  Thu Sep  4 05:51:55 2025 daemon.warn odhcpd[31802]: rfc9096: br-lan: renew 2001:db8:9a00::/64 (43177 -> 86380)
  Thu Sep  4 05:52:00 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-piofile
  Thu Sep  4 05:54:21 2025 daemon.warn odhcpd[31802]: rfc9096: eth1.10: renew 2001:db8::9a10::/64 (43030 -> 86233)
  Thu Sep  4 05:54:26 2025 daemon.warn odhcpd[31802]: rfc9096: update piofile: /etc/odhcpd-piofile
Since some ISPs assign static or semi-static prefixes to their customers,
it should be up to the user to enable flash writes for this when needed
(ISPs which assign dynamic IPv6 prefixes).
```

efahl:
> "prefix": "2001:db8:9a00::/64",

Ah, good point.  Make all the GUAs unrouteable so that if someone takes things literally, it will break locally and immediately, so as to minimize time spent debugging.

```
$ ipv6decode.py -f short -v 2001:db8:9a00::/64
2001:db8:9a00::/64 ... Global Unicast (GUA) .............................................................. 2000::/3
                       | APNIC-AP: ORG-APNI3-AP - Asia Pacific Network Information Center, Pty. Ltd. ..... 2001:c00::/23
                       | | Documentation (never route) ................................................... 2001:db8::/32
                       | | | IID: 0000:0000:0000:0000 (64-bits) .......................................... 2001:0db8:9a00:0000:0000:0000:0000:0000/64
Reference:
http://www.rfc-editor.org/rfc/rfc3849 - Documentation (never route)
```

Noltari:
> RFC3849 defines an IPv6 prefix for documentation. This commit message is documentation from my PoV.
> 
> Therefore my suggestion is:

Thanks @GoetzGoerisch, I've modified the commit description with your suggestion.

I've also improved the piofile loading robustness, including proper checks for loaded data.

GoetzGoerisch (approved):
LGTM, thank you!

Noltari:
I want to merge https://github.com/openwrt/odhcpd/pull/209 first since I want to work on an alternative version of this PR in order to see if I can achieve less flash writes.
Maybe adding the delegated prefixes to the config with a preset lifetime of 90m and marking them as stale with a starting timestamp whenever they become stale.
That way we would only write the config file whenever a new prefix is delegated instead of every 12h.

systemcrash:
@Noltari - this in general is great. 


Regarding the file writes - writing to permanent storage only at daemon shutdown (behind an option flag?) seems good `void config_shutdown(void)`.


Since you're in the area of adjusting lifetimes to zero, which is the solution described in "Multihoming" [RFC8475](https://www.rfc-editor.org/rfc/rfc8475.html) I wonder whether it makes sense to trigger these adjustments also based on link state changes, so that multiple IPv6 routes case is also taken care of. I have not verified whether this is directly handled here. Maybe it's a bridge too far for this PR 🙈 

Basically:

Link down -> RA PIO preferred lifetime 0
Link up -> RA PIO preferred lifetime default


It's a common problem encountered in the forums. 


Excerpt of overview from https://www.rfc-editor.org/rfc/rfc8475.html#section-3.1.2
```
...Stateless
   Address Autoconfiguration (SLAAC) [RFC4862] and Router Advertisements
   (RAs) [RFC4861] are the most suitable mechanisms for signaling
   network topology changes to hosts, thereby influencing the source
   address selection.  Sending an RA to change the preferred lifetime
   for a given prefix provides the following functionality:

   o  Deprecating addresses by sending an RA with preferred_lifetime set
      to 0 in the corresponding Prefix Information option (PIO)
      [RFC4861].  This indicates to hosts that addresses from that
      prefix should not be used;

   o  Making a previously unused (deprecated) prefix usable again by
      sending an RA containing a PIO with nonzero preferred lifetime.
      This indicates to hosts that addresses from that prefix can be
      used again.

   It should be noted that only the preferred lifetime for the affected
   prefix needs to be changed.  As the goal is to influence the source
   address selection algorithm on hosts rather than prevent them from
   forming addresses from a specific prefix, the valid lifetime should
   not be changed.  Actually, changing the valid lifetime would not even
   be possible for unauthenticated RAs (which is the most common
   deployment scenario), because [Section 5.5.3 of [RFC4862]] prevents
   hosts from setting the valid lifetime for addresses to zero unless
   RAs are authenticated.

   To provide the desired functionality, first-hop routers are required
   to:

   o  Send RAs triggered by defined event policies in response to an
      uplink status change event; and

   o  While sending periodic or solicited RAs, set the value in the
      given RA field (e.g., PIO preferred lifetime) based on the uplink
      status.

   The exact definition of the "uplink status" depends on the network
   topology and may include conditions like:

   o  Uplink interface status change;

   o  Presence of a particular route in the routing table;

   o  Presence of a particular route with a particular attribute (next
      hop, tag, etc.) in the routing table;

   o  Protocol adjacency change.
```

Noltari:
> @Noltari - this in general is great. 

Thanks! 😊

> Regarding the file writes - writing to permanent storage only at daemon shutdown (behind an option flag?) seems good `void config_shutdown(void)`.

But we also want to make sure to keep the announced prefixes if the router crashes or reboots due to a power outage...

> Since you're in the area of adjusting lifetimes to zero, which is the solution described in "Multihoming" [RFC8475](https://www.rfc-editor.org/rfc/rfc8475.html) I wonder whether it makes sense to trigger these adjustments also based on link state changes, so that multiple IPv6 routes case is also taken care of. I have not verified whether this is directly handled here. Maybe it's a bridge too far for this PR 🙈 
> 
> 
> 
> Basically:
> 
> 
> 
> Link down -> RA PIO preferred lifetime 0
> 
> Link up -> RA PIO preferred lifetime default
> 
> It's a common problem encountered in the forums. 
> 
> Excerpt of overview from https://www.rfc-editor.org/rfc/rfc8475.html#section-3.1.2

I believe that can be addressed on a subsequent PR because the current PR is already big enough.

Noltari:
Superseded by alternative implementation with less flash writes -> https://github.com/openwrt/odhcpd/pull/256

PR #251, opened 2025-09-16
odhcpd: support multiple per-client MAC addresses

### LINKED ISSUE #221: Support `list mac` in /etc/config/dhcp

### LINKED ISSUE #9598: odhcpd: Static leases with multiple MAC addresses get skipped when configuring odhcpd

Alphix:
Allow the configuration of multiple per-client DHCPv4 MAC addresses. This is for feature parity with dnsmasq.

Closes: https://github.com/openwrt/odhcpd/issues/221
Closes: https://github.com/openwrt/openwrt/issues/9598

### DIFF
--- a/README
+++ b/README
@@ -161,13 +161,13 @@ ntp			list	<local address>		NTP servers to announce
 
 
 Sections of type host (static leases)
-Option		Type	Default			Description
-ip		string				IP-Address to lease
-mac		string				MAC-address
-duid		string				DUID in base16
-hostid		string				IPv6 host identifier
-name		string				Hostname
-leasetime	string				DHCPv4/v6 leasetime
+Option		Type		Default			Description
+ip		string		(none)			IPv4 host address
+mac		list|string	(none)			Hexadecimal MAC address(es)
+duid		string		(none)			Hexadecimal DUID
+hostid		string		(none)			IPv6 host identifier
+name		string		(none)			Hostname
+leasetime	string		(none)			DHCPv4/v6 leasetime
 
 Sections of type boot6
 Option		Type	Required		Description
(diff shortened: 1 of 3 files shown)


Alphix:
I'm well aware that there's already a PR for this (#233), but I'd argue that this version is more idiomatic, more concise, and it also avoids additional `malloc()`s

thess:
Tested on Linksys E8450. I will merge this version of the fix, if no objections, after some additional testing.
Closing #233 for now. Thanks to both of you -- @YodaDaCoda and @Alphix

Alphix:
> Tested on Linksys E8450. I will merge this version of the fix, if no objections, after some additional testing. Closing #233 for now. Thanks to both of you -- @YodaDaCoda and @Alphix

@thess thanks for the prompt reply. For the record, I've tested it on x86 (not running OpenWRT) and on a Turris Omnia running OpenWRT 24.10.2

PR #252, opened 2025-09-16
odhcpd: bump minimum cmake version

Alphix:
This removes warnings with more recent versions of cmake.

3.13 is already required by e.g. ubus, so it seems like a reasonable minimum.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.13)
 cmake_policy(SET CMP0015 NEW)
 
 # Project Definition


CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.13)
 cmake_policy(SET CMP0015 NEW)
+cmake_policy(SET CMP0021 NEW)
```
Noltari: You are also enabling `CMP0021` here, is it needed for the version bump?
Can we add it on a separate commit/PR and justify the reason for adding it?

Alphix: You're right, I think I added it to fix a non-issue (me misusing `cmake` essentially). I'll just remove that line for now.

Noltari (approved):

PR #256, opened 2025-09-22
[RFC9096] Improve section 3.5 SLAAC compliance (alternative implementation)

Noltari:
[RFC9096 § 3.5](https://datatracker.ietf.org/doc/html/rfc9096#section-3.5)

When a CE router provides LAN-side address-configuration information
via SLAAC:

*  A CE router sending RAs that advertise prefixes belonging to a
   dynamically learned prefix (e.g., via DHCPv6-PD) SHOULD record, on
   stable storage, the list of prefixes being advertised via PIOs on
   each network segment and the state of the "A" and "L" flags of the
   corresponding PIOs.

*  Upon changes to the advertised prefixes, and after bootstrapping,
   the CE router advertising prefix information via SLAAC proceeds as
   follows:

   -  Any prefixes that were previously advertised by the CE router
      via PIOs in RA messages, but that have now become stale, MUST
      be advertised with PIOs that have the "Valid Lifetime" and the
      "Preferred Lifetime" set to 0 and the "A" and "L" bits
      unchanged.

   -  The aforementioned advertisements MUST be performed for at
      least the "Valid Lifetime" previously employed for such
      prefixes.

This should be enabled by default with a folder in "/tmp" to avoid forcing
flash writes for all the users and at least handle the stale PIOs on PPPoe
reconnections. For example:
```
  uci set dhcp.odhcpd.piofolder="/tmp/odhcpd-piofolder"
  uci commit dhcp
```

There's a new ubus call which allows getting all the IPv6 RA PIOs, with
information about their lifetime and stale status:
```
  ubus call dhcp ipv6ra
  {
    "interfaces": {
      "br-lan": [
        {
          "lifetime": 5390,
          "prefix": "2001:db8:0:100::/64",
          "stale": true
        },
        {
          "prefix": "2001:db8:0:200::/64",
          "stale": false
        },
        {
          "prefix": "fd00::/64",
          "stale": false
        }
      ],
      "eth1.10": [
        {
          "lifetime": 5390,
          "prefix": "2001:db8:0:110::/64",
          "stale": true
        },
        {
          "prefix": "2001:db8:0:210::/64",
          "stale": false
        },
        {
          "prefix": "fd00:0:0:10::/64",
          "stale": false
        }
      ]
    }
  }
```

The current implementation performs a flash write per interface whenever
a new prefix is added or an existing prefix becomes stale.
Since some ISPs assign static or semi-static prefixes to their customers,
it should be up to the user to enable flash writes when needed.

### DIFF
--- a/README
+++ b/README
@@ -67,6 +67,8 @@ leasefile	string				DHCP/v6 lease/hostfile
 leasetrigger	string				Lease trigger script
 hostsfile	string				DHCP/v6 hostfile
 loglevel	integer 6			Syslog level priority (0-7)
+piofolder	string				Folder to store IPv6 prefix information (to
+						detect stale prefixes, see RFC9096, §3.5)
 
 
 Sections of type dhcp (configure DHCP / DHCPv6 / RA / NDP service)
--- a/src/config.c
+++ b/src/config.c
@@ -1,3 +1,4 @@
+#include <errno.h>
 #include <fcntl.h>
 #include <resolv.h>
 #include <signal.h>
@@ -11,6 +12,7 @@
 
 #include <uci.h>
 #include <uci_blob.h>
+#include <json-c/json.h>
 #include <libubox/utils.h>
 #include <libubox/avl.h>
 #include <libubox/avl-cmp.h>
@@ -29,9 +31,16 @@ static void lease_update(struct vlist_tree *tree, struct vlist_node *node_new,
 
 struct vlist_tree leases = VLIST_TREE_INIT(leases, lease_cmp, lease_update, true, false);
 AVL_TREE(interfaces, avl_strcmp, false, NULL);
-struct config config = {.legacy = false, .main_dhcpv4 = false,
-			.dhcp_cb = NULL, .dhcp_statefile = NULL, .dhcp_hostsfile = NULL,
-			.log_level = LOG_WARNING};
+struct config config = {
+	.legacy = false,
+	.main_dhcpv4 = false,
+	.dhcp_cb = NULL,
+	.dhcp_statefile = NULL,
+	.dhcp_hostsfile = NULL,
+	.ra_piofolder = NULL,
+	.ra_piofolder_fd = -1,
+	.log_level = LOG_WARNING,
+};
 
 #define START_DEFAULT	100
 #define LIMIT_DEFAULT	150
@@ -200,6 +209,7 @@ enum {
 	ODHCPD_ATTR_LEASETRIGGER,
 	ODHCPD_ATTR_LOGLEVEL,
 	ODHCPD_ATTR_HOSTSFILE,
+	ODHCPD_ATTR_PIOFOLDER,
 	ODHCPD_ATTR_MAX
 };
 
@@ -210,6 +220,7 @@ static const struct blobmsg_policy odhcpd_attrs[ODHCPD_ATTR_MAX] = {
 	[ODHCPD_ATTR_LEASETRIGGER] = { .name = "leasetrigger", .type = BLOBMSG_TYPE_STRING },
 	[ODHCPD_ATTR_LOGLEVEL] = { .name = "loglevel", .type = BLOBMSG_TYPE_INT32 },
 	[ODHCPD_ATTR_HOSTSFILE] = { .name = "hostsfile", .type = BLOBMSG_TYPE_STRING },
+	[ODHCPD_ATTR_PIOFOLDER] = { .name = "piofolder", .type = BLOBMSG_TYPE_STRING },
 };
 
 const struct uci_blob_param_list odhcpd_attr_list = {
@@ -276,6 +287,7 @@ static void set_interface_defaults(struct interface *iface)
 	iface->ra_mininterval = iface->ra_maxinterval/3;
 	iface->ra_lifetime = -1;
 	iface->ra_dns = true;
+	iface->pio_update = false;
 }
 
 static void clean_interface(struct interface *iface)
@@ -318,7 +330,7 @@ static void close_interface(struct interface *iface)
 	clean_interface(iface);
 	free(iface->addr4);
 	free(iface->addr6);
-	free(iface->invalid_addr6);
+	free(iface->pios);
 	free(iface->ifname);
 	free(iface);
 }
@@ -387,6 +399,11 @@ static void set_config(struct uci_section *s)
 	if ((c = tb[ODHCPD_ATTR_HOSTSFILE])) {
 		free(config.dhcp_hostsfile);
 		config.dhcp_hostsfile = strdup(blobmsg_get_string(c));
+	}
+
+	if ((c = tb[ODHCPD_ATTR_PIOFOLDER])) {
+		free(config.ra_piofolder);
+		config.ra_piofolder = strdup(blobmsg_get_string(c));
 	}
 
 	if ((c = tb[ODHCPD_ATTR_LEASETRIGGER])) {
@@ -1480,6 +1497,8 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 		}
 	}
 
+	config_load_ra_pio(iface);
+
 	return 0;
 
 err:
@@ -1703,6 +1722,309 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
 	return ipv6_pxe_entry_new(arch, url) ? -1 : 0;
 }
 
+#define JSON_LENGTH "length"
+#define JSON_PREFIX "prefix"
+#define JSON_SLAAC "slaac"
+#define JSON_TIME "time"
+
+static inline time_t config_time_from_json(time_t json_time)
+{
+	time_t ref, now;
+
+	ref = time(NULL);
+	now = odhcpd_time();
+
+	if (now > json_time || ref > json_time)
+		return 0;
+
+	return json_time + (now - ref);
+}
+
+static inline time_t config_time_to_json(time_t config_time)
+{
+	time_t ref, now;
+
+	ref = time(NULL);
+	now = odhcpd_time();
+
+	return config_time + (ref - now);
+}
+
+static inline bool config_ra_pio_enabled(struct interface *iface)
+{
+	return config.ra_piofolder_fd >= 0 && iface->ra == MODE_SERVER && !iface->master;
+}
+
+static bool config_ra_pio_time(json_object *slaac_json, time_t *slaac_time)
+{
+	time_t pio_json_time, pio_time;
+	json_object *time_json;
+
+	time_json = json_object_object_get(slaac_json, JSON_TIME);
+	if (!time_json)
+		return true;
+
+	pio_json_time = (time_t) json_object_get_int64(time_json);
+	if (!pio_json_time)
+		return true;
+
+	pio_time = config_time_from_json(pio_json_time);
+	if (!pio_time)
+		return false;
+
+	*slaac_time = pio_time;
+
+	return true;
+}
+
+static json_object *config_load_ra_pio_json(struct interface *iface)
+{
+	json_object *json;
+	int fd;
+
+	fd = openat(config.ra_piofolder_fd, iface->ifname, O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		return NULL;
+
+	json = json_object_from_fd(fd);
+
+	close(fd);
+
+	if (!json)
+		syslog(LOG_ERR,
+			"rfc9096: %s: json read error %s",
+			iface->ifname,
+			json_util_get_last_err());
+
+	return json;
+}
+
+void config_load_ra_pio(struct interface *iface)
+{
+	json_object *json, *slaac_json;
+	size_t pio_cnt;
+	time_t now;
+
+	if (!config_ra_pio_enabled(iface))
+		return;
+
+	json = config_load_ra_pio_json(iface);
+	if (!json)
+		return;
+
+	slaac_json = json_object_object_get(json, JSON_SLAAC);
+	if (!slaac_json) {
+		json_object_put(json);
+		return;
+	}
+
+	now = odhcpd_time();
+
+	pio_cnt = json_object_array_length(slaac_json);
+	iface->pios = malloc(sizeof(struct ra_pio) * pio_cnt);
+	if (!iface->pios) {
+		json_object_put(json);
+		return;
+	}
+
+	iface->pio_cnt = 0;
+	for (size_t i = 0; i < pio_cnt; i++) {
+		json_object *cur_pio_json, *length_json, *prefix_json;
+		const char *pio_str;
+		time_t pio_lt = 0;
+		struct ra_pio *pio;
+		uint8_t pio_len;
+
+		cur_pio_json = json_object_array_get_idx(slaac_json, i);
+		if (!cur_pio_json)
+			continue;
+
+		if (!config_ra_pio_time(cur_pio_json, &pio_lt))
+			continue;
+
+		length_json = json_object_object_get(cur_pio_json, JSON_LENGTH);
+		if (!length_json)
+			continue;
+
+		prefix_json = json_object_object_get(cur_pio_json, JSON_PREFIX);
+		if (!prefix_json)
+			continue;
+
+		pio_len = (uint8_t) json_object_get_uint64(length_json);
+		pio_str = json_object_get_string(prefix_json);
+		pio = &iface->pios[iface->pio_cnt];
+
+		inet_pton(AF_INET6, pio_str, &pio->prefix);
+		pio->length = pio_len;
+		pio->lifetime = pio_lt;
+		syslog(LOG_INFO,
+			"rfc9096: %s: load %s/%u (%u)",
+			iface->ifname,
+			pio_str,
+			pio_len,
+			ra_pio_lifetime(pio, now));
+
+		iface->pio_cnt++;
+	}
+
+	json_object_put(json);
+
+	if (!iface->pio_cnt) {
+		free(iface->pios);
+		iface->pios = NULL;
+	} else if (iface->pio_cnt != pio_cnt) {
+		iface->pios = realloc(iface->pios, sizeof(struct ra_pio) * iface->pio_cnt);
+	}
+}
+
+static void config_save_ra_pio_json(struct interface *iface, struct json_object *json)
+{
+	size_t tmp_piofile_strlen;
+	char *tmp_piofile;
+	int fd, ret;
+
+	tmp_piofile_strlen = strlen(iface->ifname) + 2;
+	tmp_piofile = alloca(tmp_piofile_strlen);
+	snprintf(tmp_piofile, tmp_piofile_strlen, ".%s", iface->ifname);
+
+	fd = openat(config.ra_piofolder_fd,
+		tmp_piofile,
+		O_CREAT | O_TRUNC | O_WRONLY | O_CLOEXEC,
+		0644);
+	if (fd < 0) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %m creating temporary json file",
+			iface->ifname);
+		return;
+	}
+
+	ret = json_object_to_fd(fd, json, JSON_C_TO_STRING_PLAIN);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: json write error %s",
+			iface->ifname,
+			json_util_get_last_err());
+		close(fd);
+		unlinkat(config.ra_piofolder_fd, tmp_piofile, 0);
+		return;
+	}
+
+	ret = fsync(fd);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %m syncing %s",
+			iface->ifname,
+			tmp_piofile);
+		close(fd);
+		unlinkat(config.ra_piofolder_fd, tmp_piofile, 0);
+		return;
+	}
+
+	ret = close(fd);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %m closing %s",
+			iface->ifname,
+			tmp_piofile);
+		unlinkat(config.ra_piofolder_fd, tmp_piofile, 0);
+		return;
+	}
+
+	ret = renameat(config.ra_piofolder_fd,
+		tmp_piofile,
+		config.ra_piofolder_fd,
+		iface->ifname);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %m renaming piofile: %s -> %s",
+			iface->ifname,
+			tmp_piofile,
+			iface->ifname);
+		close(fd);
+		unlinkat(config.ra_piofolder_fd, tmp_piofile, 0);
+		return;
+	}
+
+	iface->pio_update = false;
+	syslog(LOG_WARNING,
+		"rfc9096: %s: piofile updated",
+		iface->ifname);
+}
+
+void config_save_ra_pio(struct interface *iface)
+{
+	struct json_object *json, *slaac_json;
+	char ipv6_str[INET6_ADDRSTRLEN];
+	time_t now;
+
+	if (!config_ra_pio_enabled(iface))
+		return;
+
+	if (!iface->pio_update)
+		return;
+
+	now = odhcpd_time();
+
+	json = json_object_new_object();
+	if (!json)
+		return;
+
+	slaac_json = json_object_new_array_ext(iface->pio_cnt);
+	if (!slaac_json) {
+		json_object_put(slaac_json);
+		return;
+	}
+
+	json_object_object_add(json, JSON_SLAAC, slaac_json);
+
+	for (size_t i = 0; i < iface->pio_cnt; i++) {
+		struct json_object *cur_pio_json, *len_json, *pfx_json;
+		const struct ra_pio *cur_pio = &iface->pios[i];
+
+		if (ra_pio_expired(cur_pio, now))
+			continue;
+
+		cur_pio_json = json_object_new_object();
+		if (!cur_pio_json)
+			continue;
+
+		inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str));
+
+		pfx_json = json_object_new_string(ipv6_str);
+		if (!pfx_json) {
+			json_object_put(cur_pio_json);
+			continue;
+		}
+
+		len_json = json_object_new_uint64(cur_pio->length);
+		if (!len_json) {
+			json_object_put(cur_pio_json);
+			json_object_put(pfx_json);
+			continue;
+		}
+
+		json_object_object_add(cur_pio_json, JSON_PREFIX, pfx_json);
+		json_object_object_add(cur_pio_json, JSON_LENGTH, len_json);
+
+		if (cur_pio->lifetime) {
+			struct json_object *time_json;
+			time_t pio_lt;
+
+			pio_lt = config_time_to_json(cur_pio->lifetime);
+
+			time_json = json_object_new_int64(pio_lt);
+			if (time_json)
+				json_object_object_add(cur_pio_json, JSON_TIME, time_json);
+		}
+
+		json_object_array_add(slaac_json, cur_pio_json);
+	}
+
+	config_save_ra_pio_json(iface, json);
+
+	json_object_put(json);
+}
+
 void odhcpd_reload(void)
 {
 	struct uci_context *uci = uci_alloc_context();
@@ -1755,6 +2077,17 @@ void odhcpd_reload(void)
 
 		mkdir_p(dirname(path), 0755);
 		free(path);
+	}
+
+	if (config.ra_piofolder) {
+		char *path = strdupa(config.ra_piofolder);
+
+		mkdir_p(path, 0755);
+
+		close(config.ra_piofolder_fd);
+		config.ra_piofolder_fd = open(path, O_PATH | O_DIRECTORY | O_CLOEXEC);
+		if (config.ra_piofolder_fd < 0)
+			syslog(LOG_ERR, "Unable to open piofolder '%s': %m", path);
 	}
 
 	vlist_flush(&leases);
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -140,7 +140,6 @@ struct odhcpd_ipaddr {
 		/* ipv6 only */
 		struct {
 			uint8_t dprefix;
-			uint8_t invalid_advertisements;
 			bool tentative;
 		};
 
@@ -173,6 +172,10 @@ struct config {
 	char *dhcp_cb;
 	char *dhcp_statefile;
 	char *dhcp_hostsfile;
+
+	char *ra_piofolder;
+	int ra_piofolder_fd;
+
 	int log_level;
 };
 
@@ -270,6 +273,14 @@ struct dnr_options {
 };
 
 
+// RA PIO - RFC9096
+struct ra_pio {
+	struct in6_addr prefix;
+	uint8_t length;
+	time_t lifetime;
+};
+
+
 struct interface {
 	struct avl_node avl;
 
@@ -281,8 +292,6 @@ struct interface {
 	// IPv6 runtime data
 	struct odhcpd_ipaddr *addr6;
 	size_t addr6_len;
-	struct odhcpd_ipaddr *invalid_addr6;
-	size_t invalid_addr6_len;
 
 	// RA runtime data
 	struct odhcpd_event router_event;
@@ -403,6 +412,11 @@ struct interface {
 	// DNR
 	struct dnr_options *dnr;
 	size_t dnr_cnt;
+
+	// RA PIO - RFC9096
+	struct ra_pio *pios;
+	size_t pio_cnt;
+	bool pio_update;
 };
 
 extern struct avl_tree interfaces;
@@ -432,6 +446,24 @@ inline static struct dhcp_assignment *alloc_assignment(size_t extra_len)
 	INIT_LIST_HEAD(&a->lease_list);
 
 	return a;
+}
+
+inline static bool ra_pio_expired(const struct ra_pio *pio, time_t now)
+{
+	return pio->lifetime && (now > pio->lifetime);
+}
+
+inline static uint32_t ra_pio_lifetime(const struct ra_pio *pio, time_t now)
+{
+	if (!pio->lifetime || now > pio->lifetime)
+		return 0;
+
+	return (uint32_t) (pio->lifetime - now);
+}
+
+inline static bool ra_pio_stale(const struct ra_pio *pio)
+{
+	return !!pio->lifetime;
 }
 
 // Exported main functions
@@ -469,6 +501,8 @@ struct lease *config_find_lease_by_duid(const uint8_t *duid, const uint16_t len)
 struct lease *config_find_lease_by_mac(const uint8_t *mac);
 struct lease *config_find_lease_by_hostid(const uint64_t hostid);
 struct lease *config_find_lease_by_ipaddr(const uint32_t ipaddr);
+void config_load_ra_pio(struct interface *iface);
+void config_save_ra_pio(struct interface *iface);
 int set_lease_from_blobmsg(struct blob_attr *ba);
 
 #ifdef WITH_UBUS
--- a/src/router.c
+++ b/src/router.c
@@ -446,6 +446,116 @@ struct nd_opt_dnr_info {
 	uint8_t body[];
 };
 
+/* IPv6 RA PIOs */
+static struct ra_pio *router_find_ra_pio(struct interface *iface,
+	struct nd_opt_prefix_info *p)
+{
+	for (size_t i = 0; i < iface->pio_cnt; i++) {
+		struct ra_pio *cur_pio = &iface->pios[i];
+
+		if (p->nd_opt_pi_prefix_len == cur_pio->length &&
+			!odhcpd_bmemcmp(&p->nd_opt_pi_prefix, &cur_pio->prefix, cur_pio->length))
+			return cur_pio;
+	}
+
+	return NULL;
+}
+
+static void router_add_ra_pio(struct interface *iface,
+	struct nd_opt_prefix_info *p)
+{
+	char ipv6_str[INET6_ADDRSTRLEN];
+	struct ra_pio *new_pios, *pio;
+
+	pio = router_find_ra_pio(iface, p);
+	if (pio) {
+		if (pio->lifetime) {
+			pio->lifetime = 0;
+
+			iface->pio_update = true;
+			syslog(LOG_WARNING, "rfc9096: %s: renew %s/%u",
+				iface->ifname,
+				inet_ntop(AF_INET6, &pio->prefix, ipv6_str, sizeof(ipv6_str)),
+				pio->length);
+		}
+
+		return;
+	}
+
+	new_pios = realloc(iface->pios, sizeof(struct ra_pio) * (iface->pio_cnt + 1));
+	if (!new_pios)
+        return;
+
+	iface->pios = new_pios;
+	pio = &iface->pios[iface->pio_cnt];
+	iface->pio_cnt++;
+
+	memcpy(&pio->prefix, &p->nd_opt_pi_prefix, sizeof(pio->prefix));
+	pio->length = p->nd_opt_pi_prefix_len;
+	pio->lifetime = 0;
+
+	iface->pio_update = true;
+	syslog(LOG_INFO, "rfc9096: %s: add %s/%u",
+		iface->ifname,
+		inet_ntop(AF_INET6, &pio->prefix, ipv6_str, sizeof(ipv6_str)),
+		pio->length);
+}
+
+static void router_clear_ra_pio(time_t now,
+	struct interface *iface)
+{
+	size_t i = 0, pio_cnt = iface->pio_cnt;
+	char ipv6_str[INET6_ADDRSTRLEN];
+
+	while (i < iface->pio_cnt) {
+		struct ra_pio *cur_pio = &iface->pios[i];
+
+		if (ra_pio_expired(cur_pio, now)) {
+			syslog(LOG_INFO,
+				"rfc9096: %s: clear %s/%u",
+				iface->ifname,
+				inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str)),
+				cur_pio->length);
+
+			if (i + 1 < iface->pio_cnt)
+				iface->pios[i] = iface->pios[iface->pio_cnt - 1];
+
+			iface->pio_cnt--;
+		} else {
+			i++;
+		}
+	}
+
+	if (!iface->pio_cnt) {
+		free(iface->pios);
+		iface->pios = NULL;
+	} else if (iface->pio_cnt != pio_cnt) {
+		struct ra_pio *new_pios = realloc(iface->pios, sizeof(struct ra_pio) * iface->pio_cnt);
+
+		if (new_pios)
+			iface->pios = new_pios;
+	}
+}
+
+static void router_stale_ra_pio(struct interface *iface,
+	struct nd_opt_prefix_info *p,
+	time_t now)
+{
+	struct ra_pio *pio = router_find_ra_pio(iface, p);
+	char ipv6_str[INET6_ADDRSTRLEN];
+
+	if (!pio || pio->lifetime)
+		return;
+
+	pio->lifetime = now + iface->max_valid_lifetime;
+
+	iface->pio_update = true;
+	syslog(LOG_WARNING, "rfc9096: %s: stale %s/%u",
+		iface->ifname,
+		inet_ntop(AF_INET6, &pio->prefix, ipv6_str, sizeof(ipv6_str)),
+		pio->length);
+}
+
 /* Router Advert server mode */
 static int send_router_advert(struct interface *iface, const struct in6_addr *from)
 {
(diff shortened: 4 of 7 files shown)


Noltari:
This is an lternative implementation to https://github.com/openwrt/odhcpd/pull/245 with less flash writes, ~~which depends on merging https://github.com/openwrt/odhcpd/pull/209 first~~ done in 86c0200f80670c36d200f827f3558e4f676d9171.
I'm currently testing this on my home network.

Noltari:
Now that https://github.com/openwrt/odhcpd/pull/209 has been merged, this PR is ready for review.
I'm currently testing it on my home network and so far so good.
With this alternative implementation, flash is only written whenever a new prefix is obtained or when an older prefix becomes stale. If none of that happens and the same prefix is renewed over and over, the flash won't be written at all until the router is restarted or the WAN connection is restablished.
The previous implementation relied on writting the flash every 12h when the prefix was renewed, which could potentially damage it on the long term, but this is no longer the case.
Anyhow, it should probably be enabled with a folder in `/tmp` and let the users decide if they want to trigger flash writes, which should be the case for ISPs with dynamic IPv6 prefixes.

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	time_t now;
+	size_t i;
+
+	if (!config.ra_piofolder)
+		return;
+
+	if (!iface->pio_update)
+		return;
+
+	if (!(iface->ra == MODE_SERVER && !iface->master))
```
systemcrash: Perhaps this can go first as a prerequisite?

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	struct json_object *json, *slaac_json;
+	char ipv6_str[INET6_ADDRSTRLEN];
+	time_t now;
+	size_t i;
+
+	if (!config.ra_piofolder)
+		return;
+
+	if (!iface->pio_update)
+		return;
```
systemcrash: can these two be coalesced with an `||`?

src/router.c
```diff
@@ -446,6 +446,111 @@ struct nd_opt_dnr_info {
...
+	struct interface *iface)
+{
+	char buf[INET6_ADDRSTRLEN];
+	size_t i = 0;
+
+	while (i < iface->pio_cnt) {
+		struct ra_pio *cur_pio = &iface->pio[i];
+
+		if (ra_pio_expired(cur_pio, now)) {
+			syslog(LOG_WARNING, "rfc9096: %s: rem %s/%u", iface->ifname, inet_ntop(AF_INET6, &cur_pio->prefix, buf, sizeof(buf)), cur_pio->length);
```
systemcrash: perhaps log as 'clear' or the function renamed to `rem`?

src/router.c
```diff
@@ -446,6 +446,111 @@ struct nd_opt_dnr_info {
...
+
+			if (i + 1 < iface->pio_cnt)
+				memmove(&iface->pio[i], &iface->pio[i + 1], sizeof(struct ra_pio) * (iface->pio_cnt - i - 1));
```
systemcrash: `memmove` is an expensive albeit infrequent operation. Perhaps just swap in the last array entry instead (and decrement the count)?

systemcrash: I'm not certain what's better - and this is nitpicking about cpu cycles - but if the last swap trick works, that'd be a nice optimisation.

Noltari: I just copied this implementation from: https://github.com/openwrt/odhcpd/blob/b1547922257e484cb18e1719bdfc18f4789eb560/src/router.c#L537-L562
But yes, we can probably optimise it and just do:
`memcpy(&iface->pio[i], &iface->pio[iface->pio_cnt - 1], sizeof(struct ra_pio));`
Is that OK?

systemcrash: I was thinking more the following, because I haven't seen that PIO order is important.
```c
iface->pio[i] = iface->pio[iface->pio_cnt - 1];
iface->pio_cnt--;
```
I haven't seen that there's a good reason why we don't do that. I mean, just a few lines lower down we realloc anyway?

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	ret = json_object_to_fd(fd, json, JSON_C_TO_STRING_PLAIN);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: json error %s",
+			iface->name,
+			json_util_get_last_err());
+		unlink(tmp_piofile);
+		return;
+	}
+
```
systemcrash, line 1932: Perhaps a `fsync(fd)` would be suitable here between `json_object_to_fd()` and `rename()`?

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	piofile_strlen = ifname_strlen + piofolder_strlen + 2;
+	tmp_piofile_strlen = piofile_strlen + 1;
+
+	ifname = strndup(iface->ifname, ifname_strlen);
+	piofolder = strndup(config.ra_piofolder, piofolder_strlen);
+	piofile = alloca(piofile_strlen);
+	tmp_piofile = alloca(tmp_piofile_strlen);
+
+	snprintf(piofile, piofile_strlen, "%s/%s", piofolder, ifname);
+	snprintf(tmp_piofile, tmp_piofile_strlen, "%s/.%s", piofolder, ifname);
```
systemcrash: Minor, but we could do the following and elide the strndup stuff, avoiding stack allocation:
```c
snprintf(piofile, sizeof(piofile), "%s/%s", config.ra_piofolder, iface->ifname);
snprintf(tmp_piofile, sizeof(tmp_piofile), "%s/.%s", config.ra_piofolder, iface->ifname);
```

systemcrash: Then we skip the `free()` next.

Aukansh:
2025-09-24 14:00:58 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:01:13 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:01:13 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:01:20 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:01:20 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:01:23 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:01:23 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:01:26 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:01:26 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:07:35 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:07:35 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:08:30 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:08:30 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:13:45 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:13:45 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:21:09 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:21:09 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:28:00 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
2025-09-24 14:28:00 daemon.warn odhcpd[7458]: rfc9096: br-lan: stale fd48:cba1:cfb2::/64
2025-09-24 14:34:51 daemon.warn odhcpd[7458]: rfc9096: br-lan: renew fd48:cba1:cfb2::/64
.........

logs are too frequent and noisy!

Noltari:
> logs are too frequent and noisy!

That shouldn't happen and it definitely doesn't happen to me with my current config.
Is there anything specific on your config that might be causing this?
Can you share your config?

Aukansh:
> > logs are too frequent and noisy!
> 
> That shouldn't happen and it definitely doesn't happen to me with my current config. Is there anything specific on your config that might be causing this? Can you share your config?

Sorry, it might be due to my configuration error！

Noltari:
> Sorry, it might be due to my configuration error！

@xiaobaishu1 if you could provide more info about that configuration error it would be great.
I can try to replicate and debug it in order to minimise the logging in that case.

Aukansh:
> > Sorry, it might be due to my configuration error！
> 
> @xiaobaishu1 if you could provide more info about that configuration error it would be great. I can try to replicate and debug it in order to minimise the logging in that case.

I've upgraded, the configuration has been deleted, and I'll report back if there's a recurrence!

README
```diff
@@ -67,6 +67,7 @@ leasefile	string				DHCP/v6 lease/hostfile
 leasetrigger	string				Lease trigger script
 hostsfile	string				DHCP/v6 hostfile
 loglevel	integer 6			Syslog level priority (0-7)
+piofolder	string				IPv6 RA PIO folder
```
Alphix: Maybe the description of the `piofolder` cfg parameter could be a bit more informative...?

src/config.c
Alphix: I would suggest using `open(config.ra_piofolder, O_PATH | O_DIRECTORY | O_CLOEXEC)` in a separate `config_init_ra_pio()` function, then stash that `fd` in `config` and use `openat()`/`renameat()`/etc throughout, which would remove all the `strlen()`, `strndup()` and `alloca()` calls.

(I've checked and musl supports all of the above)

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
 	return ipv6_pxe_entry_new(arch, url) ? -1 : 0;
 }
 
+#define JSON_LENGTH "length"
+#define JSON_PREFIX "prefix"
+#define JSON_SLAAC "slaac"
+#define JSON_TIME "time"
+
+static inline uint32_t config_time_from_json(time_t json_time)
```
Alphix: Why not return a `time_t` here?

src/router.c
```diff
@@ -651,8 +760,29 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
...
+
+			router_stale_ra_pio(iface, p, now);
+		}
+		else
+		{
+			p->nd_opt_pi_preferred_time = htonl(preferred_lt);
+			p->nd_opt_pi_valid_time = htonl(valid_lt);
+
+			router_add_ra_pio(iface, p);
+		}
```
Alphix: This brace style doesn't match the style used elsewhere (`if {` and `} else {`)

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	snprintf(piofile, piofile_strlen, "%s/%s", piofolder, ifname);
+	snprintf(tmp_piofile, tmp_piofile_strlen, "%s/.%s", piofolder, ifname);
+
+	free(ifname);
+	free(piofolder);
+
+	fd = open(tmp_piofile, O_CREAT | O_TRUNC | O_WRONLY | O_CLOEXEC, 0644);
+	if (fd < 0) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %d creating %s",
```
Alphix: Printing the `fd` isn't very useful, it'll just be `-1`. Use `%m` instead (here and in the other error messages where `errno` indicates some error)

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+		size_t i;
+
+		if (!slaac_json) {
+			json_object_put(json);
+			return;
+		}
+
+		now = odhcpd_time();
+
+		iface->pio_cnt = 0;
```
Alphix:
```C
iface->pio_cnt = json_object_array_length(slaac_json);
iface->pio = malloc(sizeof(struct ra_pio) * iface->pio_count);
if (!iface->pio)
        return;
```

That'll simplify the logic later and avoid all the `realloc()` calls...

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+
+		for (i = 0; i < json_object_array_length(slaac_json); i++) {
```
Alphix: I would suggest changing this to:
```C
for (size_t i = 0; i < json_object_array_length(slaac_json); i++) {
```

Here and in all other `for` loops

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+			json_object *cur_pio_json;
+			uint32_t pio_lt = 0;
+
+			cur_pio_json = json_object_array_get_idx(slaac_json, i);
+			if (!cur_pio_json)
+				continue;
+
+			if (config_ra_pio_time(cur_pio_json, &pio_lt)) {
```
Alphix: Invert the check and bail early...
```C
if (!config_ra_pio_time(cur_pio_json, &pio_lt))
        continue;
```
That'll reduce the indentation level, making the rest easier to read

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+						continue;
+
+					pio_len = (uint8_t) json_object_get_uint64(length_json);
+					pio_str = json_object_get_string(prefix_json);
+					pio = &iface->pio[iface->pio_cnt - 1];
+
+					inet_pton(AF_INET6, pio_str, &pio->prefix);
+					pio->length = pio_len;
+					pio->lifetime = pio_lt;
+					syslog(LOG_WARNING,
```
Alphix: `LOG_WARNING` for something that was successful seems a bit exaggerated? (that applies to a couple of other places as well)

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+
+	if (!iface->pio_update)
+		return;
+
+	if (!(iface->ra == MODE_SERVER && !iface->master))
+		return;
+
+	now = odhcpd_time();
+
+	json = json_object_new_object();
```
Alphix, line 1968: This (and a lot of the other `json_*()` calls) can return `NULL`

src/odhcpd.h
```diff
@@ -270,6 +270,14 @@ struct dnr_options {
 };
 
 
+// RA PIO - RFC9096
+struct ra_pio {
+	struct in6_addr prefix;
+	uint8_t length;
+	uint32_t lifetime;
```
Alphix: `time_t`?

src/ubus.c
```diff
@@ -176,6 +176,65 @@ static int handle_dhcpv6_leases(_unused struct ubus_context *ctx, _unused struct
 	return 0;
 }
 
+static int handle_ra_pio(_unused struct ubus_context *ctx, _unused struct ubus_object *obj,
+		_unused struct ubus_request_data *req, _unused const char *method,
+		_unused struct blob_attr *msg)
+{
+	char ipv6_str[INET6_ADDRSTRLEN];
```
Alphix, line 183: Nitpick: this is too short (you actually store the result from `inet_ntop()` + `/%u` in this string). I suggest adding two separate buffers in the `for` loop below (see separate comment).

src/ubus.c
```diff
@@ -176,6 +176,65 @@ static int handle_dhcpv6_leases(_unused struct ubus_context *ctx, _unused struct
...
+				continue;
+
+			cur_pio_blob = blobmsg_open_table(&b, NULL);
+
+			pio_lt = ra_pio_lifetime(cur_pio, now);
+			pio_stale = ra_pio_stale(cur_pio);
+
+			inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str));
+			ipv6_strlen = strnlen(ipv6_str, INET6_ADDRSTRLEN);
+			snprintf(&ipv6_str[ipv6_strlen], INET6_ADDRSTRLEN - ipv6_strlen, "/%u", cur_pio->length);
```
Alphix: Suggestion:

```C
char ipv6_addr[INET6_ADDRSTRLEN];
char ipv6_prefix[INET6_ADDRSTRLEN + sizeof("/128")];
...
inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_addr, sizeof(ipv6_addr));
snprintf(ipv6_prefix, sizeof(ipv6_prefix), "%s/%" PRIu8, ipv6_addr, cur_pio->length);

File: src/ubus.c [hunk 19]
```diff
@@ -176,6 +176,65 @@ static int handle_dhcpv6_leases(_unused struct ubus_context *ctx, _unused struct
...
+
+			if (pio_lt)
+				blobmsg_add_u32(&b, "lifetime", pio_lt);
```
Alphix, line 217: Nitpick: suggest `blobmsg_add_u64()` here, `time_t` is 64 bits even on 32 bit platforms these days...

src/router.c
Alphix: Also: the `realloc()` later can actually be skipped? Having a `iface->pio` allocation which is larger than necessary isn't really a problem...

```C
static void router_clear_ra_pio(time_t now, struct interface *iface)
{
	for (size_t i = 0; i < iface->pio_cnt;) {
		struct ra_pio *cur_pio = &iface->pio[i];
		char addr[INET6_ADDRSTRLEN];

		if (!ra_pio_expired(cur_pio, now)) {
			i++;
			continue;
		}

		inet_ntop(AF_INET6, &cur_pio->prefix, addr, sizeof(addr));
		syslog(LOG_INFO, "rfc9096: %s: %s/%" PRIu8 " expired",
		       iface->ifname, addr, cur_pio->length);

		iface->pio[i] = iface->pio[--iface->pio_cnt];
	}

	if (!iface->pio_cnt) {
		free(iface->pio);
		iface->pio = NULL;
	}
}
```

src/odhcpd.h
```diff
@@ -434,6 +445,24 @@ inline static struct dhcp_assignment *alloc_assignment(size_t extra_len)
 	return a;
 }
 
+inline static bool ra_pio_expired(const struct ra_pio *pio, time_t now)
+{
+	return pio->lifetime && (now > pio->lifetime);
+}
+
+inline static uint32_t ra_pio_lifetime(const struct ra_pio *pio, time_t now)
```
Alphix, line 456: `time_t`?

src/router.c
```diff
@@ -446,6 +446,111 @@ struct nd_opt_dnr_info {
...
+				inet_ntop(AF_INET6, &pio->prefix, buf, sizeof(buf)),
+				pio->length);
+		}
+
+		return;
+	}
+
+	iface->pio_cnt++;
+	iface->pio = realloc(iface->pio, sizeof(struct ra_pio) * iface->pio_cnt);
+	if (iface->pio) {
```
Alphix: If `realloc()` fails, `iface->pio_cnt` will have the wrong value and `iface->pio` will be leaked.

Also, I'd suggest inverting the `if` check and bail early:

```C
struct ra_pio *new_pios;
...
new_pios = realloc(iface->pio, sizeof(struct ra_pio) * (iface->pio_cnt + 1));
if (!new_pios)
        return;
iface->pio = new_pios;
iface->pio_cnt++;
...
```

src/odhcpd.h
```diff
@@ -403,6 +409,11 @@ struct interface {
 	// DNR
 	struct dnr_options *dnr;
 	size_t dnr_cnt;
+
+	// RA PIO - RFC9096
+	struct ra_pio *pio;
```
Alphix: Nitpick: I'd suggest naming this (and all other places where an array of `struct ra_pio` are used) as `pios` to make it clear that this is a plurality of pios.

src/config.c
```diff
@@ -1703,6 +1714,272 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+
+static void config_save_ra_pio_json(struct interface *iface, struct json_object *json)
+{
+	unsigned ifname_strlen, piofile_strlen, piofolder_strlen, tmp_piofile_strlen;
+	char *ifname, *piofile, *piofolder, *tmp_piofile;
+	struct stat sb;
+	int fd, ret;
+
+	if (stat(config.ra_piofolder, &sb) || !S_ISDIR(sb.st_mode)) {
+		ret = mkdir(config.ra_piofolder, 0755);
```
Alphix: Two things:

1) This won't work if the user sets `ra_piofolder` to `/tmp/foo/bar/pio` (use `mkdir_p()` from `libubox`)
2) The `stat` call isn't really adding anything. If you implement the `openat()` suggestion, I'd just suggest doing the `mkdir_p()` call, checking its status, then trying to open the dir and blorting out an error message if that step fails

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44

src/router.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44

src/router.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44

src/config.c
Noltari: Introduced config_ra_pio_enabled() in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44, so this change no longer applies.

src/config.c
Noltari: Introduced config_ra_pio_enabled() in https://github.com/Noltari/odhcpd/compare/0dbef5c..f080b44

systemcrash (approved):

Alphix (changes requested):

Alphix:
Sorry, I missed actually pressing the finish review button, so my comments weren't actually published...

src/config.c
systemcrash: Had a similar thought, though that might require reworking all internal time handling variables (for consistency) to time_t. time_t permits 64 bit handling, but 32 bits is fine for up to 127 or so years of leasetime :)

Perhaps a PR to rework those variables at a later stage?

README
systemcrash: suggestion: `IPv6 RA PIO folder where the current leases JSON file is permanently stored`?

src/config.c
systemcrash: instead of earlier explicit declaration as size_t?

src/config.c
systemcrash: openat suggestion?

mkdir_p handles the case where the folder might not exist. Both variants seem equally readable.

src/config.c
```diff
@@ -1703,6 +1715,298 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+			iface->name,
+			errno,
+			tmp_piofile,
+			piofile);
+		unlink(tmp_piofile);
+		return;
+	}
+
+	iface->pio_update = false;
+	syslog(LOG_WARNING,
```
systemcrash, line 1949: Similarly: LOG_INFO, since this is now success/completion?

src/router.c
```diff
@@ -446,6 +446,115 @@ struct nd_opt_dnr_info {
...
+		pio = &iface->pio[iface->pio_cnt - 1];
+
+		memcpy(&pio->prefix, &p->nd_opt_pi_prefix, sizeof(pio->prefix));
+		pio->length = p->nd_opt_pi_prefix_len;
+		pio->lifetime = 0;
+
+		iface->pio_update = true;
+		syslog(LOG_WARNING, "rfc9096: %s: add %s/%u",
```
systemcrash: LOG_INFO?

README
Alphix: The `JSON` part is just an implementation detail, and we're technically not storing "PIO" (Prefix Information Option...what gets sent in RAs), but prefix information.

How about `Folder to store IPv6 prefix information (to detect stale prefixes, see RFC9096, §3.5)`

src/ubus.c
systemcrash: Are you sure a prefix takes "/128"?

src/router.c
```diff
@@ -446,6 +446,115 @@ struct nd_opt_dnr_info {
...
+	struct interface *iface)
+{
+	char buf[INET6_ADDRSTRLEN];
+	size_t i = 0;
+
+	while (i < iface->pio_cnt) {
+		struct ra_pio *cur_pio = &iface->pio[i];
+
+		if (ra_pio_expired(cur_pio, now)) {
+			syslog(LOG_WARNING,
```
systemcrash: LOG_INFO?

src/config.c
Alphix: To make it clearer that the variable isn't used outside the scope of the loop (and for consistency with the coding style of the rest of `odhcpd`)

src/ubus.c
Alphix: I'm sure it doesn't. But I prefer being pessimistic. The address will also never be INET6_ADDRSTRLEN long, since the last 64 bits will all be zero...

(Edit: actually the prefix could be really long)

src/config.c
Alphix: openat() was something I suggested in another comment that's already closed (using a dirfd and openat()/renameat() to simplify dealing with the creation of files).

But yes, the main point was to use mkdir_p() here so that recursive folder creation works as expected.

src/config.c
Noltari: I'm just following the rest of odhcpd behaviour here, so I don't think this should be addressed in this PR...

src/ubus.c
Noltari: This is not needed since the prefix will always be /64 for SLAAC.

src/ubus.c
Noltari: This is not needed since the prefix will always be /64 for SLAAC.

src/config.c
Noltari: IMO this one should be kept as warning since it means that the flash has been written and might prevent bricking the router if for some reason we start writing the flash over and over.

src/router.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/router.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/odhcpd.h
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/router.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

However, I still had to keep a realloc after the loop, in order to reduce the allocated array in case some of the prefix had expired.

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

(Just this one, I need to find the time to review the rest of the json calls that return NULL)

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/router.c
Noltari: Fixed in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

README
Noltari: Done in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0

src/config.c
Noltari: Switched to mkdir_p in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0
However, I still think the stat() is OK, but I will try to find time to investigate your suggestion...

Noltari:
@Alphix many thanks for your thorough review and suggestions. I've implemented many of them in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0 and marked them as resolved.

src/router.c
Alphix: Strictly speaking, the array doesn't have to be realloc():ed just because less of it is being used... But not going to insist on tiny things like that :)

src/config.c
Alphix: There's nothing *wrong* with stat(), it's just a bit pointless since the open(...O_DIRECTORY) Will anyway tell you if it's not a directory..

Alphix:
> @Alphix many thanks for your thorough review and suggestions. I've implemented many of them in https://github.com/Noltari/odhcpd/compare/f080b44..0f704c0 and marked them as resolved.

Thanks, I'll have a look ASAP :)

src/config.c
Alphix: Ah, I might have been a bit quick with the time_t suggestions... I'll double check how time is handled later, but it's not a reason to hold up this PR

src/config.c
Alphix: I know, it's a real hassle...That's why I was lazy and only marked this one spot...

src/ubus.c
Alphix: Can't really check the code right now (on my phone), so I might be mistaken, but I'm guessing *all* announced prefixes are stored, and it is possible to announce prefixes that are not suitable for SLAAC (i.e. longer than 64, RFC4861, 4.6.2, prefix length).

src/config.c
Alphix: I think that makes sense

src/config.c
Noltari: > Ah, I might have been a bit quick with the time_t suggestions... I'll double check how time is handled later, but it's not a reason to hold up this PR

`preferred_lt` and `valid_lt` are both implemented as `uint32_t` https://github.com/openwrt/odhcpd/blob/bc9f9d93d4d6b8feb7b19235d7f0371480fc679d/src/odhcpd.h#L136-L137
'TIME_LEFT' checks if lifetime is UINT32_MAX or not and substracts `now`. https://github.com/openwrt/odhcpd/blob/bc9f9d93d4d6b8feb7b19235d7f0371480fc679d/src/router.c#L46
So the current implementation is based on `uint32_t` lifetimes.

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0f704c0..177a039

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0f704c0..177a039

src/ubus.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0f704c0..177a039

src/ubus.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/0f704c0..177a039

Noltari:
@Alphix the rest of your suggestions have been implemented in https://github.com/Noltari/odhcpd/compare/0f704c0..177a039 (except for the `time_t` lifetimes)
I've also fixed a bug causing odhcpd to crash in the previous push (realloc missing pio_cnt increment for the add function)

src/config.c
```diff
@@ -1703,6 +1715,315 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+{
+	json_object *json, *slaac_json;
+	size_t pio_cnt;
+	time_t now;
+
+	if (!config_ra_pio_enabled(iface))
+		return;
+
+	json = config_load_ra_pio_json(iface);
+	if (json)
```
systemcrash: Shouldn't this be `if (!json)`?

README
```diff
@@ -67,6 +67,8 @@ leasefile	string				DHCP/v6 lease/hostfile
 leasetrigger	string				Lease trigger script
 hostsfile	string				DHCP/v6 hostfile
 loglevel	integer 6			Syslog level priority (0-7)
+piofolder	string				Folder to store IPv6 prefix information (to
```
systemcrash, line 70: On second thought, we should probably note that `Folder on permanent storage to store...`

Alphix: But the user is free to choose perm/nonperm (this still has value on ephemeral storage, e.g. for handling PPPoE reconnections...)?

systemcrash: Quite true, although the purpose is that the PIOs survive a reboot in general.

Noltari: > On second thought, we should probably note that `Folder on permanent storage to store...`

It will be enabled by default on ephemeral storage as @Alphix pointed out, and it will be up to the users to decide if they want to switch to permanent storage.
Some ISPs force PPPoE reconnects every 24h, and that will definitely wear out the flash.

Even if the users stick to ephemeral storage it will still have value for them, since they will keep the stale PIOs when changing the dhcp config and restarting odhcpd.

src/config.c
Noltari: Nice catch!

Fixed in https://github.com/Noltari/odhcpd/compare/177a039..84d230a

src/odhcpd.h
Noltari: Done in https://github.com/Noltari/odhcpd/compare/84d230a..44054b1

src/ubus.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/84d230a..44054b1

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/84d230a..44054b1

src/odhcpd.h
Noltari: The returned value is relative to the current time, so it doesn't make sense to provide a time_t here since it will be `iface->max_valid_lifetime` in the worst case (about 90 minutes or 5400 seconds).

src/ubus.c
Noltari: The returned value from `ra_pio_lifetime()` is relative to the current time, so it doesn't make sense to provide a `time_t` here since it will be `iface->max_valid_lifetime` in the worst case (about 90 minutes or 5400 seconds).

src/router.c
Noltari: > Strictly speaking, the array doesn't have to be realloc():ed just because less of it is being used... But not going to insist on tiny things like that :)

You're right, but I think it's cleaner to use as less memory as possible...

src/config.c
Noltari: I think I've fixed the rest of json_*() calls in https://github.com/Noltari/odhcpd/compare/44054b1..e207381

Noltari:
@Alphix @systemcrash I think I have now addressed all your comments.
Please check again when you have time and if you have any more suggestions or changes I will look into them.
Thanks again for your reviews!

src/odhcpd.h
Alphix: Yes, it's fine, I think you can ignore my comments wrt `time_t`. It's anyway an internal detail, so it can always be changed later if it turns out that it's necessary. :)

src/ubus.c
```diff
@@ -176,6 +176,64 @@ static int handle_dhcpv6_leases(_unused struct ubus_context *ctx, _unused struct
...
+				continue;
+
+			cur_pio_blob = blobmsg_open_table(&b, NULL);
+
+			pio_lt = ra_pio_lifetime(cur_pio, now);
+			pio_stale = ra_pio_stale(cur_pio);
+
+			inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str));
+			ipv6_strlen = strnlen(ipv6_str, sizeof(ipv6_str));
+			snprintf(&ipv6_str[ipv6_strlen], sizeof(ipv6_str) - ipv6_strlen, "/%u", cur_pio->length);
```
Alphix: On second thought, why not just make `cur_pio->length` a separate variable in this blobmsg, then you can simplify this by removing the last two lines and change `char ipv6_str[INET6_ADDRSTRLEN + 4];` back to `char ipv6_str[INET6_ADDRSTRLEN];`

src/config.c
```diff
@@ -1703,6 +1715,340 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+	ifname_strlen = strlen(iface->ifname);
+	piofolder_strlen = strlen(config.ra_piofolder);
+	piofile_strlen = ifname_strlen + piofolder_strlen + 2;
+
+	piofile = alloca(piofile_strlen);
+	snprintf(piofile, piofile_strlen, "%s/%s", config.ra_piofolder, iface->ifname);
+
+	fd = open(piofile, O_RDONLY);
+	if (fd < 0)
+		return NULL;
```
Alphix: That wasn't quite what I meant with the `dirfd`/`openat()` suggestion.

It's quite a lot of minor changes, so instead of adding it as a review comment, I've made a PR in your repo :)

See https://github.com/Noltari/odhcpd/pull/1/commits/68b4fa79dd7cb67b26ff0d4a4107fbe0ff152d8e

Alphix (changes requested):

src/ubus.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/e207381..78afc91

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/e207381..78afc91

src/ubus.c
```diff
@@ -176,6 +176,62 @@ static int handle_dhcpv6_leases(_unused struct ubus_context *ctx, _unused struct
...
+
+			pio_lt = ra_pio_lifetime(cur_pio, now);
+			pio_stale = ra_pio_stale(cur_pio);
+
+			inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str));
+
+			if (pio_lt)
+				blobmsg_add_u32(&b, "lifetime", pio_lt);
+			blobmsg_add_string(&b, "prefix", ipv6_str);
+			blobmsg_add_u16(&b, "length", cur_pio->length);
```
Alphix, line 219: Shouldn't that be `blobmsg_add_u8`?

src/config.c
```diff
@@ -1703,6 +1722,309 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+		free(iface->pios);
+		iface->pios = NULL;
+	} else if (iface->pio_cnt != pio_cnt) {
+		iface->pios = realloc(iface->pios, sizeof(struct ra_pio) * iface->pio_cnt);
+	}
+}
+
+static void config_save_ra_pio_json(struct interface *iface, struct json_object *json)
+{
+	unsigned tmp_piofile_strlen;
```
Alphix: Nitpick...but `size_t`?

src/config.c
```diff
@@ -1703,6 +1722,309 @@ static int ipv6_pxe_from_uci(struct uci_section* s)
...
+
+	ret = close(fd);
+	if (ret) {
+		syslog(LOG_ERR,
+			"rfc9096: %s: error %m closing %s",
+			iface->ifname,
+			tmp_piofile);
+		unlinkat(config.ra_piofolder_fd, tmp_piofile, 0);
+		return;
+	}
```
Alphix, line 1931: Nitpick: I'd just blindly `close()` since it is guaranteed to close the `fd` on Linux and the various error codes that might be returned here are not useful (interrupted by a signal...some I/O error that would anyway be caught by the `fsync()` above...etc)...but I'll leave that up to you :)

Alphix (changes requested):

Alphix:
@Noltari just to be clear...that's really minor things. Thanks for your patience. I'm really looking forward to this feature landing (I've experience flash renumbering events in the past..)

systemcrash (approved):

src/ubus.c
Noltari: Yes, but I tried that and it gets converted to a boolean 🥲

systemcrash: I think I ran into something similar when I was doing something in netifd code.

Alphix: Oh, I didn't know that, sounds weird, but don't let that hold up the PR

src/config.c
Noltari: If you don't mind I'll keep it as is for consistency with the rest of the code (just a bit of OCD here xD).

src/config.c
Noltari: Done in https://github.com/Noltari/odhcpd/compare/78afc91..d7825c6

Noltari:
> @Noltari just to be clear...that's really minor things. Thanks for your patience. I'm really looking forward to this feature landing (I've experience flash renumbering events in the past..)

I think the quality of this PR has improved thanks to your review, so no need to worry about my patience :).
You can find the latest changes at https://github.com/Noltari/odhcpd/compare/78afc91..d7825c6

src/config.c
Alphix: I know OCD

Alphix (approved):

Alphix:
LGTM

systemcrash:
Two questions: 

1) If RFC8475 says we cannot set valid lifetime to 0 unless the RA is authenticated, how does this RFC rationalise it? I guess there is no broad support for SeND (I've never seen it), so this RFC just does ¯\\_ (ツ)_/¯
2) How can we be sure that when we have multiple upstreams, the correct PIO is made stale?

e.g. ISP_A updates, so only `2001:db8:1:1::/64` should be marked stale or cleared. 
```

                                                              --------
                                             ,-------,       /        \
                   +----+ 2001:db8:1::/48  ,'         ',    :          :
                   |    |-----------------+    ISP_A    +--+:          :
 2001:db8:1:1::/64 |    |                  ',         ,'    :          :
                   |    |                    '-------'      :          :
H1-----------------| R1 |                                   : INTERNET :
                   |    |                    ,-------,      :          :
 2001:db8:2:1::/64 |    | 2001:db8:2::/48  ,'         ',    :          :
                   |    |-----------------+    ISP_B    +--+:          :
                   +----+                  ',         ,'    :          :
                                             '-------'       \        /
                                                              --------
```

Alphix:
> * If RFC8475 says we cannot set valid lifetime to 0 unless the RA is authenticated, how does this RFC rationalise it? I guess there is no broad support for SeND (I've never seen it), so this RFC just does ¯\_ (ツ)_/¯

I assume you mean [RFC8475, §3.1.2](https://www.rfc-editor.org/rfc/rfc8475#section-3.1.2), which refers to [RFC4862, §5.5.3](https://www.rfc-editor.org/rfc/rfc4862#section-5.5.3) where paragraph `e-2)` says:

```
2.  If RemainingLifetime is less than or equal to 2 hours, ignore
    the Prefix Information option with regards to the valid
    lifetime, unless the Router Advertisement from which this
    option was obtained has been authenticated (e.g., via Secure
    Neighbor Discovery [RFC3971]).  If the Router Advertisement
    was authenticated, the valid lifetime of the corresponding
    address should be set to the Valid Lifetime in the received
    option.
```

Some notes:

1. Like you said, I've never seen `SEND` in the wild even though the standard ([RFC3971](https://www.rfc-editor.org/rfc/rfc3971)) is 20+ years old. It requires a pre-existing certificate infrastructure and I doubt it'll ever be widely deployed (at least not in it's current format).
2. RFC8475 post-dates RFC4862, so I guess it could be seen as implicitly updating the latter 😄 
3. I think the target audience of RFC8475 ("Using Conditional Router Advertisements for <ins>Enterprise</ins> Multihoming") and RFC9096 ("Improving the Reaction of <ins>Customer Edge Routers</ins> to IPv6 Renumbering Events") is quite different.
4. Quite a lot of RFCs just go "rogue RAs are bad, mkay" (point in case, [RFC9762, §10](https://www.rfc-editor.org/rfc/rfc9762#section-10)). And the only realistic solution that works right now is `RA-Guard` ([RFC6105](https://www.rfc-editor.org/rfc/rfc6105.html), that's what I'm using). But `RA-Guard` is:
   a) outside the scope of `odhcpd`, since it realistically needs to be implemented in a L2 switch
   b) not something one could expect a normal consumer network to support

It's interesting to note that RFC6105 says in the abstract (note the reference to an `alternative...to SEND`):

```
Routed protocols are often susceptible to spoof attacks.  The
canonical solution for IPv6 is Secure Neighbor Discovery (SEND), a
solution that is non-trivial to deploy.  This document proposes a
light-weight alternative and complement to SEND based on filtering in
the layer-2 network fabric
```

And e.g. RFC9762, §10 says:
```
It should be noted that if the network allows rogue RAs to be sent,
the attacker would be able to disrupt hosts' connectivity anyway,
so this document doesn't introduce any fundamentally new security
considerations.
```

Which makes sense. If you have rogue routers on your network, RAs with zero lifetime prefixes is just a very small part of the problem. And the realistic solution today (if rogue RAs is a concern) is elsewhere.

More importantly, this issue has already been recognised by the IETF, and they're working on changing it, see [draft-ietf-6man-slaac-renum-10, §5.4](https://datatracker.ietf.org/doc/html/draft-ietf-6man-slaac-renum-10#section-5.4), which explicitly refers to RFC9096 and which is referred to by e.g. [RFC8978, §2.4](https://www.rfc-editor.org/rfc/rfc8978.html#section-2.4) and has already been used as the rationale to [implement the non-RFC4862 behaviour in the Linux kernel](https://patchwork.ozlabs.org/project/netdev/patch/20200419122457.GA971@archlinux-current.localdomain/).

IOW, that part of RFC4862 is on it's way out and can be ignored.

Noltari:
Merged, thanks @Alphix @systemcrash

PR #258, opened 2025-09-27
odhcpd: update cmake file

Alphix:
Now that the minimum cmake version has been bumped to 3.13, the cmake file can be modernized a bit. Although it might look like a lot of changes, most of them are quite straightforward.

Every library is located in a consistent manner (allowing command-line overrides of library locations, useful for local development, and also makes it trivial to do static linking, if desired).

The compiler flags have been broken up to have one per line (making it easy in the future to add/remove flags, which would create a simple one-line diff).

Although it might look like this bumps the C standard level, cmake is actually smart enough to pick an earlier version if C11 isn't supported by the compiler (quite unlikely on any versions of gcc currently in use in OpenWrt and even on old distros).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,58 +1,92 @@
 cmake_minimum_required(VERSION 3.13)
 cmake_policy(SET CMP0015 NEW)
 
+
 # Project Definition
-project(odhcpd C)
-
-set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g3 -std=gnu99")
-
-FIND_PATH(ubox_include_dir libubox/uloop.h)
-FIND_PATH(libnl-tiny_include_dir netlink-generic.h PATH_SUFFIXES libnl-tiny)
-INCLUDE_DIRECTORIES(${ubox_include_dir} ${libnl-tiny_include_dir})
-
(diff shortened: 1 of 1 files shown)


Alphix:
@dedeckeh could you have a look at this PR? I'm mostly interested in the cmake changes to simplify devel work, they shouldn't really affect a normal OpenWrt build (and I've test built `odhcpd` with this patch applied in the OpenWrt build system as well).

Alphix:
Note: this needs to be updated once PR #256 lands

Alphix:
Updated and rebased post-PR #256 

@Noltari @systemcrash Care to have a look? I'm no `cmake` expert, but this file seems much more ergonomic for me (local devel builds) and it has no impact on a normal OpenWrt builtroot build

Noltari:
> Updated and rebased post-PR #256
> 
> @Noltari @systemcrash Care to have a look? I'm no `cmake` expert, but this file seems much more ergonomic for me (local devel builds) and it has no impact on a normal OpenWrt builtroot build

@Alphix TBH I'm not a `cmake` expert either, but I will try to take a look when I have time...

Alphix:
> @Alphix TBH I'm not a `cmake` expert either, but I will try to take a look when I have time...

I also don't ever want to become one...

Noltari:
Merged https://github.com/openwrt/odhcpd/commit/c2810fe8b681238cab8a3be1a8ae0cc222514306, thanks @Alphix!

PR #259, opened 2025-09-27
dhcpv4: allow lease takeover (bugfix)

Alphix:
With commit c9816de148cc3e8096e67519d2c081ffeb9b4857 (PR #251), multiple MAC addresses are supported per static lease. I tested the three basic scenarios:

```
config host
        option name 'test'
        option ip '192.168.44.4'
        list mac '<macA>'
        list mac '<macB>'
```
```
1) REQUEST with <macA> or <macB> - works

2) REQUEST with <macA> or <macB>
   RELEASE with <macA> or <macB>
   REQUEST with <macB> or <macA> - works

3) REQUEST with <macA> or <macB>
   REQUEST with <macB> or <macA> - semed to work
```

Due to a bug in my test script, I didn't notice that in the third case, odhcpd will return a NACK (because the addr is taken). This is contrary to how dnsmasq currently works (the assignment is simply handed over to the new client and the old one is silently dropped).

This fixes test case 3) by freeing any old assignments for a given lease.

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -1111,6 +1111,20 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 	struct lease *l = config_find_lease_by_mac(mac);
 	time_t now = odhcpd_time();
 
+	/*
+	 * If we found a static lease cfg, but no old assignment for this
+	 * hwaddr, we need to clear out any old assignments given to other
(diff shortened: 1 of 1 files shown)


Alphix:
@thess gentle ping :)

thess:
Merged and closed by bc9f9d93d4d6b8feb7b19235d7f0371480fc679d

PR #263, opened 2025-10-01
odhcpd: convert README to markdown

Alphix:
Convert the README file to GitHub style markdown. I don't think it affects the readability of the plain text file much (or not at all, rather). But it makes it much easier to read for people looking at the GitHub repo.

I've intentionally kept the fancy markup at a minimum and also haven't changed any help text messages in order to not get bogged down in how each help message should read. That is something which can be addressed later.


Alphix:
To see the results, compare the [before](https://github.com/openwrt/odhcpd) and [after](https://github.com/Alphix/odhcpd/tree/readme)

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #264, opened 2025-10-02
dhcpv4: cleanup part 1

Alphix:
I'm working on cleaning up `dhcpv4.c`. As a first step, I'd like to reorganise the file a bit by moving the functions around. Right now, it's written in a kind of upside-down manner compared to what you'd expect from a C file (i.e. the most generic/init style functions are at the top, and the utility functions are further down, where you'd normally expect it to be the other way around). This currently necessitates a lot of forward declarations.

This might look like a massive change (and the diff looks a bit confusing), but looking at the individual commits, it's mostly just shifting functions around.

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -329,7 +329,7 @@ static bool dhcpv4_assign(struct interface *iface, struct dhcp_assignment *a,
 	uint32_t count = end - start + 1;
 	uint32_t seed = 0;
 	bool assigned;
-	char buf[INET_ADDRSTRLEN];
+	char ipv4_str[INET_ADDRSTRLEN];
 
 	/* Preconfigured IP address by static lease */
 	if (a->addr) {
@@ -338,7 +338,7 @@ static bool dhcpv4_assign(struct interface *iface, struct dhcp_assignment *a,
 
 		if (assigned)
 			syslog(LOG_DEBUG, "Assigning static IP: %s",
-			       inet_ntop(AF_INET, &a->addr, buf, sizeof(buf)));
+			       inet_ntop(AF_INET, &a->addr, ipv4_str, sizeof(ipv4_str)));
 
 		return assigned;
 	}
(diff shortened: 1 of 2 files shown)


src/dhcpv4.c
```diff
@@ -1252,3 +1198,53 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
...
+				free_assignment(a);
+		}
+	}
+	uloop_timeout_set(event, 1000);
+}
+
+/* Create socket and register events */
+int dhcpv4_init(void)
+{
+	static struct netevent_handler dhcpv4_netevent_handler = { .cb = dhcpv4_netevent_cb, };
```
Noltari: I know this was copy & pasted, but we can change it to:
`static struct netevent_handler dhcpv4_netevent_handler = {.cb = dhcpv4_netevent_cb};`
That way it matches the valid_until_timeout declaration.

src/dhcpv4.c
```diff
@@ -768,14 +757,16 @@ static bool dhcpv4_assign(struct interface *iface, struct dhcp_assignment *a,
 	uint32_t count = end - start + 1;
 	uint32_t seed = 0;
 	bool assigned;
+	char buf[INET_ADDRSTRLEN];
```
Noltari: `buf` is a bit generic. Can we change it to `ipv4_str` or something more descriptive?

src/dhcpv4.c
Alphix: I had to stare at your suggested line 10 times before I got it...fixed :)

src/dhcpv4.c
Alphix: Oki, but I did it as a separate patch since I'd have to rebase almost all of them otherwise and that's a real PITA

Noltari: > Oki, but I did it as a separate patch since I'd have to rebase almost all of them otherwise and that's a real PITA

It's fine, no worries

Noltari (approved):
LGTM as this will save headaches for future dhcpv4 changes

Alphix:
> LGTM as this will save headaches for future dhcpv4 changes

And it would be nice to have this go in before other PRs that touch `dhcpv4.c`, since rebasing this patchset is not exactly easy...

Noltari:
> And it would be nice to have this go in before other PRs that touch `dhcpv4.c`, since rebasing this patchset is not exactly easy...

@Alphix your wish is my command :)

PR #266, opened 2025-10-03
dhcpv4: cleanup part 2

Alphix:
Now that #264 has been merged, this is the second round of cleanups to dhcpv4.c.

It might be weird to see a "FIXME" in the code, but it is an issue I spotted while working on this cleanup - the DHCP cookie in client messages is never checked. I didn't want to change that behaviour in this PR though, since it is something that I think warrants a separate PR (I have no idea if there might be broken DHCPv4 clients in the wild that send incorrect cookie values).


Noltari:
Merged, thanks @Alphix!

PR #268, opened 2025-10-05
odhcpd: print compiled-in features in help message

Alphix:
This is mostly useful for hacking, as a way to quickly determine from the command-line how odhcpd has been built. But I also plan to use it in LuCI [1].

[1] https://github.com/openwrt/luci/blob/master/modules/luci-base/root/usr/share/rpcd/ucode/luci


Noltari:
Merged https://github.com/openwrt/odhcpd/commit/63617a2d2df02054d9d55b2aa2dd56249339fdf1, thanks @Alphix!

PR #270, opened 2025-10-07
odhcpd: misc ubus fixes

Alphix:
This adds a bunch of little fixes and cleanups to the ubus code and also documents the ubus interface so that people won't have to dig through the code to find out how to interact with `odhcpd` via ubus.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2100,9 +2100,7 @@ void odhcpd_reload(void)
 
 	vlist_flush(&leases);
 
-#ifdef WITH_UBUS
 	ubus_apply_network();
-#endif
 
 	bool any_dhcpv6_slave = false, any_ra_slave = false, any_ndp_slave = false;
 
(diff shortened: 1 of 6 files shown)


Noltari (approved):
LGTM

Alphix:
Pushed a new version (I missed one comment line in `src/odhcpd.c`)

src/config.c
```diff
@@ -2093,9 +2093,7 @@ void odhcpd_reload(void)
 
 	vlist_flush(&leases);
 
-#ifdef WITH_UBUS
 	ubus_apply_network();
-#endif
```
systemcrash, line 2105: could you briefly walk me though this @Alphix ? redundant?

systemcrash (approved):

src/config.c
Alphix: I changed the headers so that they contain dummy functions when ubus is not compiled in... That way the code looks the same for both cases but it's a `nop` in case ubus isn't compiled in..

systemcrash: Yep. See it now. The ifdef was hidden from the diff.

Noltari:
Merged, thanks @Alphix

Issue #269, opened 2025-10-07
odhcpd SLAAC broken following #256 for subnets larger /64


GoetzGoerisch:
Using odhcpd with the update from #256 breaks SLAAC for subnets >`/64`.

If a subnet larger /64 is configured the prefix /64 to be anounced to allow hosts to use SLAAC are only announces with lifetime of `0`

/etc/config/network
```
config interface 'main'
	option proto 'static'
	option device 'br-lan.20'
	option ipaddr '10.1.20.1'
	option netmask '255.255.255.0'
	option ip6assign '60'
	option ip6hint '20'
	list ip6class 'local'
	list ip6class 'wan6'
```

/etc/config/dhcp
```
config dhcp 'main'
	option instance 'main_dns'
	option interface 'main'
	option start '20'
	option limit '200'
	option leasetime '15m'
	option ra 'server'
	option preferred_lifetime '90m'
	option ra_pref64 '64:ff9b::/96'
	list domain 'main.internal'
	option ra_useleasetime '1'
	option dhcpv6 'server'
	list ra_flags 'managed-config'
	list ra_flags 'other-config'
...

config dnsmasq 'main_dns'
	option domainneeded '1'
	option rebind_protection '1'
	option local '/main.internal/'
	option domain 'main.internal'
	option expandhosts '1'
	option authoritative '1'
	option leasefile '/tmp/dhcp.leases.main'
	option resolvfile '/tmp/resolv.conf.d/resolv.conf.auto'
	list interface 'main'
	option localservice '0'
	option allservers '1'
	list notinterface 'loopback'
	option strictorder '1'
	option stripmac '1'
	option stripsubnet '1'
	option rebind_localhost '1'
	list cache_rr 'HTTPS'
	list cache_rr 'PTR'
	list cache_rr 'SOA'
	option ignore_hosts_dir '1'
	list addnhosts '/tmp/hosts/odhcpd'
	option boguspriv '0'
...

config odhcpd 'odhcpd'
	option maindhcp '0'
	option leasefile '/tmp/hosts/odhcpd'
	option leasetrigger '/usr/sbin/odhcpd-update'
	option loglevel '6'
	option piofolder '/tmp/odhcpd-piofolder'

```

log:
```
Tue Oct  7 09:08:10 2025 daemon.warn odhcpd[24932]: rfc9096: br-lan.20: renew 2a02:xxxx:xxxx:xxxx::/64
[...]

PR #271, opened 2025-10-07
router: fix SLAAC on subnets > 64

Noltari:
Fix SLAAC on subnets with prefix > 64.

More info: https://github.com/openwrt/odhcpd/issues/269#issuecomment-3376633048

CC @GoetzGoerisch


GoetzGoerisch (approved):
run-tested: x86/64

Thank you!

PR #272, opened 2025-10-09
odhcpd: fix a compilation error

Alphix:
odhcpd_get_interface_dns_addr() calls odhcpd_get_interface_linklocal_addr(), the former takes a const struct interface ptr, the latter takes a non-const struct interface ptr.

The end result is an unhappy compiler.


Alphix:
@Noltari ping :)

Noltari (approved):
LGTM

PR #273, opened 2025-10-09
odhcpd: switch to simple log macros, support stderr logging

Alphix:
This is mostly for developer ergonomics. My `/var/log/debug` is 150 megs and counting :D

There is one fix wrt. loglevel handling that I'd consider important though. Right now, a loglevel set in the cfg file will override one specified on the command line, which doesn't seem right.

Also, `syslog.h` is already included from `odhcpd.h`, so no need to include it again in a bunch of places.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -41,6 +41,8 @@ struct config config = {
 	.ra_piofolder_fd = -1,
 	.uci_cfgfile = "dhcp",
 	.log_level = LOG_WARNING,
+	.log_level_cmdline = false,
+	.log_syslog = true,
 };
 
 #define START_DEFAULT	100
@@ -423,9 +425,11 @@ static void set_config(struct uci_section *s)
 	if ((c = tb[ODHCPD_ATTR_LOGLEVEL])) {
 		int log_level = (blobmsg_get_u32(c) & LOG_PRIMASK);
 
-		if (config.log_level != log_level) {
+		if (config.log_level != log_level && config.log_level_cmdline) {
 			config.log_level = log_level;
-			setlogmask(LOG_UPTO(config.log_level));
+			if (config.log_syslog)
+				setlogmask(LOG_UPTO(config.log_level));
+			notice("Log level set to %d\n", config.log_level);
 		}
 	}
 }
--- a/src/config.c
+++ b/src/config.c
@@ -8,7 +8,6 @@
 #include <net/if.h>
 #include <string.h>
 #include <sys/stat.h>
-#include <syslog.h>
 
 #include <uci.h>
 #include <uci_blob.h>
@@ -479,14 +478,14 @@ static bool parse_duid(struct duid *duid, struct blob_attr *c)
 
 		/* IAID = uint32, RFC8415, §21.4, §21.5, §21.21 */
 		if (iaid_str_len < 1 || iaid_str_len > 2 * sizeof(uint32_t)) {
-			syslog(LOG_ERR, "Invalid IAID length '%s'", iaid_str);
+			error("Invalid IAID length '%s'", iaid_str);
 			return false;
 		}
 
 		errno = 0;
 		duid->iaid = strtoul(iaid_str, &end, 16);
 		if (errno || *end != '\0') {
-			syslog(LOG_ERR, "Invalid IAID '%s'", iaid_str);
+			error("Invalid IAID '%s'", iaid_str);
 			return false;
 		}
 
@@ -495,13 +494,13 @@ static bool parse_duid(struct duid *duid, struct blob_attr *c)
 	}
 
 	if (duid_str_len < 2 || duid_str_len > DUID_MAX_LEN * 2 || duid_str_len % 2) {
-		syslog(LOG_ERR, "Invalid DUID length '%.*s'", (int)duid_str_len, duid_str);
+		error("Invalid DUID length '%.*s'", (int)duid_str_len, duid_str);
 		return false;
 	}
 
 	duid_len = odhcpd_unhexlify(duid->id, duid_str_len / 2, duid_str);
 	if (duid_len < 0) {
-		syslog(LOG_ERR, "Invalid DUID '%.*s'", (int)duid_str_len, duid_str);
+		error("Invalid DUID '%.*s'", (int)duid_str_len, duid_str);
 		return false;
 	}
 
@@ -690,10 +689,10 @@ static int parse_dnr_str(char *str, struct interface *iface)
 	if (!priority) {
 		goto err;
 	} else if (sscanf(priority, "%" SCNu16, &dnr.priority) != 1) {
-		syslog(LOG_ERR, "Unable to parse priority '%s'", priority);
+		error("Unable to parse priority '%s'", priority);
 		goto err;
 	} else if (dnr.priority == 0) {
-		syslog(LOG_ERR, "Invalid priority '%s'", priority);
+		error("Invalid priority '%s'", priority);
 		goto err;
 	}
 
@@ -707,13 +706,13 @@ static int parse_dnr_str(char *str, struct interface *iface)
 		adn[adn_len - 1] = '\0';
 
 	if (adn_len >= sizeof(adn_buf)) {
-		syslog(LOG_ERR, "Hostname '%s' too long", adn);
+		error("Hostname '%s' too long", adn);
 		goto err;
 	}
 
 	adn_len = dn_comp(adn, adn_buf, sizeof(adn_buf), NULL, NULL);
 	if (adn_len <= 0) {
-		syslog(LOG_ERR, "Unable to parse hostname '%s'", adn);
+		error("Unable to parse hostname '%s'", adn);
 		goto err;
 	}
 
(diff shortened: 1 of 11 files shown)


Alphix:
@systemcrash @Noltari No idea if either one of you will find this controversial. Bad idea? Good idea? I find it very practical to be able to not spam syslog while hacking...

src/config.c
```diff
@@ -419,9 +420,11 @@ static void set_config(struct uci_section *s)
 	if ((c = tb[ODHCPD_ATTR_LOGLEVEL])) {
 		int log_level = (blobmsg_get_u32(c) & LOG_PRIMASK);
 
-		if (config.log_level != log_level) {
+		if (config.log_level != log_level && config.log_level_cmdline) {
 			config.log_level = log_level;
-			setlogmask(LOG_UPTO(config.log_level));
+			if (config.log_syslog)
+				setlogmask(LOG_UPTO(config.log_level));
+			notice("Log level set to %d\n", config.log_level);
```
systemcrash, line 432: Would this ever get logged if we're only ever at error level? Chicken and egg, right?

src/config.c
```diff
@@ -1938,12 +1936,8 @@ void config_load_ra_pio(struct interface *iface)
...
 		pio->length = pio_len;
 		pio->lifetime = pio_lt;
-		syslog(LOG_INFO,
-			"rfc9096: %s: load %s/%u (%u)",
-			iface->ifname,
-			pio_str,
-			pio_len,
-			ra_pio_lifetime(pio, now));
+		info("rfc9096: %s: load %s/%u (%u)", iface->ifname, pio_str,
+		     pio_len, ra_pio_lifetime(pio, now));
```
systemcrash: I find the 'display item per line' formatting easier to correlate (read). The change is fine, the formatting should be retained I think.

src/config.c
Alphix: Yeah, not sure what the best solution is....blorting stuff at `syslog(LOG_EMERG, ...` and `fprintf(stderr, ...` both seem bad. I just decided that the message isn't all that important so it's ok if it gets lost...

src/config.c
Alphix: Fair enough, I can change that in v2

src/config.c
systemcrash: Agreed. Leave as is.

src/config.c
Alphix: Done

Alphix:
@systemcrash Ok, I pushed a new version where the formatting of log msgs is changed as little as possible.

systemcrash:
I'm fine with tab alignment where you used spaces on subsequent lines. All good. 

See what @Noltari says.

systemcrash:
I figure my recent commits also need amending here :)

Alphix:
I know, will do when I have time :)

Alphix:
> I figure my recent commits also need amending here :)

@systemcrash rebased and updated to cover your commits as well (and ping @Noltari as well)

src/dhcpv4.c
```diff
@@ -436,8 +436,8 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
 				/* Create new binding */
 				a = alloc_assignment(0);
 				if (!a) {
-					syslog(LOG_WARNING, "Failed to alloc assignment on interface %s",
-							    iface->ifname);
+					warn("Failed to alloc assignment on interface %s",
+					     iface->ifname);
```
systemcrash, line 440: Maybe just tab indent so it doesn't look weird for those of varying tab widths. Just a nit-pick.

src/odhcpd.c
```diff
@@ -93,29 +93,38 @@ static bool ipv6_enabled(void)
...
-	while ((opt = getopt(argc, argv, "c:l:h")) != -1) {
+	while ((opt = getopt(argc, argv, "c:l:fh")) != -1) {
 		switch (opt) {
 		case 'c':
 			config.uci_cfgfile = realpath(optarg, NULL);
 			fprintf(stderr, "Configuration will be read from %s\n", config.uci_cfgfile);
 			break;
 		case 'l':
 			config.log_level = (atoi(optarg) & LOG_PRIMASK);
+			config.log_level_cmdline = true;
```
systemcrash, line 125: Are these two booleans mutex? Or can/should they be?

src/dhcpv4.c
systemcrash: which I forgot I already wrote above :D

src/odhcpd.c
Alphix: No, one is the actual log level and the other one records whether the log level was set via cmdline (i.e. shouldn't be overridden by e.g. cfg files)

src/dhcpv4.c
Alphix: Varying tab widths? Blasphemy...I usually just follow the kernel coding style and let `vim` do the formatting for me 🤷🏻‍♂️

Alphix:
@systemcrash I think this is good to go? It's kind of holding up other PRs I'm working on :)

src/odhcpd.c
systemcrash: All fine by me then.

systemcrash:
@Noltari jam this bad boy in then.

Alphix:
> @Noltari jam this bad boy in then.

Just a question: I thought you had commit rights? So does that mean you want @Noltari to review first? Or am I confused?

systemcrash:
> I thought you had commit rights?

Not in this repo. :/

Alphix:
> > I thought you had commit rights?
> 
> Not in this repo. :/

Do you plan to get commit access? What's the procedure?

src/odhcpd.h
```diff
@@ -69,6 +69,26 @@ struct nl_sock;
...
 
+#define __iflog(lvl, fmt, ...)							\
+	do {									\
+		if (lvl > config.log_level)					\
+			break;							\
+		if (config.log_syslog)						\
+			syslog(lvl, fmt __VA_OPT__(, ) __VA_ARGS__);		\
+		else								\
+			fprintf(stderr, fmt "\n" __VA_OPT__(, ) __VA_ARGS__);	\
+	} while(0)
```
Noltari: @Alphix I was going to merge this PR but then I realized that this is increasing the binary size by a considerable amount without adding real value to the users (only for development purposes).

| bin             | master | no-syslog | test    |
|:---------------:|:------:|:---------:|:-------:|
| odhcpd          | 60486  | 64574     | 60593   |
| odhcpd-ipv6only | 53124  | 56448     | 53252   |

With test being:
```
#define __iflog(lvl, fmt, ...) syslog(lvl, fmt __VA_OPT__(, ) __VA_ARGS__)
```

Therefore, we need to avoid duplicating the code inside the define by adding a new function or we need to add a new define to enable that code on compilation, which you could use for your development/testing but won't be enabled on shipped OpenWrt binaries.

I'm sorry, but some OpenWrt devices are very constrained and we need to be very careful with things like this...

systemcrash: Are those byte units? 4kb?

Perhaps an amendment to the CI which saves the binary (so one could download it for use), and logs its size in the action.

Alphix: Good catch, didn't consider that. I guess the "\n" in the fprintf case means the strings are effectively duplicated. If so, it should be easy to fix. I'll look into it...

Alphix: Separating the "\n" only regained half the extra size....but making a proper function out of __iflog returned all of the additional size in my testing...I've pushed a fix

Alphix: (another option would be a compile-time switch)

Noltari:
Merged, thanks @Alphix!

Noltari:
@Alphix after testing this I've realized that it's broken if we just call odhcpd without `-l <int>` or `-f`.
In OpenWrt, `odhcpd` is called without arguments and therefore everything is set from the uci config.
In that case, even if `config.log_level` has a value, we won't log anything at all...

Alphix:
> @Alphix after testing this I've realized that it's broken if we just call odhcpd without `-l <int>` or `-f`. In OpenWrt, `odhcpd` is called without arguments and therefore everything is set from the uci config. In that case, even if `config.log_level` has a value, we won't log anything at all...

Sorry about that, will look into it this evening

Alphix:
Oh, I think the check in `config.c:set_config()` is wrong, it reads:

```C
if (config.log_level != log_level && config.log_level_cmdline) {
```

But should probably be:
```C
if (config.log_level != log_level && !config.log_level_cmdline) {
```

I'll test later...

Noltari:
> Oh, I think the check in `config.c:set_config()` is wrong, it reads:
> 
> ```c
> if (config.log_level != log_level && config.log_level_cmdline) {
> ```
> 
> But should probably be:
> 
> ```c
> if (config.log_level != log_level && !config.log_level_cmdline) {
> ```
> 
> I'll test later...

I've submitted a PR with a different fix at https://github.com/openwrt/odhcpd/pull/288/files

PR #274, opened 2025-10-10
dhcpv6: support a global DUID

Alphix:
This adds support for a globally defined static DUID.

Currently, odhcpd generates a per-interface DUID-LL, meaning that the DUID is
not stable across interfaces and will change if/when hardware changes.

Supporting a stable DUID brings odhcpd's behaviour in line with [RFC8415, §11](https://datatracker.ietf.org/doc/html/rfc8415#section-11):

```
                                                      ...The DUID is
   designed to be unique across all DHCP clients and servers, and stable
   for any specific client or server.  That is, the DUID used by a
   client or server SHOULD NOT change over time if at all possible; for
   example, a device's DUID should not change as a result of a change in
   the device's network hardware or changes to virtual interfaces...
```

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -137,43 +137,74 @@ static size_t get_preferred_addr(const struct odhcpd_ipaddr *addrs, const size_t
 	return m;
 }
 
+enum {
+	IOV_HDR = 0,
+	IOV_SERVERID,
+	IOV_CLIENTID,
+	IOV_MESSAGE,
+	IOV_AUTH,
+	IOV_TOTAL
+};
+
 static int send_reconf(struct dhcp_assignment *assign)
 {
-	struct {
-		struct dhcpv6_client_header hdr;
-		uint16_t srvid_type;
-		uint16_t srvid_len;
-		uint16_t duid_type;
-		uint16_t hardware_type;
-		uint8_t mac[6];
-		uint16_t msg_type;
-		uint16_t msg_len;
-		uint8_t msg_id;
-		struct dhcpv6_auth_reconfigure auth;
-		uint16_t clid_type;
-		uint16_t clid_len;
-		uint8_t clid_data[128];
-	} __attribute__((packed)) reconf_msg = {
-		.hdr = {DHCPV6_MSG_RECONFIGURE, {0, 0, 0}},
-		.srvid_type = htons(DHCPV6_OPT_SERVERID),
-		.srvid_len = htons(10),
-		.duid_type = htons(3),
-		.hardware_type = htons(1),
-		.msg_type = htons(DHCPV6_OPT_RECONF_MSG),
-		.msg_len = htons(1),
-		.msg_id = DHCPV6_MSG_RENEW,
-		.auth = {htons(DHCPV6_OPT_AUTH),
-				htons(sizeof(reconf_msg.auth) - 4), 3, 1, 0,
-				{htonl(time(NULL)), htonl(++serial)}, 2, {0}},
-		.clid_type = htons(DHCPV6_OPT_CLIENTID),
-		.clid_len = htons(assign->clid_len),
-		.clid_data = {0},
-	};
 	struct interface *iface = assign->iface;
+	struct dhcpv6_client_header hdr = {
+		.msg_type = DHCPV6_MSG_RECONFIGURE,
+		.transaction_id = { 0, 0, 0 },
+	};
+	struct {
+		uint16_t code;
+		uint16_t len;
+		uint8_t data[DUID_MAX_LEN];
+	} _packed serverid = {
+		.code = htons(DHCPV6_OPT_SERVERID),
+		.len = 0,
+		.data = { 0 },
+	};
+	struct {
+		uint16_t code;
+		uint16_t len;
+		uint8_t data[DUID_MAX_LEN];
+	} _packed clientid = {
+		.code = htons(DHCPV6_OPT_CLIENTID),
+		.len = htons(assign->clid_len),
+		.data = { 0 },
+	};
+	struct {
+		uint16_t code;
+		uint16_t len;
+		uint8_t id;
+	} _packed message = {
+		.code = htons(DHCPV6_OPT_RECONF_MSG),
+		.len = htons(1),
+		.id = DHCPV6_MSG_RENEW,
+	};
+	struct dhcpv6_auth_reconfigure auth =  {
+		.type = htons(DHCPV6_OPT_AUTH),
+		.len = htons(sizeof(struct dhcpv6_auth_reconfigure)),
+		.protocol = 3,
+		.algorithm = 1,
+		.rdm = 0,
+		.replay = { htonl(time(NULL)), htonl(++serial) },
+		.reconf_type = 2,
+		.key = { 0 },
+	};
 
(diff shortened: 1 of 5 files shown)


systemcrash:
DUID options are in now.

Alphix:
I know, I need to rework this a bit... And test it, but I'm working on it

Alphix:
> I know, I need to rework this a bit... And test it, but I'm working on it

And thanks to the latest PRs being merged, I rebased and reworked this a bit. Should be good to go now.

Noltari (approved):
LGTM

@systemcrash any objections?

src/dhcpv6-ia.c
```diff
@@ -156,43 +156,79 @@ static size_t get_preferred_addr(const struct odhcpd_ipaddr *addrs, const size_t
...
-		.auth = {htons(DHCPV6_OPT_AUTH),
-				htons(sizeof(reconf_msg.auth) - 4), 3, 1, 0,
-				{htonl(time(NULL)), htonl(++serial)}, 2, {0}},
-		.clid_type = htons(DHCPV6_OPT_CLIENTID),
-		.clid_len = htons(assign->clid_len),
-		.clid_data = {0},
+		uint16_t code;
+		uint16_t len;
+		uint8_t data[DUID_MAX_LEN];
+	} _packed serverid = {
```
systemcrash, line 160: Is there a semantic difference here to `__attribute__((packed))`?

src/dhcpv6-ia.c
```diff
@@ -156,43 +156,79 @@ static size_t get_preferred_addr(const struct odhcpd_ipaddr *addrs, const size_t
...
-	struct iovec iov = {&reconf_msg, sizeof(reconf_msg) - 128 + assign->clid_len};
+	memcpy(clientid.data, assign->clid_data, assign->clid_len);
+
+	struct iovec iov[IOV_MAX] = {
+		[IOV_HDR] = { &hdr, sizeof(hdr) },
+		[IOV_SERVERID] = { &serverid, sizeof(serverid) },
+		[IOV_CLIENTID] = { &clientid, sizeof(clientid) },
+		[IOV_MESSAGE] = { &message, sizeof(message) },
+		[IOV_AUTH] = { &auth, sizeof(auth) },
+	};
```
systemcrash, line 212: *chefs kiss*

src/dhcpv6.c
```diff
@@ -364,27 +365,38 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
-		.clientid_buf = {0}
+		.serverid_length = 0,
+		.serverid_buf = { 0 },
+	};
+
+	if (config.default_duid_len > 0) {
+		memcpy(dest.serverid_buf, config.default_duid, config.default_duid_len);
+		dest.serverid_length = htons(config.default_duid_len);
+	} else {
+		uint8_t duid_ll_hdr[] = { 0x00, 0x03, 0x00, 0x01 };
```
systemcrash: Does this signify 0x1 for ethernet?

src/dhcpv6-ia.c
Alphix: No, none whatsoever, we've had a `#define _packed __attribute__((packed))` in `odhcpd.h` for a long time...it's just more readable, I usually convert the longer style over piece-by-piece...but eventually there should only be `_packed` in the codebase.

src/dhcpv6.c
Alphix: Yep, it's unchanged from the old code. It's a [DUID-LL](https://datatracker.ietf.org/doc/html/rfc8415#section-11.4), which starts with 2 byte DUID-type (0x0003 = DUID-LL) and then hwtype (0x0001 = Ethernet...if you want to be precise, the [IANA description](https://www.iana.org/assignments/arp-parameters/arp-parameters.xhtml#arp-parameters-2) is `ethernet (10mb)` 😄)

src/dhcpv6-ia.c
Alphix: Am I supposed to resolve this conversation? I'm not sure how to... :)

Noltari: > Am I supposed to resolve this conversation? I'm not sure how to... :)

With another kiss... obviously 😘

Alphix: 💄💋

src/dhcpv6.c
systemcrash: I'm OK with this - although for clarity it'd be nice to have those defined so it doesn't look like we're just jamming in random bytes :)

src/dhcpv6-ia.c
systemcrash: Ah, great.

src/dhcpv6.c
systemcrash: Maybe that's awkward though?

Alphix: I could jam in `ARPHRD_ETHER` as the fourth byte....for the second byte I'd have to create new `defines`...or I could go for a uint16_t, I guess...

systemcrash: Either - uint16_t is in line with the RFC.

Alphix: Ok, I went with a `uint16_t` and a new `enum`

Alphix:
@systemcrash have another look :)

src/dhcpv6-ia.c
```diff
@@ -156,43 +156,79 @@ static size_t get_preferred_addr(const struct odhcpd_ipaddr *addrs, const size_t
...
+
+	if (config.default_duid_len > 0) {
+		memcpy(serverid.data, config.default_duid, config.default_duid_len);
+		serverid.len = htons(config.default_duid_len);
+	} else {
+		uint16_t duid_ll_hdr[] = { htons(DUID_TYPE_LL), htons(ARPHRD_ETHER) };
+		memcpy(serverid.data, duid_ll_hdr, sizeof(duid_ll_hdr));
+		odhcpd_get_mac(iface, &serverid.data[sizeof(duid_ll_hdr)]);
+		serverid.len = htons(sizeof(duid_ll_hdr) + ETH_ALEN);
+	}
```
systemcrash, line 202: We do this twice. Would it make sense to house it in a function? Or does the function signature cause it to grow 300%?

Alphix: Would probably shrink the binary size, rather...

Alphix: Ok, I gave it a go, it is doable, but in the future I'm going to change this further - read: making better use of IOVs by stashing the fallback `DUID-LL` in `struct interface` (`odhcpd` already has very nice `netlink` support that tells us if the `MAC` changes, so it'll always be correct). That way we won't do any `memcpy()`s at all, and will only do one `ioctl` call to get the MAC address (instead of one every time we generate a message). Then the function would disappear and the code will just end up looking like:

```C
struct iovec iov[IOV_TOTAL] = {
    ...
    [IOV_SERVERID_HDR] = { &serverid_hdr, sizeof(serverid_hdr) },
    [IOV_SERVERID_DATA] = { config.default_duid_len ? config.default_duid : iface->duid,
                            config.default_duid_len ? config.default_duid_len : iface->duid_len },
    ...
```

So if you don't mind, I'd prefer keeping it as-is right now

systemcrash: Sure. As-is.

PR #275, opened 2025-10-10
config: remove legacy/maindhcp opts

Alphix:
If someone wants to use odhcpd as a DHCPv4 server on a standard OpenWrt install they would have to:

a) remove dnsmasq, or disable its DHCPv4 server
b) remove odhcpd-ipv6only
c) install odhcpd
d) uci set dhcp.odhcpd.legacy=1
e) uci set dhcp.odhcpd.maindhcp=1
f) uci set dhcp.lan.start=100 (usually already done)
g) *not* uci set dhcp.lan.dhcpv4=server

The last part is because dhcpv4=server will be automagically set when dhcp.lan.start is set *and* dhcp.odhcpd.legacy and dhcp.odhcpd.maindhcp are set.

This seems a bit superfluous. Installing odhcpd instead of odhcpd-ipv6only and setting dhcp.lan.dhcpv4=server already seems like a pretty clear indication of what the user wants, so remove the extra options and magic dhcpv4=server logic.

(This also needs separate changes to odhcpd.defaults in the main openwrt repo).

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,6 @@ and may also receive information from ubus
 
 | Option	| Type	|Default| Description |
 | :------------ | :---- | :----	| :---------- |
-| legacy	| bool	| 0	| Enable DHCPv4 if start but no dhcpv4 option set |
-| maindhcp	| bool	| 0	| Use odhcpd as the main DHCPv4 service |
 | leasefile	| string|	| DHCP/v6 lease/hostfile |
 | leasetrigger	| string|	| Lease trigger script |
 | hostsfile	| string|	| DHCP/v6 hostfile |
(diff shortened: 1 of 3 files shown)


saldry:
https://github.com/openwrt/openwrt/blob/main/package/network/services/dnsmasq/files/dnsmasq.init#L947

dnsmasq init reads `dhcp.odhcpd.maindhcp` and then decides whether to act as dhcpv4 server.

Such an option seems necessary as users may switch between dnsmasq and odhcpd with both apps installed.

Alphix:
> https://github.com/openwrt/openwrt/blob/main/package/network/services/dnsmasq/files/dnsmasq.init#L947
> 
> dnsmasq init reads `dhcp.odhcpd.maindhcp` and then decides whether to act as dhcpv4 server.
> 
> Such an option seems necessary as users may switch between dnsmasq and odhcpd with both apps installed.

Good, thanks for the heads-up, I had a feeling there must be more to it...I'll look into that...

Noltari (changes requested):
As pointed out by @saldry, `maindhcp` option is used to switch between dnsmasq & odhcpd for DHCPv4.

Alphix:
Yeah, I know, I'll rework the PR

Alphix:
This is superseded by #294

PR #276, opened 2025-10-12
github: add CI build

Noltari:
Add Github CI supporting different architectures and odhcpd build options.

https://github.com/Noltari/odhcpd/actions/runs/18445914417

CC @Alphix @systemcrash

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,166 @@
+name: odhcpd
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  build:
+    name: Build ${{ matrix.arch }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: aarch64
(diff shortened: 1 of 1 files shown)


Alphix:
Excellent stuff!

Question: would it make sense to run "make clean" between each "cmake; make; make install" invocation? I have the impression that cmake can sometimes be a bit too lazy in rebuilding things that should be rebuilt with different flags...

Noltari:
> Excellent stuff!
> 
> Question: would it make sense to run "make clean" between each "cmake; make; make install" invocation? I have the impression that cmake can sometimes be a bit too lazy in rebuilding things that should be rebuilt with different flags...

I don't think that's needed since the building directories passed to cmake are different...

Noltari:
BTW, I updated the PR with multiple architectures support.

Alphix:
> I don't think that's needed since the building directories passed to cmake are different...

Ah, good, I checked it quickly on my phone, makes sense

systemcrash:
Welcome addition!

PR #278, opened 2025-10-13
dhcpv4: cleanups, use iovec

Alphix:
This is the first (and biggest) PR I have cooking for `dhcpv4.c`. There's a lot of changes, but what it boils down to is getting rid of `dhcpv4_put()` (which copies a lot of data around on the stack before sending it out, which I consider to be a poor choice...I think DHCPv4 has definitely gotten less love than its DHCPv6 counterpart).

In the process, the monster function that deals with incoming DHCPv4 requests is substantially reworked so that it has two distinct phases, parse incoming options (one switch statement), determine whether a reply is due, then preparing the outgoing options (another switch statement).

This will make it a lot easier to to hack on `dhcpv4.c` in the future (compare the files in the pre and post PR versions and I think you'll agree).

Right now, this is based on top of the other PR which reworks `syslog` messages (PR #273), so that needs attention first, but I'm posting this now already so that @systemcrash and @Noltari can have a look :)

There's another follow-up PR or two after this one, but that can wait 😄


systemcrash:
The change set looks great in general. Good use of iovec. Looking forward to a more maintainable interface. 

Will be easier to review once the log stuff goes in.

systemcrash:
ping @Alphix for rebase

Alphix:
> ping @Alphix for rebase

🫡

Alphix:
@Noltari when you have time, please have a look :)
@systemcrash should be easier to review now

src/dhcpv4.c
```diff
@@ -218,20 +264,19 @@ static void dhcpv4_fr_send(struct dhcp_assignment *a)
...
-		dhcpv4_put(&fr_msg, &cursor, DHCPV4_OPT_END, 0, NULL);
+		iov[IOV_FR_SERVERID].iov_len = sizeof(fr_serverid);
+		dhcpv4_add_padding(iov, ARRAY_SIZE(iov));
 	}
 
-	if (sendto(iface->dhcpv4_event.uloop.fd, &fr_msg, PACKET_SIZE(&fr_msg, cursor),
-		   MSG_DONTWAIT, (struct sockaddr*)&dest, sizeof(dest)) < 0)
-		error("Failed to send %s to %s - %s: %m", dhcpv4_msg_to_string(msg),
+	if (dhcpv4_send_reply(iov, ARRAY_SIZE(iov), (struct sockaddr *)&dest, sizeof(dest),
+			      &a->iface->dhcpv4_event.uloop.fd) < 0)
```
systemcrash, line 275: nitpick: tabs

src/dhcpv4.c
```diff
@@ -549,263 +586,535 @@ dhcpv4_lease(struct interface *iface, enum dhcpv4_msg msg, const uint8_t *mac,
...
+		[IOV_ROUTER]		= { &reply_router, 0 },
+		[IOV_ROUTER_ADDR]	= { NULL, 0 },
+		[IOV_DNSSERVER]		= { &reply_dnsserver, 0 },
+		[IOV_DNSSERVER_ADDR]	= { NULL, 0 },
+		[IOV_HOSTNAME]		= { &reply_hostname, 0 },
+		[IOV_HOSTNAME_NAME]	= { NULL, 0 },
+		[IOV_MTU]		= { &reply_mtu, 0 },
+		[IOV_BROADCAST]		= { &reply_broadcast, 0 },
+		[IOV_NTP]		= { &reply_ntp, 0 },
+		[IOV_NTP_ADDR]		= { iface->dhcpv4_ntp, 0 },
```
systemcrash, line 798: could you fit in SIP servers (120) also? It's basically an `dhcpv4_option`.
```

   The DHCP option for this encoding has the following format:


     Code  Len   enc   DNS name of SIP server
   +-----+-----+-----+-----+-----+-----+-----+-----+--
   | 120 |  n  |  0  |  s1 |  s2 |  s3 |  s4 | s5  |  ...
   +-----+-----+-----+-----+-----+-----+-----+-----+--



   As an example, consider the case where the server wants to offer two
   outbound proxy servers, "example.com" and "example.net". These would
   be encoded as follows:


     +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
     |120|27 | 0 | 7 |'e'|'x'|'a'|'m'|'p'|'l'|'e'| 3 |'c'|'o'|'m'| 0 |
     +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
     +---+---+---+---+---+---+---+---+---+---+---+---+---+
     | 7 |'e'|'x'|'a'|'m'|'p'|'l'|'e'| 3 |'n'|'e'|'t'| 0 |
     +---+---+---+---+---+---+---+---+---+---+---+---+---+
```

systemcrash:
> @Noltari when you have time, please have a look :) @systemcrash should be easier to review now

Yeah - good. Can't fault anything.

src/dhcpv4.c
Alphix: Yeah, I can add it in later PRs. After this PR (and it's successor) have landed, it's quite easy to add options.

I already have `clientid` (DUID+IAID for DHCPv4, RFC4361) queued up...and option `108` (IPv6-only preferred, RFC8925, and yes, I know there's already a PR for it)...and custom options (for all the esoteric options that we might never support natively)

systemcrash: All good.

src/dhcpv4.c
Alphix: I think that's the same thing we've discussed earlier...`vim` and kernel coding style and all that...it lines up correctly if you have tabs = 8 characters :) I guess you want it to be three plain tabs?

Noltari (approved):

systemcrash:
LGTM

Noltari:
Merged, thanks @Alphix!

Alphix:
Yay!

src/dhcpv4.c
```diff
@@ -859,76 +1171,35 @@ void dhcpv4_handle_msg(void *addr, void *data, size_t len,
...
+		      dhcpv4_msg_to_string(reply_msg.data),
+		      dest_addr.sin_addr.s_addr == INADDR_BROADCAST ?
 		      "ff:ff:ff:ff:ff:ff": odhcpd_print_mac(req->chaddr, req->hlen),
-		      inet_ntoa(dest.sin_addr));
+		      inet_ntoa(dest_addr.sin_addr));
 	else
-		debug("Sent %s to %s - %s",
-		      dhcpv4_msg_to_string(msg),
-		      dest.sin_addr.s_addr == INADDR_BROADCAST ?
+		error("Sent %s to %s - %s",
```
testuser7, line 1196: @Alphix was this change from `debug` to `error` intentional?

systemcrash: Hmm. Trivial to fix.

systemcrash: See #375

PR #279, opened 2025-10-14
dhcpv6: message type check refactor

systemcrash:
A few minor optimisations which should provide a slight performance increase. The two commits type-check DHCPv6 messages in `relay_client_request` and `handle_client_request` earlier at function entry. The relay case performs fewer comparisons with switch vs an if which has many comparisons. Added comments for better clarity. 

This sets the ground for easier addition of future message types (I was investigating adding other RFC support).


systemcrash:
Yay - free compile test :D 

ping @Alphix

src/dhcpv6.c
```diff
@@ -318,6 +318,36 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+#ifndef DHCPV4_SUPPORT
+	/* if we omit DHCPV4 support, ignore this client message type */
+	case DHCPV6_MSG_DHCPV4_QUERY:	// 20
+#endif
+	case DHCPV6_MSG_DHCPV4_RESPONSE:// 21
+	default:
+		return;
+	}
+
 	syslog(LOG_DEBUG, "Got a DHCPv6-request on %s", iface->name);
```
Alphix: This debug message used to be unconditional...maybe move it above the switch to still get debug msgs for invalid or ignored messages?

src/dhcpv6.c
```diff
@@ -318,6 +318,36 @@ static void handle_client_request(void *addr, void *data, size_t len,
 	if (len < sizeof(*hdr))
 		return;
 
+	switch (hdr->msg_type) {
+	/* Valid message types for clients */
+	case DHCPV6_MSG_SOLICIT:	// 1
```
Alphix: Not sure it's valuable to add the value of the MSG type as a comment? It only risks being incorrect and it doesn't really add any value?

src/dhcpv6.c
```diff
@@ -318,6 +318,36 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+	case DHCPV6_MSG_REQUEST:	// 2
```
Alphix: In the second patch in PR #278, I added a `#define _fallthrough __attribute__((__fallthrough__))` to `odhcpd.h`....might want to consider using something like that for all these fallthroughs, and we should add warnings about implicit fallthroughs in the future....just something to consider

src/dhcpv6.c
systemcrash: Contra point: if a broken client were to spam us, do we want to log every packet we received or just the correct ones to be processed?

src/dhcpv6.c
systemcrash: I checked :) and it adds value at least for me when I refer to them again. It helps with at least a logical numerical ordering and how they're ordered in the RFC which makes the code generally easier to follow.

src/dhcpv6.c
systemcrash: so each case: should take an explicit _fallthrough; ?

src/dhcpv6.c
Alphix: Contra-contra point. It's just a debug level msg...so normally it won't go anywhere.

But yeah, I'm fine with either, I just noted that it's a change of behaviour. We also drop short packets silently already, with no debug message.

src/dhcpv6.c
Alphix: Yes, I think so. I mean, it is pretty obvious to the casual reader that a fallthrough is not a mistake here, but we can't enable implicit fallthrough warnings in GCC without explicit annotations. But it could also be added later (because I'm sure we have several places where it would need to be added), not something to delay this PR over (esp. since the _fallthrough definition is in a different PR)

src/dhcpv6.c
systemcrash: The message alludes to a request via `Got a DHCPv6-request` and the switch case is designed to filter away e.g. replies from servers that the RFC mandates we do not process. So I think we're in the clear here. Interesting point tho, logging incorrect stuff if we get any. Don't really want to get carried away with a 'simple' change :)

Alphix: Yes, and I think it might be a good idea in a very busy network

src/dhcpv6.c
```diff
@@ -891,11 +895,29 @@ static void relay_client_request(struct sockaddr_in6 *source,
...
 
-	if (h->msg_type == DHCPV6_MSG_RELAY_REPL ||
-	    h->msg_type == DHCPV6_MSG_RECONFIGURE ||
-	    h->msg_type == DHCPV6_MSG_REPLY ||
-	    h->msg_type == DHCPV6_MSG_ADVERTISE)
-		return; /* Invalid message types for client */
+	switch (h->msg_type) {
+	/* Valid message types from clients */
+	case DHCPV6_MSG_SOLICIT:	// 1
+	case DHCPV6_MSG_REQUEST:	// 2
```
Noltari: DHCPV6_MSG_REQUEST has a value of `3`, not `2`.
Honestly I think we should remove this comments since most IDEs will show the value when hovering over them.

Noltari (changes requested):

src/dhcpv6.c
systemcrash: Oh! well spotted. I use Sublime, and either it's not configured for me, or it doesn't do it. It's like a last step sanity check, which helps (at least me) when I'm reading the RFCs.

src/dhcpv6.c
Noltari: @Alphix isn't fallthrough only needed when there is code between one case and another?

src/dhcpv6.c
Noltari: > Oh! well spotted. I use Sublime, and either it's not configured for me, or it doesn't do it. It's like a last step sanity check, which helps (at least me) when I'm reading the RFCs.

I understand what you mean but this is prone to errors as the one I discovered, so please remove the comments.

src/dhcpv6.c
systemcrash: I think it's just about a warning since fallthrough is implied without a break. https://gcc.gnu.org/onlinedocs/gcc/Statement-Attributes.html

systemcrash: (some users forget the break and are surprised by ensuing behaviour)

src/dhcpv6.c
systemcrash: you and your fancy IDEs... :D

systemcrash: done

Noltari:
Merged, thanks @systemcrash!

src/dhcpv6.c
Alphix: Yeah, and it's something that can be added later, so don't let it hold up the PR

Alphix: Ah, it's already committed....all good

PR #282, opened 2025-10-20
odhcdp: use a more suitable clock

Alphix:
In my testing, CLOCK_MONOTONIC_COARSE uses about 1/3 of the CPU cycles.

And the resolution is completely irrelevant (only the seconds are used).


systemcrash:
Smart! (can do the same for odhcp6c)

systemcrash (approved):

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #284, opened 2025-10-21
dhcpv6: RFC4833 timezones

systemcrash:
ping @Alphix and @Noltari 

Can you see anything that should be improved here? This option differs from the others in that it's not an interface option, but instead just an internal `system_config` thing. I made it 'opt in', so the user must specifically enable its usage, so if there are any gremlins through its usage, they can be disabled.

====

This implements RFC4833 - supplying timezone information to clients that request them. Both forms are possible, when timezone is configured in the uci system settings (it saves both forms to the config).

e.g.
```
config system
	option zonename 'America/Puerto Rico'
	option timezone 'AST4'
```

There is also an odhcpd flag to toggle their use, set in uci dhcp.

```
config odhcpd 'odhcpd'
	option enable_tzdb '1'
```

Once enabled, the options, when requested, are sent: 

NEW_POSIX_TIMEZONE 41 // 'AST4'
NEW_TZDB_TIMEZONE 42 // 'America/Puerto_Rico'

Wireshark disassemble of options sent to client:

```
...
Time Zone Database
    Option: Time Zone Database (42)
    Length: 19
    TZ-database: America/Puerto_Rico
Time Zone
    Option: Time Zone (41)
    Length: 4
    Time-zone: AST4
```


src/config.c
```diff
@@ -33,6 +34,7 @@ struct vlist_tree leases = VLIST_TREE_INIT(leases, lease_cmp, lease_update, true
 AVL_TREE(interfaces, avl_strcmp, false, NULL);
 struct config config = {
 	.legacy = false,
+	.enable_tzdb = false,
```
Alphix: I think this should default to `true`. The client has to request the TZ to get it, and not many clients do. If we do come across issues, it would be simple for the reporter to change this, but if it defaults to `false`, I doubt it'll ever get any testing in real-world scenarios?

src/config.c
```diff
@@ -45,6 +47,12 @@ struct config config = {
 	.log_syslog = true,
 };
 
+struct system_config system_config = {
```
Alphix: Why not put this in `struct config`?

src/config.c
```diff
@@ -45,6 +47,12 @@ struct config config = {
...
+	.uci_cfgfile = "system",
```
Alphix, line 51: Why make this a variable when it cannot be changed?

(and on an unrelated note, my clever patch to make the cfg file a command line option slowly seems less and less useful...I might have to rework it...I have another patch where we need to read `/etc/config/network` as well for the global DUID)

src/config.c
```diff
@@ -226,13 +235,30 @@ static const struct blobmsg_policy odhcpd_attrs[ODHCPD_ATTR_MAX] = {
...
 
+enum {
+	SYSTEM_ATTR_TIMEZONE,
+	SYSTEM_ATTR_ZONENAME,
+	SYSTEM_ATTR_MAX
+};
+
+static const struct blobmsg_policy system_attrs[SYSTEM_ATTR_MAX] = {
+	[SYSTEM_ATTR_TIMEZONE] = { .name = "timezone", .type = BLOBMSG_TYPE_STRING },
+	[SYSTEM_ATTR_ZONENAME] = { .name = "zonename", .type = BLOBMSG_TYPE_STRING },
```
Alphix, line 256: "zonename" is pretty generic/vague isn't it? Maybe follow the naming from the RFC and call the options something like "POSIX_TIMEZONE" and "TZDB_TIMEZONE" (but lowercase)?

src/config.c
```diff
@@ -432,6 +458,49 @@ static void set_config(struct uci_section *s)
...
+
+}
+
+static char *sanitize_tz_string(const char *s)
+{
+	/* replace any spaces with '_' in tz strings. luci, where these strings
+	are normally set, (had a bug that) replaced underscores for spaces in the
+	names. */
+	if (!s)
+		return strdup(""); // safe to free later, also safe for length checks.
```
Alphix: This is a big footgun. `NULL` is safe for `free()` (and I'd consider code that checks a pointer before passing it to `free()` to be in bad taste).

One example of why it's not a good idea is a couple of lines below where `return strdup("");` can return NULL.

I'd strongly suggest going with NULL here, and separate variables for the length of the string, so 4 variables instead of 2. Then you don't need any `strlen()` calls later and the `iovec` interface can be simplified (see separate comment)

src/config.c
```diff
@@ -2287,6 +2368,7 @@ void odhcpd_reload(void)
 	}
 
 	uci_unload(uci, dhcp);
+	uci_unload(uci, system);
```
Alphix, line 2378: Nit: I'd move that up to right after the `if(!uci_load(...&system)`

src/dhcpv6.c
```diff
@@ -185,6 +185,8 @@ enum {
 	IOV_DHCPV4O6_SERVER,
 	IOV_DNR,
 	IOV_BOOTFILE_URL,
+	IOV_TZDB_TZ,
+	IOV_POSIX_TZ,
```
Alphix, line 188: You can simplify this if you use 4 `iovecs`...2 for the headers and 2 for the data, then you point the data ones straight at the TZ data you have stored in `struct config` and skip all copies, even to stack.

src/dhcpv6.c
```diff
@@ -587,6 +631,10 @@ static void handle_client_request(void *addr, void *data, size_t len,
 		[IOV_NTP_ADDR] = {ntp_ptr, (ntp_cnt) ? ntp_len : 0},
 		[IOV_SNTP] = {&dhcpv6_sntp, (sntp_cnt) ? sizeof(dhcpv6_sntp) : 0},
 		[IOV_SNTP_ADDR] = {sntp_addr_ptr, sntp_cnt * sizeof(*sntp_addr_ptr)},
+		[IOV_POSIX_TZ] = { client_wants_posix_tz ? posix_tz : NULL,
+							client_wants_posix_tz ? posix_tz_optlen : 0 },
+		[IOV_TZDB_TZ]  = { client_wants_tzdb_tz ? tzdb_tz : NULL,
+							client_wants_tzdb_tz ? tzdb_tz_optlen : 0 },
```
Alphix: Set the `iov_base` unconditionally to point to the `struct config` or `struct system_config` TZ strings, then use `iov_len` to turn this on or off.

src/dhcpv6.c
```diff
@@ -421,6 +423,40 @@ static void handle_client_request(void *addr, void *data, size_t len,
...
+	if (have_tzdb_tz > 0 && config.enable_tzdb) {
+		can_use_tzdb_tz = true;
+		size_t tzlen = strlen(system_config.zonename_tzdb_tz);
+		tzdb_tz = alloca(sizeof(*tzdb_tz) + tzlen);
+		tzdb_tz->type = htons(DHCPV6_OPT_NEW_TZDB_TIMEZONE);
+		tzdb_tz->len  = htons(tzlen);
+		memcpy(tzdb_tz->tz_name, system_config.zonename_tzdb_tz, tzlen);
+		tzdb_tz_optlen = sizeof(*tzdb_tz) + tzlen;
+	}
+
```
Alphix, line 459: If you use 4 `iovecs` and structure this more like e.g. the `NTP` option, you can cut down the number of variables.

src/dhcpv6.h
```diff
@@ -159,6 +163,27 @@ struct dhcpv6_cer_id {
...
+ * e.g. Europe/Zurich
+ * Variable-length opaque data blob.
+ */
+struct dhcpv6_new_tzdb_timezone {
+	uint16_t type;		/* = htons(42) */
+	uint16_t len;		/* length of data[] */
+	uint8_t  tz_name[];	/* opaque payload */
+} __attribute__((packed));
+
+
```
Alphix: And with the 4 `iovecs`, these structs won't be necessary...

Alphix (changes requested):

src/config.c
systemcrash: True - was exercising caution.

src/config.c
systemcrash: Yes, but that's what they've been set to for years on the uci side. So, those are what we use. :/

Alphix:
Good stuff!

Two more things:
- The `README.md` needs to be updated as well
- I guess [the init script](https://github.com/openwrt/openwrt/blob/f5fd7ef8886faadc654b627e5e4b4484c6ca31d7/package/network/services/odhcpd/files/odhcpd.init#L20) needs to be updated to also list `system` as a reload trigger

src/config.c
Alphix: In other words, make the strings look exactly like you want them to be in the body of the TZ DHCPv6 opt, then you can just include a direct pointer in the `iovec`

systemcrash: Ah thanks. Wasn't certain about the NULL thing. It's just that if it were set to NULL, an earlier patch iteration did a strlen on it and segfaulted.

I was just working on the extra len variables and came to the same idea myself. :O

src/config.c
systemcrash: Just felt like the right place to put it with the other unloads.

src/dhcpv6.c
systemcrash: Oh cool. Let me figure this out..

src/config.c
systemcrash: The thinking was that the code documents where the variables come from - not from the odhcpd config, but system config.

src/config.c
Alphix: Oh, I was reading too fast...ignore that then :)

src/dhcpv6.c
systemcrash: Only for the `*_tz_optlen` part, right?

Alphix: You can set the `iov_base` to e.g. `config->posix_tz` unconditionally and then set `iov_len` to either `0` or `config->posix_tz_len` depending on whether the client requests the option or not. Like how `IOV_NTP` and `IOV_NTP_ADDR` are defined. If `iov_len` is `0`, `iov_base` won't be read.

src/dhcpv6.c
systemcrash: ya, much simpler :)

systemcrash: with the caveat of managing the two IOVs separately...

src/dhcpv6.c
systemcrash: much cleaner :)

src/dhcpv6.h
systemcrash: poof gone

src/config.c
systemcrash: Yeah, jow set them as such several years ago. It is what it is...

src/config.c
systemcrash: consistency with how it's done for `config` 🤷

systemcrash:
> Good stuff!
> 
> Two more things:
> 
>     * The `README.md` needs to be updated as well
> 
>     * I guess [the init script](https://github.com/openwrt/openwrt/blob/f5fd7ef8886faadc654b627e5e4b4484c6ca31d7/package/network/services/odhcpd/files/odhcpd.init#L20) needs to be updated to also list `system` as a reload trigger

Potentially yes - but generally once a user sets a timezone, they rarely change it. Wondering if it's worth it, even though it's 'correct' :)

All done.

Alphix:
> > Good stuff!
> > Two more things:
> > ```
> > * The `README.md` needs to be updated as well
> > 
> > * I guess [the init script](https://github.com/openwrt/openwrt/blob/f5fd7ef8886faadc654b627e5e4b4484c6ca31d7/package/network/services/odhcpd/files/odhcpd.init#L20) needs to be updated to also list `system` as a reload trigger
> > ```
> 
> Potentially yes - but generally once a user sets a timezone, they rarely change it. Wondering if it's worth it, even though it's 'correct' :)

It was partially a reminder to self...I'm going to have to do something similar when I get around to updating the global DUID PR

src/config.c
```diff
@@ -432,6 +460,58 @@ static void set_config(struct uci_section *s)
...
+
+	}
+
+	if ((c = tb[SYSTEM_ATTR_ZONENAME])) {
+		free(sys_conf.zn_tzdb_tz);
+		sys_conf.zn_tzdb_tz = (uint8_t *)sanitize_tz_string(blobmsg_get_string(c));
+		if (sys_conf.zn_tzdb_tz && *sys_conf.zn_tzdb_tz)
+			sys_conf.zn_tzdb_tz_len = strlen((char *)sys_conf.zn_tzdb_tz);
+		else
+			sys_conf.zn_tzdb_tz_len = 0;
```
Alphix: The exact same thing is done for both attributes here....if you're up for some code golf, you could pass a pointer to the string ptr and the size_t to `sanitize_tz_string()` and put all the logic inside that function...so you end up with:

```C
if ((c = tb[SYSTEM_ATTR_TIMEZONE]))
    sanitize_tz_string(blobmsg_get_string(c), &config.tz_posix_tz, &config.tz_posix_tz_len);

if ((c = tb[SYSTEM_ATTR_ZONENAME]))
    sanitize_tz_string(blobmsg_get_string(c), &config.zn_tzdb_tz, &config.zn_tzdb_tz_len);
```

src/odhcpd.h
```diff
@@ -193,6 +195,14 @@ struct config {
...
 };
 
+struct sys_conf {
+	uint8_t *tz_posix_tz;
+	size_t tz_posix_tz_len;
+	char *uci_cfgfile;
+	uint8_t *zn_tzdb_tz;
+	size_t zn_tzdb_tz_len;
+};
+
```
Alphix, line 210: Still think this should go into `struct config`...and the variable names are a bit overcomplicated? `tz_posix_tz` as in `timezone_posix_timezone`? How about dropping the prefixes and going with `posix_tz`, `posix_tz_len`, `tzdb_tz`, `tzdb_tz_len`?

src/config.c
```diff
@@ -226,13 +237,30 @@ static const struct blobmsg_policy odhcpd_attrs[ODHCPD_ATTR_MAX] = {
 	[ODHCPD_ATTR_LOGLEVEL] = { .name = "loglevel", .type = BLOBMSG_TYPE_INT32 },
 	[ODHCPD_ATTR_HOSTSFILE] = { .name = "hostsfile", .type = BLOBMSG_TYPE_STRING },
 	[ODHCPD_ATTR_PIOFOLDER] = { .name = "piofolder", .type = BLOBMSG_TYPE_STRING },
+	[ODHCPD_ATTR_ENABLE_TZDB] = { .name = "enable_tzdb", .type = BLOBMSG_TYPE_BOOL },
```
Alphix: Nit: how about "enable_tz"?

Alphix:
The second version looks so much cleaner (IMHO)

src/odhcpd.h
systemcrash: I was half way to doing that, but I realised I wanted to link those property names to the rather confusingly named uci parameter names, so it's easier to trace where they come from.

Alphix: Counter-point: you perpetuate the confusing names throughout the code then rather than have one single place where you map the (confusing) `uci` <-> `odhcpd` (not confusing, I hope) variable names

src/config.c
systemcrash: done

src/odhcpd.h
systemcrash: done.

src/config.c
systemcrash: Oh man, stuff starts to get hairy with pointers to pointers :D 

All done now tho. Thanks for the idea!

systemcrash:
OK - sweet. Thanks for the pointers. Low fat. Way better than v0.

If there are no further comments, LGTM.

Alphix:
Purty. Still don't like `struct sys_conf` but not my call. LGTM ❤️

systemcrash:
Ready when you are @Noltari. I've tested the latest iteration on my production system and it performs as expected.

Noltari:
Merged, thanks @systemcrash!

Alphix:
@Noltari out of curiosity...why are some PRs closed after merge (like this one) and some marked as merged?

Noltari:
> @Noltari out of curiosity...why are some PRs closed after merge (like this one) and some marked as merged?

Normally Github will automatically close PRs if the same commits are added to the branch referenced by the PR.
Since this repository is a mirror of https://git.openwrt.org/?p=project/odhcpd.git, whenever I push there, @openwrt-bot will automatically push the same changes to Github. That's why it's always flagged as merged by @openwrt-bot.
However, in this specific case I forgot to push the PR branch first (my bad, sorry for that) and Github couldn't match the commits with the master branch ones.

PR #285, opened 2025-10-21
odhcpd: improve odhcpd_urandom()

Alphix:
First, note that not a single caller checks the return value - which is quite reasonable. What are they supposed to do with a failure?

Second, none of the callers do anything that's *really* security-sensitive, the closest we have is the force reconf nonce, and that is blorted out over the network, so it's really a best-effort kind of thing.

Third, odhcpd_urandom() currently doesn't check if it e.g. got interrupted by a signal.

So, simplify and modernize this a bit by using getrandom(), which allows us to skip one fd, and which avoids syscalls by using the vDSO approach instead. Also, check for things like signal interrupts (don't really happen on calls for entropy < 256 bytes, but still). And make a reasonable effort, but not much more.

### DIFF
--- a/src/odhcpd.c
+++ b/src/odhcpd.c
@@ -39,12 +39,12 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <sys/syscall.h>
+#include <sys/random.h>
 
 #include <libubox/uloop.h>
 #include "odhcpd.h"
 
 static int ioctl_sock = -1;
-static int urandom_fd = -1;
 
 void __iflog(int lvl, const char *fmt, ...)
 {
@@ -148,11 +148,7 @@ int main(int argc, char **argv)
 	}
 
 	ioctl_sock = socket(AF_INET, SOCK_DGRAM | SOCK_CLOEXEC, 0);
-
 	if (ioctl_sock < 0)
-		return 4;
-
-	if ((urandom_fd = open("/dev/urandom", O_RDONLY | O_CLOEXEC)) < 0)
 		return 4;
 
 	signal(SIGUSR1, SIG_IGN);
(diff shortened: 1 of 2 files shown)


Alphix:
@systemcrash and @Noltari

systemcrash:
> by using getrandom(), which allows us to skip one fd

Aren't we just kicking the can down the road here? getrandom() [reads from /dev/urandom](https://www.man7.org/linux/man-pages/man2/getrandom.2.html) ya?

Just an idea - given that odhcpd and odhcp6c can be thought of as client/server team here, and that new patchset for odhcp6c provides support for RKAP, which also uses HMAC-MD5, I guess we're no worse off. 

If we're not striving for high quality randomness, DHCP is the last place we should worry about if the NSA is attempting to gain access to our system :D


Bottom line: OK by me.

Alphix:
> > by using getrandom(), which allows us to skip one fd
> 
> Aren't we just kicking the can down the road here? getrandom() [reads from /dev/urandom](https://www.man7.org/linux/man-pages/man2/getrandom.2.html) ya?

Yes and no...it uses the same _source_ of randomness (the `urandom` pool), but it gets the random bytes from the pool using the [vDSO](https://man7.org/linux/man-pages/man7/vdso.7.html) mechanism, which avoids both syscalls and file descriptors (meaning it uses less CPU cycles and is also oblivious to things like being `chroot()`:ed, etc). It also handles `EINTR` better (reads < 256 bytes are guaranteed to not be interrupted...which is all the calls we currently do).

> If we're not striving for high quality randomness, DHCP is the last place we should worry about if the NSA is attempting to gain access to our system :D

Yeah, the quality of the random data is not improved at all by this patch :)

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #287, opened 2025-10-22
dhcpv4: bump problem scenario up to warn

### LINKED ISSUE #228: Inconsistent syslog priorities
I recently had a device which exhausted my DHCPv4 pool. It took me a while to debug this, since there was nothing from odhcpd in the logs. The `loglevel` for odhcpd is set to 4 (WARN).

Of course odhcpd logs a lack of addresses - but a NOTICE level, so I didn't get it with my settings.
https://github.com/openwrt/odhcpd/blob/5585b969c18f84bd3bb7d02a0034c19ebae79e79/src/dhcpv4.c#L1098

I think this is somewhat inconsistent with other messages at this level. In particular the following seem pretty normal and not "[requiring special handling](https://pubs.opengroup.org/onlinepubs/7908799/xsh/syslog.h.html)".
https://github.com/openwrt/odhcpd/blob/5585b969c18f84bd3bb7d02a0034c19ebae79e79/src/router.c#L889
https://github.com/openwrt/odhcpd/blob/5585b969c18f84bd3bb7d02a0034c19ebae79e79/src/router.c#L957
https://github.com/openwrt/odhcpd/blob/5585b969c18f84bd3bb7d02a0034c19ebae79e79/src/router.c#L987
https://github.com/openwrt/odhcpd/blob/5585b969c18f84bd3bb7d02a0034c19ebae79e79/src/router.c#L1026

I believe "address space is full" is a fairly severe error condition (because address assignment fails and devices can't join the network) and should probably be logged at WARN or even ERROR. However, this messages was changed from WARN to NOTICE in d60f0a6284e8f5fac733a1b269ea614ccdf47fcd in response to #150.

RA and RS status messages seem less important to me and I suggest logging them at INFO. Note that this was the case for "Sending a RA" until it was promoted to NOTICE in 207f8e02e9011e42375837e8875885e01013a171 (probably for consistence with other messages).
#191 even suggests DEBUG for "Sending a RA".

Implementing both changes I suggest would leave the NOTICE level pretty much unused and therefore useless. I'm fine with doing any one of them. I just want to be able to configure odhcpd to be silent unless something is broken - which currently is not quite possible. I think I set the loglevel to WARN because the RA messages I got every few minutes when I tried NOTICE.

@dedeckeh Since you made those changes to the log priorities back in 2019, maybe you can shed some light on the rationale and the best way forward?

I'll be happy to make a PR for the changes, if that's any help to you.

systemcrash:
This way a potential failure scenario is visible with odhcpds default log level of warn.

Closes #228

@Noltari low effort fix :)

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -423,7 +423,7 @@ static bool dhcpv4_assign(struct interface *iface, struct dhcp_assignment *a,
 		}
 	}
 
-	notice("Can't assign any IP address -> address space is full");
+	warn("Can't assign any IP address -> address space is full");
 
 	return false;
 }


Noltari:
Merged, thanks @systemcrash!

PR #288, opened 2025-10-22
odhcpd: fix syslog

Noltari:
Remove unneeded log_level_cmdline from config, which prevents properly setting the loglevel from the uci config.

This fixes the log issue reported at https://github.com/openwrt/odhcpd/pull/273#issuecomment-3431453342

CC @Alphix @systemcrash


Alphix:
I don't think it should be removed altogether, just that the check should be corrected. It's kind of customary that command line options override cfg options (and confusing when they don't).

Noltari:
> I don't think it should be removed altogether, just that the check should be corrected. It's kind of customary that command line options override cfg options (and confusing when they don't).

You're totally right.
Excuse me, but I was in a hurry and I didn't check it properly :D

PR #289, opened 2025-10-25
dhcpv6: send timezone options only when client actually requests them

systemcrash:
follow-up to 7956f4271b4efc23dbb9b88e82f1720eadb655a2

ping @Alphix 

I see the error now 🤦 Let's use a boolean to fix it.


Noltari (approved):

Noltari:
Merged, thanks @systemcrash!

PR #290, opened 2025-10-25
dhcpv6-ia: simplify/fix IID calculations

Alphix:
This is a first stab at simplifying (and fixing) the IID calculations introduced with commit 7000557cd8f6396b82cb35e0cdf9b72b18288fbc.

See the discussion in:
https://github.com/openwrt/openwrt/issues/20523

Note that a "cold" `jrand48()` doesn't have much of an avalance effect in case of two very similar DUIDs, but it shouldn't matter because we're not looking for any kind of cryptographic randomness...just a predictable PRNG where the same DUID will (usually) get the same IPv6 addr (if available).

Also note that I haven't removed the "shift the IID" code, because I'm not 100% sure why it's there...

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -53,25 +53,6 @@ static struct uloop_timeout valid_until_timeout = {.cb = valid_until_cb};
 static uint32_t serial = 0;
 static uint8_t statemd5[16];
 
-static inline uint32_t mask_low_for_hostid(uint8_t dhcpv6_hostid_len)
-{
-	if (dhcpv6_hostid_len >= 32)
-		return UINT32_MAX;
-
-	return (1 << dhcpv6_hostid_len) - 1;
-}
-
-static inline uint32_t mask_high_for_hostid(uint8_t dhcpv6_hostid_len)
-{
-	if (dhcpv6_hostid_len >= 64)
-		return UINT32_MAX;
-
-	if (dhcpv6_hostid_len >= 32)
-		return (1 << (dhcpv6_hostid_len - 32)) - 1;
-
-	return 0;
-}
-
 int dhcpv6_ia_init(void)
 {
 	uloop_timeout_set(&valid_until_timeout, 1000);
@@ -236,13 +217,33 @@ static void dhcpv6_ia_free_assignment(struct dhcp_assignment *a)
 	free(a->managed);
 }
 
+static void in6_copy_iid(struct in6_addr *dest, uint64_t iid, unsigned n)
+{
+	uint64_t iid_be = htobe64(iid);
+	uint8_t *iid_bytes = (uint8_t *)&iid_be;
+	unsigned bytes = n / 8;
+	unsigned bits = n % 8;
+
+	if (n == 0 || n > 64)
+		return;
+
+	memcpy(&dest->s6_addr[16 - bytes], &iid_bytes[8 - bytes], bytes);
+
+	if (bits > 0) {
+		unsigned dest_idx = 16 - bytes - 1;
+		unsigned src_idx = 8 - bytes - 1;
+		uint8_t mask = (1 << bits) - 1;
+		dest->s6_addr[dest_idx] = (dest->s6_addr[dest_idx] & ~mask) |
+					  (iid_bytes[src_idx] & mask);
+	}
+}
+
 void dhcpv6_ia_enum_addrs(struct interface *iface, struct dhcp_assignment *c,
 			  time_t now, dhcpv6_binding_cb_handler_t func, void *arg)
 {
 	struct odhcpd_ipaddr *addrs = (c->managed) ? c->managed : iface->addr6;
 	size_t addrlen = (c->managed) ? (size_t)c->managed_size : iface->addr6_len;
 	size_t m = get_preferred_addr(addrs, addrlen);
-	uint32_t mask_high, mask_low;
 
 	for (size_t i = 0; i < addrlen; ++i) {
 		struct in6_addr addr;
@@ -269,16 +270,7 @@ void dhcpv6_ia_enum_addrs(struct interface *iface, struct dhcp_assignment *c,
 			if (!ADDR_ENTRY_VALID_IA_ADDR(iface, i, m, addrs))
 				continue;
 
-			mask_low = mask_low_for_hostid(iface->dhcpv6_hostid_len);
-			mask_high = mask_high_for_hostid(iface->dhcpv6_hostid_len);
-			addr.s6_addr32[2] = htonl(
-				((c->assigned_host_id >> 32) & mask_high)
-				& (~mask_high & ntohl(addrs[i].addr.in6.s6_addr32[2]))
-			);
-			addr.s6_addr32[3] = htonl(
-				((c->assigned_host_id & UINT32_MAX) & mask_low)
-				& (~mask_low & ntohl(addrs[i].addr.in6.s6_addr32[3]))
-			);
+			in6_copy_iid(&addr, c->assigned_host_id, iface->dhcpv6_hostid_len);
 		} else {
 			if (!valid_prefix_length(c, addrs[i].prefix))
 				continue;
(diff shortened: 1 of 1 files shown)


systemcrash:
Maybe this? 

https://www.rfc-editor.org/rfc/rfc4291.html#section-2.5.1

Need to think more deeply about this before I could say definitively. Observing address handout with and without that flag should be revealing. Stuff like this is why I like code comments that call out RFC portions :)

Alphix:
> Maybe this?
> 
> https://www.rfc-editor.org/rfc/rfc4291.html#section-2.5.1
> 
> Need to think more deeply about this before I could say definitively. Observing address handout with and without that flag should be revealing. Stuff like this is why I like code comments that call out RFC portions :)

Which flag?

systemcrash:
This bit:
   For all unicast addresses, except those that start with the binary
   value 000, Interface IDs are required to be 64 bits long and to be
   constructed in Modified EUI-64 format.

So that address behaviour change we saw might actually be the correct behaviour?

Alphix:
> This bit: For all unicast addresses, except those that start with the binary value 000, Interface IDs are required to be 64 bits long and to be constructed in Modified EUI-64 format.
> 
> So that address behaviour change we saw might actually be the correct behaviour?

No, it's very weird behaviour.

```C
/* If prefix is < 64 bits, we need to shift the interface ID to the left */
if (iface->addr6_len < 64)
        try <<= (64 - iface->addr6_len);
```

`iface->addr6_len` is the number of IPv6 addresses assigned to that interface, so it'll generate different values depending on whether the interface has 1, 2 or x addresses. Typically it'll shift `try` by 62-63 bits, discarding almost all of it...

And the modified EUI-64 IID (the RFC you referred to) is for SLAAC and that kind of stuff (it's also quite old, from a time when that kind of IID was expected to become the _only_ kind of IID; it's since been revised...e.g. RFC7136, §5 explicitly softened the language...EUI-64 is not what's used here where we're dealing with random address selection within a pool of DHCP addresses)

systemcrash:
it’s just implied behaviour in later dhcpv6 RFC which all mandate 64 bits.

Aaah - of course. It’s the amount of addresses. Length in this case was ambiguous enough to be interpreted as the length of the address in bits. Time for a refactor 🤓

That bit is most certainly wrong then. I suppose bit length is what the author was aiming for.

Alphix:
> Aaah - of course. It’s the amount of addresses. Length in this case was ambiguous enough to be interpreted as the length of the address in bits. Time for a refactor 🤓

Yes, I've already had the same thought...we should standardize on `*_cnt` variable names for things that are a count of something...and `_len` for things that are actual lengths (of e.g. a string).

> That bit is most certainly wrong then. I suppose bit length is what the author was aiming for.

I guess so as well...but even if that's the correct assumption, the code is confusing...a prefix with < 64 bit means we meddle around with the IID? It's like the IID was defined as 128 bits and supposed to move partially into the first 64 bits of the complete IPv6 addr?....dunno

systemcrash:
The code comments says what it should be doing - but until we can find precedent in an RFC or validate the behaviour, I think it can be excluded for now.

systemcrash:
ping @xvzf

src/dhcpv6-ia.c
```diff
@@ -878,31 +874,24 @@ static bool assign_na(struct interface *iface, struct dhcp_assignment *a)
...
+	for (size_t i = 0; i < 100; i++, try++) {
+		if (try > pool_end)
+			try = pool_start;
 
+#if 0
+		// FIXME: que?
 		/* If prefix is < 64 bits, we need to shift the interface ID to the left */
 		if (iface->addr6_len < 64)
 			try <<= (64 - iface->addr6_len);
+#endif
```
Noltari: I also think we can drop this...

src/dhcpv6-ia.c
```diff
@@ -236,13 +217,33 @@ static void dhcpv6_ia_free_assignment(struct dhcp_assignment *a)
 	free(a->managed);
 }
 
+static void in6_copy_iid(struct in6_addr *dest, uint64_t iid, unsigned n)
+{
+	if (n <= 0 || n > 64) return;
```
Noltari: `n` is unsigned, so it can't be < 0:
```
if (n == 0 || n > 64)
	return;
```
(also move the return to the next line please)

Noltari (changes requested):

src/dhcpv6-ia.c
Alphix: Yes, and I moved the vars to the top as well

src/dhcpv6-ia.c
Alphix: Ok, that makes three of us...

AndyChiang888:
LGTM, The issue of commit 7000557cd8f6396b82cb35e0cdf9b72b18288fbc has been solved, my test is normal.
Thanks @Alphix

Alphix:
> LGTM, The issue of commit [7000557](https://github.com/openwrt/odhcpd/commit/7000557cd8f6396b82cb35e0cdf9b72b18288fbc) has been solved, my test is normal. Thanks @Alphix

@AndyChiang888 Thanks for testing this :)

Noltari:
@Alphix should we mark this as ready for review and merge it?
(It's still a draft)

Alphix:
Hmm...unfortunately I suspect there's more bugs lurking here...not introduced by this PR...but more things to fix...I'm debugging right now, but I have a strong feeling that there's more places in `src/dhcpv6-ia.c` that expect the prefix to not be longer than `/64`...

Alphix:
OTOH, merging this should already fix some bugs...

src/dhcpv6-ia.c
```diff
@@ -878,31 +874,17 @@ static bool assign_na(struct interface *iface, struct dhcp_assignment *a)
...
-		}
-
-		/* If prefix is < 64 bits, we need to shift the interface ID to the left */
-		if (iface->addr6_len < 64)
-			try <<= (64 - iface->addr6_len);
+	/* Pick a starting point, using the last bytes of the DUID as seed... */
+	memcpy(xsubi,
+	       a->clid_data + (a->clid_len > sizeof(xsubi) ? a->clid_len - sizeof(xsubi) : 0),
+	       min(a->clid_len, sizeof(xsubi)));
+	try = ((uint64_t)jrand48(xsubi) << 32) | (jrand48(xsubi) & UINT32_MAX);
```
systemcrash, line 881: Is it an idea to throw in some md5 randomness where jrand48 is, since we're already using md5?
```c
	md5_ctx_t md5;
	md5_begin(&md5);
	try = (uint64_t)md5_hash(xsubi, sizeof(xsubi), &md5); // some truncation
	...
```

systemcrash: (untested)

Alphix: If the lack of an avalanche effect turns out to be a problem, there's a lot of things we could do...like MurmurHash3's finaliser (which is designed to avalanche ~50% of the bits) plus a `jrand48()` burn-in:

```C
void init_xsubi(unsigned short xsubi[3], uint64_t seed) {
    // Hash first
    seed ^= seed >> 33;
    seed *= 0xff51afd7ed558ccdULL;
    seed ^= seed >> 33;
    seed *= 0xc4ceb9fe1a85ec53ULL;
    seed ^= seed >> 33;
    
    xsubi[0] = seed & 0xFFFF;
    xsubi[1] = (seed >> 16) & 0xFFFF;
    xsubi[2] = (seed >> 32) & 0xFFFF;
    
    // Warm up
    for (int i = 0; i < 10; i++) {
        jrand48(xsubi);
    }
}
```

But I think we can cross that bridge once we get to it...i.e. if this turns out to be a problem...

Alphix:
@Noltari and @systemcrash Ok, I've found some more issues, but I want to deal with them in a follow-up PR to not complicate this one any further. I think this can go in now.

(pushed a new version with "Closes:" tags)

GoetzGoerisch (approved):
LGTM 👍🏻
run-tested on: x86/64

* fixes #293

Noltari (approved):

PR #291, opened 2025-10-25
odhcpd: change "-c" cmd line arg to take a dir

Alphix:
After the TZ support was added, odhcpd now reads two cfg files potentially. Instead of adding a separate switch for the system UCI file, I've changed the "-c" argument to take a directory. This is more future-proof and also serves as a preparation for the global DUID PR (which will necessitate reading from the UCI "network" file as well).

While I was at it, I also made some tiny improvements to the odhcpd_reload() function to avoid an unnecessary allocation.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -41,14 +41,13 @@ struct config config = {
 	.dhcp_hostsfile = NULL,
 	.ra_piofolder = NULL,
 	.ra_piofolder_fd = -1,
-	.uci_cfgfile = "dhcp",
+	.uci_cfgdir = NULL,
 	.log_level = LOG_WARNING,
 	.log_level_cmdline = false,
 	.log_syslog = true,
 };
 
 struct sys_conf sys_conf = {
-	.uci_cfgfile = "system",
 	.posix_tz = NULL, // "timezone"
 	.posix_tz_len = 0,
 	.tzdb_tz = NULL, // "zonename"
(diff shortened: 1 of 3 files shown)


Alphix:
Ping @systemcrash and @Noltari

systemcrash:
Are you certain this works as intended? Because it's the uci library which defines the folder. So it seems any `%s` folder path you inject becomes `/etc/config</newpathfrom-cparam/configfile>`:

https://github.com/openwrt/uci/blob/2e46a7405f26ec131918c821c30ac0b26fdb22f0/uci.h#L39

Alphix:
> Are you certain this works as intended? Because it's the uci library which defines the folder. So it seems any `%s` folder path you inject becomes `/etc/config</newpathfrom-cparam/configfile>`:
> 
> https://github.com/openwrt/uci/blob/2e46a7405f26ec131918c821c30ac0b26fdb22f0/uci.h#L39

Yeah, the file backend (which is the only one we care about for devel purposes, for other purposes, the "-c" option won't be used) takes relative and absolute paths:
https://github.com/openwrt/uci/blob/2e46a7405f26ec131918c821c30ac0b26fdb22f0/file.c#L919-L943

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #292, opened 2025-10-25
odhcpd: struct reorganization

Alphix:
As promised, I had another big reorg PR in waiting :)

The gist is that `struct lease` is renamed `struct lease_cfg` (since it isn't a *lease*, i.e. the actual "live" contract between server and client, it's the configuration from UCI `host` sections which can be used to provide the cfg for a static lease).

Then `struct dhcp_assignment` is split into two separate structs for dhcpv4 and dhcpv6 leases.

### DIFF
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -251,26 +251,9 @@ struct dhcpv4_lease {
 	struct odhcpd_ref_ip *fr_ip;		// FR message old serverid/IP
 };
 
-/* This corresponds to a UCI host section, i.e. a static lease cfg */
-struct lease_cfg {
-	struct vlist_node node;
-	struct list_head assignments;
-	struct dhcpv4_lease *dhcpv4_lease;
-	uint32_t ipaddr;
-	uint64_t hostid;
-	size_t mac_count;
-	struct ether_addr *macs;
-	size_t duid_count;
-	struct duid *duids;
-	uint32_t leasetime;		// duration of granted leases, UINT32_MAX = inf
-	char *hostname;
-};
-
-struct dhcp_assignment {
+struct dhcpv6_lease {
 	struct list_head head;
 	struct list_head lease_cfg_list;
-
-	void (*dhcp_free_cb)(struct dhcp_assignment *a);
 
 	struct interface *iface;
 	struct lease_cfg *lease_cfg;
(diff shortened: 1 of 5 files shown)


Alphix:
@Noltari and @systemcrash this should be ready for review now

systemcrash:
Not for want of derailing your effort, but while host_cfg seems fine in the context of how it's named in uci, it's easy (for me) to forget when you're buried in odhcpd code. Would you be against something like

`host_lease` or even `lease_cfg` instead? This disambiguates host_config (odhcpd host?) a bit more since odhcpd interacts with 'hosts' beyond loading uci host config entries. Maybe I just need to get used to this more, or need more code comments :)

Otherwise, this is fine and functional and more readable than `l`.

Alphix:
> Not for want of derailing your effort, but while host_cfg seems fine in the context of how it's named in uci, it's easy (for me) to forget when you're buried in odhcpd code.

Yeah, it's easy for me as well...I've noticed that when switching between branches.

> Would you be against something like `host_lease` or even `lease_cfg` instead? This disambiguates host_config (odhcpd host?) a bit more since odhcpd interacts with 'hosts' beyond loading uci host config entries. Maybe I just need to get used to this more, or need more code comments :)
> 
> Otherwise, this is fine and functional and more readable than `l`.

I don't like `host_lease` since it's ambiguous (could mean an actual active lease, whether dynamic or static, while this struct is actually storing the cfg related to a "future" static lease...).

But `lease_cfg` would be fine...I think...but I'd need to actually try the name change...I've played with many many different variations, and some I discarded simply because the function names ended up looking really weird.

Alphix:
@systemcrash I've pushed an amended version where it's `struct lease_cfg` instead of `struct host_cfg`

src/odhcpd.h
```diff
@@ -468,32 +472,17 @@ struct interface {
...
-}
+enum {
+	HOST_ATTR_IP,
+	HOST_ATTR_MAC,
+	HOST_ATTR_DUID,
+	HOST_ATTR_HOSTID,
+	HOST_ATTR_LEASETIME,
+	HOST_ATTR_NAME,
+	HOST_ATTR_MAX
+};
```
Noltari, line 485: Doesn't it make more sense to rename all these to `LEASE_CFG_ATTR_*` or keep them as they were (`LEASE_ATTR_*`)?

Noltari (changes requested):

src/odhcpd.h
Alphix: Good catch, I missed that when I changed to struct lease_cfg... I'll rename them as well

Alphix: Fixed in the new version I just pushed

Alphix:
@Noltari @systemcrash I think I've addressed all issues now...

Noltari (approved):

PR #294, opened 2025-10-26
odhcpd: document undocumented options, remove deprecated options

Alphix:
I've gone through `src/config.c` and reviewed the options defined there, comparing them to what's currently in `README.md`. There are a couple of options that are currently undocumented.

Going through them, I found a number of options that are no longer necessary, so this PR removes a bunch of options for a nice cleanup (especially happy about the `CER`/`pd_managed` removal, which makes the code in `src/dhcpv6-ia.c` much easier to follow).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,10 +58,6 @@ target_link_libraries(${PROJECT_NAME} PRIVATE ${libjson})
 
 
 # Optional Features
-if (${EXT_CER_ID})
-	target_compile_definitions(${PROJECT_NAME} PRIVATE EXT_CER_ID=${EXT_CER_ID})
-endif(${EXT_CER_ID})
-
 if(${UBUS})
 	target_compile_definitions(${PROJECT_NAME} PRIVATE WITH_UBUS)
 	target_sources(${PROJECT_NAME} PRIVATE src/ubus.c)
--- a/README.md
+++ b/README.md
@@ -121,7 +121,6 @@ and may also receive information from ubus
 | ntp			|list	|`<local address>`| NTP servers to announce accepts IPv4 and IPv6 |
 | ra_management		|string	| -	| TBD |
 | upstream		|list	| -	| TBD |
-| filter_class		|string | -	| TBD |
 | pd_manager		|string | -	| TBD |
 | pd_cer		|string | -	| TBD |
 | ra_advrouter		|bool   | -	| TBD |
--- a/README.md
+++ b/README.md
@@ -119,6 +119,14 @@ and may also receive information from ubus
 | ndp_from_link_local	|bool	| 1	| Use link-local source addresses for NDP operations (RFC 4861, §4.2 compliance) and macOS compatibility |
 | prefix_filter		|string	|`::/0`	| Only advertise on-link prefixes within the provided IPv6 prefix; others are filtered out. [IPv6 prefix] |
 | ntp			|list	|`<local address>`| NTP servers to announce accepts IPv4 and IPv6 |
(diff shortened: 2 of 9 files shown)


Alphix:
@Noltari and @systemcrash please have a look when you have time, the individual commit messages explain each change in more detail

systemcrash (commented):
odcpd?
```
The "filter_class" option was first introduced as part of the effort to support
the Homenet Control Protocol (HNCP, RFC7368, RFC7788). At first it was
hardcoded to filter out the user class "HOMENET", and later it was made
configurable, but there appears to be no other use-cases than Homenet. There's
also legacy comments in the odhcpd code pointing at filtering out homenet user
classes (the comments are misleading since "filter_class" defaults to being
unset).
```

systemcrash (commented):
The which option?
```
LuCI has already been updated so that the "dhcpv4" option is exposed in the UI
and the option also gets set when the user clicks the "Enabled DHCP" option in
the LuCI UI. Furthermore, there are instructions in the UI to enable the DHCPv4
server on a per-interface basis.
```

... clicks to enable the "DHCPv4" odhcpd option?

systemcrash:
With homenet gone - now the obvious candidate is RFC8475 ( mentioned previously https://github.com/openwrt/odhcpd/pull/245#issuecomment-3313331423 ). :D

Alphix:
> odcpd?

opdhcp? ochdpd? odpchp?

Alphix:
> The which option?

The "dhcpv4" option? [This one...](https://github.com/openwrt/odhcpd/blob/d44af6dd8f4e1dd0d858ae19419057ab4f319310/README.md?plain=1#L84)

> ```
> LuCI has already been updated so that the "dhcpv4" option is exposed in the UI
> and the option also gets set when the user clicks the "Enabled DHCP" option in
> the LuCI UI. Furthermore, there are instructions in the UI to enable the DHCPv4
> server on a per-interface basis.
> ```
> 
> ... clicks to enable the "DHCPv4" odhcpd option?

I meant the "Set up DHCP Server" button...[defined here](https://github.com/openwrt/luci/blob/c935260b3026d805dd414bd266884cfe0600d03a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L679-L702)

Alphix:
@systemcrash I think I've addressed your comments :D

Noltari (approved):

systemcrash:
@Alphix 
> odcpd?
> 
> ```
> The "filter_class" option was first introduced as part of the effort to support
> the Homenet Control Protocol (HNCP, RFC7368, RFC7788). At first it was
> hardcoded to filter out the user class "HOMENET", and later it was made
> configurable, but there appears to be no other use-cases than Homenet. There's
> also legacy comments in the odhcpd code pointing at filtering out homenet user
> classes (the comments are misleading since "filter_class" defaults to being
> unset).
> ```

I also snuck in the "defaults to being unset" rather than the more logically taxing "defaults to not being set". But this is a nitpick :)

Edit: you could just elide the word being: "defaults to unset". Alas we're nitpicking at something which is gone.

Alphix:
> I also snuck in the "defaults to being unset" rather than the more logically taxing "defaults to not being set". But this is a nitpick :)
> 
> Edit: you could just elide the word being: "defaults to unset". Alas we're nitpicking at something which is gone.

"is not set by default" is the new wording :)

systemcrash:
Ok. LGTM.

Noltari:
Merged, thanks @Alphix!

PR #295, opened 2025-10-26
odhcpd: more fixes for IID calculations

Alphix:
This is the followup to pull request #290.

First, note that in dhcpv6_ia_enum_addrs(), the prefix + IID are currently combined based on iface->dhcpv6_hostid_len, which is wrong. The latter only defines the number of bits (typically 12) which should be used to generate/constrain IIDs, but not the actual IID length.

The actual length of the IID is defined by the length of the prefix (i.e.  128 - prefix length), the reason this hasn't been noticed is probably that the higher bits of the IID area in the prefix (the struct odhcpd_ipdaddr which corresponds to the DHCP server's own IPv6 address on the interface) are often anyway zero (i.e. people often use an address like fd00:1234:5678:abcd::1/64 for the server if they have received a prefix like fd00:1234:5678::/48 from the upstream router).

Second, build_ia() unconditionally writes 64 bits of struct dhcp_assignment->assigned_host_id (IID) to the struct in6_addr to use for a given client (i.e. assumes the prefix to be at least /64 or shorter).

This is a first stab at fixing it. I've tried it, and successfully obtained a lease using a /96 prefix.

There are still some murky corners here though, for example, if "dhcpv6_hostidlength" is set to 64, there are no real checks that the prefix length is compatible with that (so we might end up generating an IID of 64 bits, making sure that it is unique and not taken, and then truncate it silently to a 32 bit IID in case the prefix is /96). But that's a topic for a different day (and can only be triggered by non-standard configurations).

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -275,6 +275,17 @@ static void in6_copy_iid(struct in6_addr *dest, uint64_t iid, unsigned n)
 	}
 }
 
+static struct in6_addr in6_from_prefix_and_iid(const struct odhcpd_ipaddr *prefix, uint64_t iid)
+{
+	struct in6_addr addr;
+	uint8_t iid_len = min(128 - prefix->prefix, 64);
+
+	addr = prefix->addr.in6;
+	in6_copy_iid(&addr, iid, iid_len);
+
+	return addr;
+}
+
 void dhcpv6_ia_enum_addrs(struct interface *iface, struct dhcp_assignment *c,
 			  time_t now, dhcpv6_binding_cb_handler_t func, void *arg)
 {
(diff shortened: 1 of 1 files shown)


Alphix:
@systemcrash and @Noltari I'll leave this PR here...it's just a few lines of code but it took me many hours of staring at a very large monitor to wrap my head around this. I'll be on vacation the next week (without a computer 😱), so I won't be very useful...

I think this fixes the issues uncovered in #290 more thoroughly though :)

Noltari:
> I'll be on vacation the next week (without a computer 😱), so I won't be very useful...

Enjoy!

Noltari (approved):
LGTM

@systemcrash any objections?

systemcrash:
@Noltari This looks juicy. LGTM. 

> There are still some murky corners here though, for example, if "dhcpv6_hostidlength" is set to 64, there are no real checks that the prefix length is compatible with that (so we might end up generating an IID of 64 bits, making sure that it is unique and not taken, and then truncate it silently to a 32 bit IID in case the prefix is /96). But that's a topic for a different day (and can only be triggered by non-standard configurations).

We will *never* encounter these, ofc. 😉

Noltari:
Merged, thanks @Alphix!

Issue #265, opened 2025-10-02
Allow overriding ra_dns and ra_mtu in upstream to downstream Relay mode messages


KA2107:
Router: GL.iNet GL-MT6000 (Flint 2)
OpenWrt 24.10.3, r28872-daca7c049b
odhcpd - 2024.05.08~a2988231-r1

I have a FTTH ISP (PPPoE) which does not support Bridge Mode and many of the settings in the ONT cannot be changed in the web interface by me as a user, only the ISP technician or backend team can change those settings. 

I am using odhcpd's IPv6 Relay mode to send IPv6 RA messages from WAN to LAN.

The ONT has MTU hardcoded to 1492 bytes, whoever I was told that 1480 bytes is the most ideal MTU value. ISP provided ONT seems to provide this MTU value in it's RA message. 

Even though I have set ra_mtu = 1480 bytes in my /etc/config/dhcp config, odhcpd seems to ignore it and pass along the unmodified ra_mtu value of 1492 bytes (sent by the ONT in it's RA Relay message) from WAN to LAN. I see both Windows 11 and Arch Linux showing interface MTU as 1492 bytes instead of expected 1480 bytes.

Same thing happens for ra_dns values, odhcpd ignores the ra_dns setting and passes the unmodified DNS service address values (in ONT's RA Relay message) from WAN to LAN.

Please provide odhcpd /etc/config/dhcp config file option(s) to override ra_mtu and ra_dns values in Relay RA messages passed from WAN to LAN.

Thanks in advance.

/etc/config/network
```
config interface 'loopback'
	option device 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1/24'

config globals 'globals'
	option ula_prefix 'fd00:aaaa:aaaa:bbb0::/60'
	option packet_steering '2'
	option steering_flows '128'

config device
	option name 'br-lan_1'
	option type 'bridge'
	option bridge_empty '1'
	option mtu '1480'
	option mtu6 '1480'
	list ports 'lan1'
	list ports 'lan2'
	list ports 'lan3'
	list ports 'lan4'
	list ports 'lan5'

config interface 'lan_1'
	option device 'br-lan_1'
	option proto 'static'
	option ipaddr '192.168.114.1/24'
	option delegate '0'
	option defaultroute '0'
	option ip6assign '64'
	option ip6hint '1'
	option ip6ifaceid '::1'
	list ip6class 'wan6'
	list ip6class 'local'

config interface 'wan'
[...]

PR #296, opened 2025-10-28
router: Rewrite the ingress MTU to one configured for the interface

systemcrash:
There are edge cases where routers forward ICMPv6 RA which contain an MTU which is too large/small for the receiving client. Override the ingress MTU with the value configured on the interface.

ping @KA2107

Take this for a spin to see whether it fixes your problem.

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -257,6 +257,13 @@ static int handle_rtm_link(struct nlmsghdr *hdr)
 
 		iface->ifflags = ifi->ifi_flags;
 
+		/* store current MTU if available */
+		if (nla[IFLA_MTU]) {
+			iface->if_mtu = nla_get_u32(nla[IFLA_MTU]);
+			debug("Netlink setting interface '%s' if_mtu MTU to %d",
+					iface->name, iface->if_mtu);
+		}
+
 		/*
 		 * Assume for link event of the same index, that link changed
 		 * and reload services to enable or disable them based on the
--- a/src/router.c
+++ b/src/router.c
@@ -569,7 +569,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 	 * for shortest lived prefix
 	 */
 	uint32_t lowest_found_lifetime = UINT32_MAX, highest_found_lifetime = 0, maxival, ra_lifetime;
-	int msecs, mtu = iface->ra_mtu, hlim = iface->ra_hoplimit;
+	int msecs, hlim = iface->ra_hoplimit;
 	bool default_route = false;
 	bool valid_prefix = false;
 	char buf[INET6_ADDRSTRLEN];
@@ -602,13 +602,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 	adv.mtu.nd_opt_mtu_type = ND_OPT_MTU;
 	adv.mtu.nd_opt_mtu_len = 1;
 
-	if (mtu == 0)
-		mtu = odhcpd_get_interface_config(iface->ifname, "mtu");
-
-	if (mtu < 1280)
-		mtu = 1280;
-
-	adv.mtu.nd_opt_mtu_mtu = htonl(mtu);
+	adv.mtu.nd_opt_mtu_mtu = htonl(iface->ra_mtu);
 
 	iov[IOV_RA_ADV].iov_base = (char *)&adv;
 	iov[IOV_RA_ADV].iov_len = sizeof(adv);
--- a/src/router.c
+++ b/src/router.c
@@ -1089,19 +1089,34 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t 
 	uint8_t *mac_ptr = NULL;
 	struct in6_addr *dns_ptr = NULL;
 	size_t dns_count = 0;
+	// MTU option
+	struct nd_opt_mtu *mtu_opt = NULL;
+	uint32_t ingress_mtu_val = 0;
 
+	/* Parse existing options */
 	icmpv6_for_each_option(opt, &adv[1], end) {
-		if (opt->type == ND_OPT_SOURCE_LINKADDR) {
-			/* Store address of source MAC-address */
+		switch (opt->type) {
+		case ND_OPT_SOURCE_LINKADDR:
 			mac_ptr = opt->data;
-		} else if (opt->type == ND_OPT_RECURSIVE_DNS && opt->len > 1) {
-			/* Check if we have to rewrite DNS */
-			dns_ptr = (struct in6_addr*)&opt->data[6];
-			dns_count = (opt->len - 1) / 2;
+			break;
+
+		case ND_OPT_RECURSIVE_DNS:
+			if (opt->len > 1) {
+				dns_ptr = (struct in6_addr *)&opt->data[6];
+				dns_count = (opt->len - 1) / 2;
+			}
+			break;
+
+		case ND_OPT_MTU:
+			if (opt->len == 1 && (uint8_t *)opt + sizeof(struct nd_opt_mtu) <= end) {
+				mtu_opt = (struct nd_opt_mtu *)opt;
+				ingress_mtu_val = ntohl(mtu_opt->nd_opt_mtu_mtu);
+			}
+			break;
 		}
 	}
 
-	notice("Got a RA on %s", iface->name);
+	info("Got a RA on %s", iface->name);
 
 	/* Indicate a proxy, however we don't follow the rest of RFC 4389 yet */
 	adv->nd_ra_flags_reserved |= ND_RA_FLAG_PROXY;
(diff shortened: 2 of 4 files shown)


systemcrash:
@Alphix good if you can give this a once-over for code-quality issues.

Alphix:
Review on a cell phone screen sucks, but I think it looks sane. The only thing that seems to be missing is to clamp the overridden value to the range [RA_MTU_MIN, <underlying_MTU_of_egress_iface>]? And that should be done whenever the MTU is present in the forwarded RA message (i.e. whether ra_mtu is set or not)...

systemcrash:
Thanks. Fair point. The user configured ra_mtu is clamped at load, so by virtue of us overwriting any ingress mtu, it will anyway be replaced by a clamped value. Were you thinking of pulling some value from `/sys/class/net/xx/mtu`?

Alphix:
> Thanks. Fair point. The user configured ra_mtu is clamped at load, so by virtue of us overwriting any ingress mtu, it will anyway be replaced by a clamped value. Were you thinking of pulling some value from `/sys/class/net/xx/mtu`?

Yeah, but ra_mtu is just clamped to [1280, UINT16_MAX] when loaded, so we still need to clamp the upper end of the range to iface MTU at RA resend time. And we also need to handle clamping for when ra_mtu isn't set at all.

For the time being, I think ioctl(SIOCGIFMTU) every time the RA is forwarded is reasonable (to catch MTU changes). Longer term we should improve the netlink logic so that up-to-date values for things like the MTU and MAC are stored (and automatically kept up-to-date) in "struct iface", but that's a topic for a future PR.

systemcrash:
> For the time being, I think ioctl(SIOCGIFMTU) every time the RA is forwarded is reasonable (to catch MTU changes). Longer term we should improve the netlink logic so that up-to-date values for things like the MTU and MAC are stored (and automatically kept up-to-date) in "struct iface", but that's a topic for a future PR.

When is a reasonable time to stash these? At interface bringup time? It's not often the MTU changes, but change of MTU when an interface is up is possible IIRC (`ip link set mtu "${mtu}" dev "${iface}"`), but I don't think such complexity is necessary, although it certainly covers all the bases.

> Yeah, but ra_mtu is just clamped to [1280, UINT16_MAX] when loaded, so we still need to clamp the upper end of the range to iface MTU at RA resend time. And we also need to handle clamping for when ra_mtu isn't set at all.

I figure - grab the MTU at config init, and use that as clamp ceiling.

Alphix:
> When is a reasonable time to stash these? At interface bringup time? It's not often the MTU changes, but change of MTU when an interface is up is possible IIRC (`ip link set mtu "${mtu}" dev "${iface}"`), but I don't think such complexity is necessary, although it certainly covers all the bases.

I think that odhcpd already gets a netlink event and reloads when the MTU on the underlying device changes, so yes, at interface recfg time should be enough (thats in handle_rtm_link() in src/netlink.c IIRC). But feel free to add the current version without clamping the upper bound to the link MTU, we can add it later.

> > Yeah, but ra_mtu is just clamped to [1280, UINT16_MAX] when loaded, so we still need to clamp the upper end of the range to iface MTU at RA resend time. And we also need to handle clamping for when ra_mtu isn't set at all.
> 
> I figure - grab the MTU at config init, and use that as clamp ceiling.

Basically, yes, grab the MTU, stash it in struct interface

systemcrash:
Updated. This seems functionally complete. I've tested the clamping part, and verified the rewrite:
```
[30 Oct 2025, 17:29:23 CET] daemon.debug: odhcpd[2103]: Stashed MTU of 1500 for interface 'wan6'
[30 Oct 2025, 17:29:23 CET] daemon.debug: odhcpd[2103]: Enabling services with eth1 running
[30 Oct 2025, 17:29:23 CET] daemon.notice: odhcpd[2103]: Sending RS to wan6
[30 Oct 2025, 17:29:23 CET] daemon.debug: odhcpd[2103]: Sent 8 bytes to ff02::2%wan6@eth1
[30 Oct 2025, 17:29:23 CET] daemon.notice: netifd: Interface 'wan6' has link connectivity
[30 Oct 2025, 17:29:24 CET] daemon.debug: odhcpd[2103]: Received 152 Bytes from fe80::9683:c4ff:fea5:3ec6%wan6@eth1
[30 Oct 2025, 17:29:24 CET] daemon.notice: odhcpd[2103]: Got a RA on wan6
[30 Oct 2025, 17:29:24 CET] daemon.debug: odhcpd[2103]: Rewriting RA MTU from 1700 to 1500 on lan
[30 Oct 2025, 17:29:24 CET] daemon.notice: odhcpd[2103]: Forward a RA on lan
```

LAN side RA
```
Internet Control Message Protocol v6
    Type: Router Advertisement (134)
    Code: 0
    Checksum: 0xb6dd [correct]
    [Checksum Status: Good]
    Cur hop limit: 64
    Flags: 0x04, Prf (Default Router Preference): Medium, ND Proxy
    Router lifetime (s): 1800
    Reachable time (ms): 0
    Retrans timer (ms): 0
    ICMPv6 Option (Source link-layer address : xx:xx:xx:b6:b1:89)
    ICMPv6 Option (MTU : 1500)
        Type: MTU (5)
        Length: 1 (8 bytes)
        Reserved
        MTU: 1500
    ICMPv6 Option (Prefix information : 2001:db8:20:1e00::/64)
    ICMPv6 Option (Prefix information : fd51:1c2a:8909::/64)
    ICMPv6 Option (Recursive DNS Server fd51:1c2a:8909::1)
    ICMPv6 Option (DNS Search List Option lan)
    ICMPv6 Option (Advertisement Interval : 20000)
```

systemcrash:
Thoughts on this @Noltari ?

src/netlink.c
```diff
@@ -257,6 +257,13 @@ static int handle_rtm_link(struct nlmsghdr *hdr)
 
 		iface->ifflags = ifi->ifi_flags;
 
+		/* store current MTU if available */
+		if (nla[IFLA_MTU]) {
+			iface->if_mtu = nla_get_u32(nla[IFLA_MTU]);
+			debug("Stashed MTU of %d for interface '%s'",
```
Alphix: I'd say that the "stashing" part is less interesting than reporting where the MTU value came from....how about something like "Netlink setting interface '%s' MTU to %d"? (I tried to match the wording to what is used in the rest of `netlink.c`)

src/router.c
```diff
@@ -1146,8 +1155,16 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+		/* Rewrite MTU option if local RA MTU is configured */
+		if (c->ra_mtu && mtu_opt) {
+			if (ingress_mtu_val != c->ra_mtu) {
+				debug("Rewriting RA MTU from %u to %u on %s",
+				       ingress_mtu_val, c->ra_mtu, c->name);
+				mtu_opt->nd_opt_mtu_mtu = htonl(c->ra_mtu);
+			}
+		}
 
+		notice("Forward a RA on %s", c->name);
```
Alphix: Maybe drop the prio here a bit? Forwarding a RA doesn't seem like a "normal but significant condition"...more like `info()` or `debug()`?

Alphix:
@systemcrash Overall, the change looks pretty straightforward, just some small nits wrt logging...other than that, LGTM

src/router.c
systemcrash: It was like that before my commit, but I'll amend it

src/netlink.c
systemcrash: Done

src/router.c
Alphix: Yeah, I know...but of course, the "+" lines in a patch automatically draw attention to themselves 😃

systemcrash:
Oh.
```
Ign:5 https://security.ubuntu.com/ubuntu noble-updates/main amd64 binutils-aarch64-linux-gnu amd64 2.42-4ubuntu2.5
Err:5 mirror+file:/etc/apt/apt-mirrors.txt noble-updates/main amd64 binutils-aarch64-linux-gnu amd64 2.42-4ubuntu2.5
  404  Not Found [IP: 52.147.219.192 80]
Fetched 41.4 MB in 1s (48.7 MB/s)
E: Failed to fetch mirror+file:/etc/apt/apt-mirrors.txt/pool/main/b/binutils/binutils-aarch64-linux-gnu_2.42-4ubuntu2.5_amd64.deb  404  Not Found [IP: 52.147.219.192 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

Alphix:
That's unrelated to your PR though? Some mirror wonkiness?

Noltari:
> That's unrelated to your PR though? Some mirror wonkiness?

That should be fixed now with https://github.com/openwrt/odhcpd/commit/d174e25e85a1fd703c36f39e154042cbc0a3e36a

Noltari (approved):
LGTM

systemcrash:
We can also close #265

PR #297, opened 2025-11-02
router: always use link-local src address for RAs

Alphix:
This is a follow-up to:
https://github.com/openwrt/odhcpd/pull/242

As noted by @willmo [1], RAs are currently not strictly limited to using a link-local source address, which they should according to RFC4861, §4.2.

This is usually not an issue, since router solicitations typically originate from link-local source addresses or the undefined address, meaning that odhcpd will reply with its own link-local address (auto-selected by the kernel).

But if a solicitation is sent from e.g. a GUA/ULA address, odhpcd will currently reply using it's own GUA/ULA address.

While fixing this, correct some misleading comments.

[1] https://github.com/openwrt/odhcpd/pull/242#issuecomment-3475020864

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -1002,7 +1002,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 
 	debug("Sending a RA on %s", iface->name);
 
-	if (odhcpd_send(iface->router_event.uloop.fd, &dest, iov, ARRAY_SIZE(iov), iface) > 0) {
+	if (odhcpd_try_send_with_src(iface->router_event.uloop.fd, &dest, iov, ARRAY_SIZE(iov), iface) > 0) {
 		iface->ra_sent++;
 
 		config_save_ra_pio(iface);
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari and @systemcrash pls have a look

Noltari (approved):

willmo (approved):

systemcrash (approved):

systemcrash:
Can we re-word the commit message a bit?: 


```
As noted by @willmo [1], RAs are currently not strictly limited to using a
link-local source address, which they should according to RFC4861, §4.2:

      Source Address
                     MUST be the link-local address assigned to the
                     interface from which this message is sent.

This is usually not an issue, since router solicitations typically
originate from link-local source addresses or the undefined address,
meaning that odhcpd will reply with its own link-local address
(auto-selected by the kernel).

But if a solicitation is sent from e.g. a GUA/ULA address, odhpcd replied
using its own GUA/ULA address.

This is now changed so that it only replies using its LLA.

While fixing this, correct some misleading comments.
```

systemcrash:
Ah nuts :)

Noltari:
> Can we re-word the commit message a bit?:

Bad timing >.<

Noltari:
Merged, thanks @Alphix!

systemcrash:
Just the message speaks in present tense and doesn't allude to how behaviour changed with the commit :/

PR #298, opened 2025-11-04
dhcpv4: lazy store statefiles

Alphix:
Right now, dhcpv4 saves the statefile on every change and also calls the leasetrigger script on every change (which triggers a dnsmasq reload).

With this change, the statefile is written every N seconds, assuming that there are any changes.

Before this change (without leasetrigger):

```shell
$ time sudo ./build/dhcpdig -4 benchmark foo-client

Thread[**]: req 10000101/0 rel 10000101/0 rep 10000101/0

real	55m29.406s
user	0m0.005s
sys	0m0.010s
```

(This is a simple benchmark tool I wrote, it runs 10 threads which divide the DHVPv4 pool into 10 chunks and then performs 1 million random addr req/release per thread in a loop).

After this change:
```shell
$ time sudo ./build/dhcpdig -4 benchmark foo-client

Thread[**]: req 10000101/0 rel 10000101/0 rep 10000101/0

real	1m48.123s
user	0m0.005s
sys	0m0.005s
```

A 3082% speedup.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -344,6 +344,7 @@ static void set_interface_defaults(struct interface *iface)
 	iface->ra_lifetime = 3 * iface->ra_maxinterval; /* RFC4861: AdvDefaultLifetime: Default: 3 * MaxRtrAdvInterval */
 	iface->ra_dns = true;
 	iface->pio_update = false;
+	iface->update_statefile = true;
 }
 
 static void clean_interface(struct interface *iface)
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -336,6 +336,9 @@ void dhcpv4_free_lease(struct dhcpv4_lease *lease)
 	if (lease->fr_ip)
 		dhcpv4_fr_stop(lease);
 
+	if (lease->iface)
+		lease->iface->update_statefile = true;
+
 	list_del(&lease->head);
 	if (lease->lease_cfg)
 		lease->lease_cfg->dhcpv4_lease = NULL;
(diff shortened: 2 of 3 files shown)


Alphix:
@systemcrash @Noltari ping

Alphix:
(and a similar thing can be done for the DHCPv6 case as well, of course, but that's for a different PR)

systemcrash:
Juicy. Simple. Like.

You bumped the loop up to 5000 msec yes? Seems OK.

src/config.c
```diff
@@ -344,6 +344,7 @@ static void set_interface_defaults(struct interface *iface)
 	iface->ra_lifetime = 3 * iface->ra_maxinterval; /* RFC4861: AdvDefaultLifetime: Default: 3 * MaxRtrAdvInterval */
 	iface->ra_dns = true;
 	iface->pio_update = false;
+	iface->update_statefile = true;
```
Noltari, line 347: Why is this initialized to `true`?

Noltari (approved):
LGTM

src/config.c
Alphix: It's to make sure that any old state file will be overwritten with an empty state file when odhcpd is restarted...

Right now odhcpd loses all state on restart... That's something I plan to fix later (by reading the old state file)

Alphix:
> You bumped the loop up to 5000 msec yes? Seems OK.

Yeah, I took some inspiration from the polling intervals in LuCI which currently seem to be around 5 secs. If it turns out to be a problem, we can always lower it again...

Noltari:
Merged, thanks @Alphix!

PR #299, opened 2025-11-05
dhcpv4: add support for RFC4361-style clientid

Alphix:
This adds support for DHCPv6-style client identifiers to the dhcpv4 codebase, as per RFC4361. It also ensures RFC6842-compliance by including the client identifier (no matter which kind it is) in replies.


Alphix:
@Noltari @systemcrash This is the last piece that is missing in the DUID puzzle (adds DUID/IAID support for DHCPv4 clients as well, there's no equivalent for the DHCPv4 server to identify itself). Right now it'll conflict with some of the other PRs, and it needs a bit more testing, but I'm posting this here now so that you can review it.

src/odhcpd.h
```diff
@@ -249,6 +249,11 @@ struct dhcpv4_lease {
 	unsigned fr_cnt;			// FR messages sent
 	uint8_t key[16];			// FR nonce
 	struct odhcpd_ref_ip *fr_ip;		// FR message old serverid/IP
+
+	// RFC4361
+	uint32_t iaid;
+	uint8_t duid_len;
+	uint8_t duid[];
```
systemcrash, line 256: could we add a duid struct instead?

src/dhcpv4.c
```diff
@@ -345,21 +345,28 @@ void dhcpv4_free_lease(struct dhcpv4_lease *lease)
...
+		   size_t hwaddr_len, const uint8_t *duid, size_t duid_len,
+		   uint32_t iaid)
 {
 	struct dhcpv4_lease *lease;
 
 	if (!iface || !hwaddr || hwaddr_len == 0 || hwaddr_len > sizeof(lease->hwaddr))
 		return NULL;
 
-	lease = calloc(1, sizeof(*lease));
+	lease = calloc(1, sizeof(*lease) + duid_len);
```
systemcrash, line 360: we need duid_len because it's the only variable length structure?

src/odhcpd.h
Alphix: We could, and I considered it, but: 

a) it would make it less clear that "struct dhcpv4_lease" has a Flexible Array Member (which must come last).

b) to be really pedantic, the struct isn't the same for "struct dhcpv4_lease" and "struct dhcpv6_lease" since the length is an uint8_t Vs uint16_t (sure, we could use a uint16_t for both)

The advantage of such a struct is that we could pass it instead of three arguments to various functions...

Note to self: clid_len and clid_data should be renamed in struct dhcpv6_lease

src/dhcpv4.c
Alphix: Yeah, because the struct has a flexible array member (the DUID)

src/odhcpd.h
Alphix: (note that we already have a struct duid, but it's for use in struct lease_cfg)

systemcrash: Yeah, I had in mind the duid struct which is just above this, which is:

```c
struct duid {
	uint8_t len;
	uint8_t id[DUID_MAX_LEN];
	uint32_t iaid;
	bool iaid_set;
};
```

it reduces argument passing. Altho pedantic: do we increase risks when passing pointers and culling old records that might otherwise be in use?

Alphix: Risks when passing pointers? Not sure I follow...?

systemcrash: I was thinking about how the duid structs are separate from the lease structure, and maybe this gives rise to some errant handling of the data, but the lease structure has other pointers to other structs so I don’t see why it’s a bad idea.

Alphix: But they're not separate? In the current version the DUID is part of the lease struct and even if I replaced the flexible array member with a struct duid, it would still be embedded in the lease struct (as a struct-in-struct)

systemcrash: You're right, they're not separate - I meant "would be separate" and thus could be problematic. Sorry for the confusion.

Alphix: The current `struct duid` also has the member `iaid_set` which is necessary for DUIDs that come from uci `host` sections, but are superfluous/confusing in this context.

Also, `struct duid` has a fixed memory allocation of 130 bytes for the DUID, whereas the current version of the patch will only allocate exactly the number of bytes actually used for the DUID (not that it's a lot of memory, but still).

And another thing is that we won't really be able to pass `struct duid` to the matching functions, because we typically pass pointers straight into the relevant part of the received DHCPv4/6 msg...(meaning that we'd have to do `memcpy()`s into a temporary `struct duid` just to call some functions)...

Alphix:
@Noltari @systemcrash Is this good to go?

systemcrash:
Yeah, good.

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

Issue #172, opened 2021-07-10
RA relay mode copies all RA flags as-is, regardless of DHCPv6 relay state


Shine-:
### Problem description:
odhcpd in RA relay mode always copies RA flags (e.g. M, O, A) to downlink as-is. It doesn't take into account that these flags relate to DHCPv6, not RA and should therefore be set according to **downlink** DHCPv6 configuration.

Note: tested in OpenWRT 19.07.x (with old ra_management syntax, same behavior), 21.02.0rc2, 21.02.0rc3 and several snapshots past 21.02.0rc3.

### Current behavior, example 1:
Uplink provides RA with SLAAC only, but no DHCP service. Therefore, uplink's RA has M=0, O=0 and A=1 for the given prefix.
Downlink is able to assign addresses via SLAAC from the uplink prefix, thanks to odhcp's RA relay mode.
Downlink does have an own DHCPv6 service running, e.g. in order to assign DNS servers and/or ULA.
However, since odhcpd copies M=0 and O=0 to downlink, any clients connected to downlink ignore the locally available DHCPv6 service, and won't get either DNS servers or ULA.
This is not hypothetical, I'm facing exactly **this** issue!
Note that clients sometimes receive their DHCPv6 info just fine, but drop it again shortly afterwards - depending on what comes first, the DHCP reply or the RA. Can be nicely viewed in Wireshark.

### Current behavior, example 2:
Uplink provides both, SLAAC and DHCP. Therefore, uplink's RA has M=1, O=1 and A=1 for the given prefix.
Downlink prefers SLAAC-only and chooses not to relay DHCPv6 from uplink, nor to run DHCPv6 locally.
However, due to M=1 and O=1 being copied as-is from uplink to downlink, downlink's clients keep firing DHCPv6 requests and would gladly receive addresses/other configuration from any rogue DHCPv6 server -> security problem!

### Current behavior, example 3:
(Note: This is not incorrect behavior per se, but limits specific use-cases) Uplink is known to provide DHCP, and may or may not offer SLAAC in addition. Therefore, uplink's RA has M=1, O=1 and A=0 or 1 for the given prefix. Downlink prefers DHCPv6-only and doesn't want their clients to auto-self-assign IAs, even if it were possible. So downlink would like to set A=0, regardless of whether uplink actually offers SLAAC or not. With flags always being relayed as-is, downlink doesn't have this possibility, all clients will happily self-assign IAs if offered by uplink.

### Expected behavior:

[...]

Issue #253, opened 2025-09-17
[RFC9762] (Proposed Standard) Add P flag in prefix information options


zekica:
This is a new feature that isn't a standard yet (at least that's my understanding) but reading [Android developers blog post](https://android-developers.googleblog.com/2025/09/simplifying-advanced-networking-with.html) they plan on implementing it and backporting all the way to Android 11 using "Google Play System Update".

This new flag signals the preference of the network that even non-router devices should use PD and not assign IA_NA or SLAAC addresses.

This new flag should be defined as something like:
`#define ND_OPT_PI_FLAG_PDPREF 0x10`
and set in `send_router_advert` based on a new configuration option.

I am not sure whether this is doable before musl/uclibc/glibc add that flag in their `netinet/icmp6.h` header files.

brada4:
XREF to unsplit horizons https://forum.openwrt.org/t/support-for-p-flag-in-ipv6-router-advertisement-rfc-9762/240666

GoetzGoerisch:
Android already assignes a /64 PD without any P flag if DHCPv6-PD is enabled in the subnet.

systemcrash:
How should this be presented as an option to the administrator?

My read of the RFC:

The RFC defines the P flag for use within PIOs, and defines its meaning
as "purely a positive indicator, telling nodes that DHCPv6 PD is
available and the network prefers that nodes use it".

The RFC abstract lays out:

The (P) flag is used to indicate that the network prefers that clients
use the deployment model in RFC 9663 instead of using individual
addresses in the on-link prefix assigned using SLAAC, or DHCPv6 address
assignment.

It's also suggested that the PIO P bit is configurable as a per PIO option,
and not simply a global one.

So the choices are:
1. RFC9663 (DHCPv6-PD) - set PIO P bit
2. SLAAC - set PIO A bit
3. DHCPv6 (single) address assignment - set neither PIO A nor P bit(?)

A GUI should indicate a choice between one of these three (and these
code changes then set the A and P bits accordingly).


So I guess prerequisites are:
1. DHCPv6 server mode, and you have tons of address space.
2. SLAAC is on
3. ? relay?

brada4:
2a / it is opportunistic, if you have no more /64 spare you have to hand out individual addresses.
3 / their problem

systemcrash:
[...]

Issue #243, opened 2025-07-27
Add network to the DHCP section and specify a network segment as the DHCP address pool


peterwillcn:
The problem is to distinguish between two subnets under the same interface and specify subnetworks as the default address pool

 **option network '172.16.1.0/24'  # It seems to have no effect**

dhcp
```
config dhcp 'lan'
        option interface 'lan'
        option start '100'
        option limit '150'
        **option network '172.16.1.0/24'  # It seems to have no effect**
        option leasetime '12h'
        option dhcpv4 'server'
        option dhcpv6 'server'
        option ra 'server'
        option ra_slaac '1'
        list ra_flags 'managed-config'
        list ra_flags 'other-config'
```
network
```
config interface 'lan'
    option ifname 'eth1'
    option proto 'static'
    list ipaddr '172.16.1.1/24'  
    list ipaddr '192.168.2.1/24'  
```

Alphix:
Perhaps `prefix_filter` could be extended to:
a) take an array instead of a string
b) accept both IPv4 and IPv6 addresses in CIDR notation

Changing `start` and `limit` would be other options, but that risks causing complications with `dnsmasq` compatibility...

PR #300, opened 2025-11-05
router:  Modify relayed RA PIO L and RA M/O flags according to interface policy

systemcrash:
I found this patch lurking in #172 - seems OK. Updated for current master. 

Thoughts @Noltari @Alphix ?

h/t user @Shine-

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -1099,7 +1099,7 @@ static void forward_router_solicitation(const struct interface *iface)
 }
 
 
-/* Forward a router advertisment from master to slave interfaces */
+/* Forward a router advertisement from master to slave interfaces */
 static void forward_router_advertisement(const struct interface *iface, uint8_t *data, size_t len)
 {
 	struct nd_router_advert *adv = (struct nd_router_advert *)data;
@@ -1115,6 +1115,32 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t 
 	// MTU option
 	struct nd_opt_mtu *mtu_opt = NULL;
 	uint32_t ingress_mtu_val = 0;
+	/* PIO L flag and RA M/O Flags */
+	uint8_t ra_flags;
+	size_t pio_count = 0;
+	struct fwd_pio_flags {
+		uint8_t *ptr;
+		uint8_t flags;
+	} *pio_flags = NULL;
+
+	icmpv6_for_each_option(opt, &adv[1], end) {
+		/* check our packet content is not truncated */
+		if (opt->len == 0 || (uint8_t *)opt + opt->len * 8 > end) {
+			error("Ingress RA packet option for relaying has incorrect length");
+			return;
+		}
+
+		switch(opt->type) {
+		case ND_OPT_PREFIX_INFORMATION:
+			pio_count++;
+			break;
+		}
+	}
+
+	if (pio_count > 0) {
+		pio_flags = alloca(sizeof(*pio_flags) * pio_count);
+		pio_count = 0;
+	}
 
 	/* Parse existing options */
 	icmpv6_for_each_option(opt, &adv[1], end) {
@@ -1136,13 +1162,21 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t 
 				ingress_mtu_val = ntohl(mtu_opt->nd_opt_mtu_mtu);
 			}
 			break;
+		case ND_OPT_PREFIX_INFORMATION:
+			/* Store options for each PIO */
+			pio_flags[pio_count].ptr = &opt->data[1];
+			pio_flags[pio_count].flags = opt->data[1];
+			pio_count++;
+			break;
 		}
 	}
 
 	info("Got a RA on %s", iface->name);
 
-	/* Indicate a proxy, however we don't follow the rest of RFC 4389 yet */
-	adv->nd_ra_flags_reserved |= ND_RA_FLAG_PROXY;
+	/*	Indicate a proxy, however we don't follow the rest of RFC 4389 yet
+	 *	store original upstream RA state 
+	 */
+	ra_flags = adv->nd_ra_flags_reserved | ND_RA_FLAG_PROXY;
 
 	/* Forward advertisement to all slave interfaces */
 	memset(&all_nodes, 0, sizeof(all_nodes));
@@ -1156,6 +1190,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t 
 		/* Fixup source hardware address option */
 		if (mac_ptr)
 			odhcpd_get_mac(c, mac_ptr);
+
+		if (pio_count > 0)
+			debug("RA forward: Rewriting RA PIO flags");
+
+		for (size_t i = 0; i < pio_count; i++) {
+			/* restore the flags byte to its upstream state before applying per-interface policy */
+			*pio_flags[i].ptr = pio_flags[i].flags;
+			/* ensure L flag (on-link) cleared; relayed == not on-link */
+			*pio_flags[i].ptr &= ~ND_OPT_PI_FLAG_ONLINK;
+		}
+
+		/* Apply per-interface modifications of upstream RA state */
+		adv->nd_ra_flags_reserved = ra_flags;
+		/* Rewrite M/O flags unless we relay DHCPv6 */
+		if (c->dhcpv6 != MODE_RELAY) {
+			/* Clear the relayed M/O bits */
+			adv->nd_ra_flags_reserved &= ~(ND_RA_FLAG_MANAGED | ND_RA_FLAG_OTHER);
+			/* Apply the locally configured ra_flags for M and O */
+			adv->nd_ra_flags_reserved |= c->ra_flags & (ND_RA_FLAG_MANAGED | ND_RA_FLAG_OTHER);
+		}
 
 		/* If we have to rewrite DNS entries */
 		if (c->always_rewrite_dns && dns_ptr && dns_count > 0) {


Shine-:
Wow, I'd love to see this applied, after almost 5 years :-)

systemcrash:
> Wow, I'd love to see this applied, after almost 5 years :-)

Yeah man, really unfortunate it took so long.

src/router.c
```diff
@@ -1113,13 +1118,29 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+				? realloc(pfx_flags, (pfx_count+1) * sizeof *pfx_flags)
+				: malloc(sizeof *pfx_flags);
+			if (pfx_flags_new) {
+				pfx_flags = pfx_flags_new;
+				pfx_flags[pfx_count].ptr = &opt->data[1];
+				pfx_flags[pfx_count].flags = opt->data[1];
+				pfx_count++;
+			}
+			else {
+				error(LOG_ERR, "Couldn't allocate memory for relayed IPv6 prefix options");
```
Alphix: This looks bogus...should be `error("Couldn't...` right?

src/router.c
```diff
@@ -1134,6 +1155,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 		if (mac_ptr)
 			odhcpd_get_mac(c, mac_ptr);
 
+		/* Rewrite A flag for each PIO where downlink interface disallows SLAAC */
+		if (pfx_flags) {
```
Alphix: Not necessary, if pfx_flags == NULL, pfx_count == 0, so just do the `for` loop

src/router.c
```diff
@@ -1134,6 +1155,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+			for (size_t i=0; i < pfx_count; ++i) {
```
Alphix:
```C
for (size_t i = 0; i < pfx_count; i++) {
```

src/router.c
```diff
@@ -1113,13 +1118,29 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+			pfx_flags_new = pfx_flags
+				? realloc(pfx_flags, (pfx_count+1) * sizeof *pfx_flags)
+				: malloc(sizeof *pfx_flags);
+			if (pfx_flags_new) {
+				pfx_flags = pfx_flags_new;
+				pfx_flags[pfx_count].ptr = &opt->data[1];
+				pfx_flags[pfx_count].flags = opt->data[1];
+				pfx_count++;
+			}
+			else {
```
Alphix: Wonky brace formatting

src/router.c
```diff
@@ -1134,6 +1155,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 		if (mac_ptr)
 			odhcpd_get_mac(c, mac_ptr);
 
+		/* Rewrite A flag for each PIO where downlink interface disallows SLAAC */
+		if (pfx_flags) {
+			for (size_t i=0; i < pfx_count; ++i) {
+				*pfx_flags[i].ptr = pfx_flags[i].flags;
+				if (!c->ra_slaac) {
```
Alphix: No braces for single-line if body

src/router.c
```diff
@@ -1134,6 +1155,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+					*pfx_flags[i].ptr &= ~ND_OPT_PI_FLAG_AUTO;
```
Alphix: I might be wrong, but we clear the flag if iface->ra_slaac isn't set....shouldn't we also _set_ the flag if iface->ra_slaac _is_ set?

src/router.c
```diff
@@ -1113,13 +1118,29 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 				ingress_mtu_val = ntohl(mtu_opt->nd_opt_mtu_mtu);
 			}
 			break;
+		case ND_OPT_PREFIX_INFORMATION:
+			/* Store prefix options for each prefix */
+			pfx_flags_new = pfx_flags
+				? realloc(pfx_flags, (pfx_count+1) * sizeof *pfx_flags)
+				: malloc(sizeof *pfx_flags);
```
Alphix: `sizeof` should be written like a function (`sizeof()`), and `realloc()` accepts `NULL` just fine, so if you want to go down this route:

```C
pfx_flags_new = realloc(pfx_flags, (pfx_count + 1) * sizeof(*pfx_flags));
```

**But**: you have a memory leak here...`pfx_flags` is never `free()`d

Personally, I'd suggest a double loop, avoiding all `malloc()`/`realloc()` calls.

In addition, there's no need to store the pointer to the flags and the pointed-to-flags (if we assume that I'm right and the `A` flag should always be set/unset)...something like this (not tested, just meant to show what I mean):

```C

size_t pfx_count = 0;
uint8_t *pfx_flags[] = NULL;

...

icmpv6_for_each_option(opt, &adv[1], end) {
   switch(opt->type) {
       ...
       case ND_OPT_PREFIX_INFORMATION:
           pfx_count++;
           break;
       ...
}

if (pfx_count > 0) {
    size_t i = 0;
    pfx_flags = alloca(sizeof(*pfx_flags) * pfx_count);

    icmpv6_for_each_option(opt, &adv[1], end) {
        if (opt->type == ND_OPT_PREFIX_INFORMATION)
            pfx_flags[i++] = &opt->data[1];
    }
}

...
avl_for_each_element(&interfaces, c, avl) {
    ...
    for (size_t i = 0; i < pfx_count; i++) {
        if (c->ra_slaac)
            *pfx_flags[i] &= ~ND_OPT_PI_FLAG_AUTO
        else
            *pfx_flags[i] |= ND_OPT_PI_FLAG_AUTO
    }
    ...
}
```

(And if so....you can ignore the other comments...)

Alphix:
Also, there's a mix of `/* single-line comments */` and `// single-line comments`

src/router.c
systemcrash: Absolutely.

src/router.c
Shine-: True, @systemcrash forgot the `free(pfx_flags);` further down, my original patch has it.

src/router.c
Shine-: From my original patch description: "For prefix-related flags, specifically A, **relay** it unconditionally if it's set to 0 at uplink, but honor ra_slaac if it's set to 1 at uplink." We want to do whatever uplink allows, **unless** our own ra_slaac forbids it. We must keep the logic, we mustn't ever set A if uplink doesn't allow it!

Alphix: Ok, I read that part...but it's late, and I'm tired...and the quote describes what the code does...can you describe the thinking behind it? As in, the "why" rather than the "what"?

Shine-: If uplink doesn't want us to do SLAAC, we mustn't **ever** do it. So either we relay what uplink wants, or we **don't** allow SLAAC (if **we** don't want it). So we will never ever set ra_slaac ourselves.

src/router.c
Shine-: There was a reason why I specifically used malloc in the case of NULL. I really don't remember details from years ago, but after reading into some documentation, I decided it's safer to not use realloc on NULL.
Edit: Or, I simply didn't want to have 2 unnecessary arithmetic operations in place...

src/router.c
Alphix: I'm probably slow (and I'm not trying to annoy you), but just saying "we mustn't" doesn't really help me understand. Why must we never do it? Why can't different network segments have different policies wrt. SLAAC?

Shine-: This is about **relay**. If your uplink does NOT allow you to do SLAAC, you MUST NOT do it. So far, so good, and that's how odhcpd was already working before. But what if your uplink allows SLAAC, but **you** don't want it for any reason? The previous behavior of odhcpd was, you can do whatever you want, your clients were always offered SLAAC as allowed from your uplink network. You couldn't prevent it.

The new behavior is, you can - even if uplink would allow SLAAC - forbid it on your downlink.

But still, if your uplink **does not** allow SLAAC, you **must never** do it. You can't arbitrarily set the A flag if your uplink connection (e.g. your ISP) doesn't allow it. At best, your clients would lose IPv6 connectivity, because your ISP doesn't route the illegaly allocates addresses, At worst, you're causing IP conflicts.

Alphix: And what if uplink isn't my ISP?

Shine-: Then replace "ISP" in my text above with whatever your uplink is?

I don't understand now, what else is unclear in my explanation that I have now repeated multiple times? It's also late for me and I'd also like to go to sleep, not repeat myself over and over again...

src/router.c
systemcrash: Gah. I missed the free there. My bad.

@Alphix thanks for the feedback. I decided to skip restoring the original upstream flags, since each PIOs flags depend on the downstream interface. L? it's likely not on-link. A? autoassign. This makes me wonder whether we should do something similar for the M+O flags in the RA. I'll look at this later.

Shine-: Edit: Moved this text down, doesn't belong here.

Shine-:
Please don't **remove** the relay funcionality here. My patch was not meant to remove DHCP/NDP relay functionality from odhcpd, it was meant to give you (or rather: me) more control of what I want to relay and what I don't want to relay. I was **never** planning to remove any functionality, but I have the fear that with this discussion, it's going into that direction.

Please also read and try to understand the comments in #172. This was carefully thought out to not break anything. The thoughts I read here, however, are basically planning to break DHCP/NDP relay, not give the user additional control over it.

src/router.c
Alphix: > And what if uplink isn't my ISP?

Example 1:
"upstream" router A and "downstream" router B are both on the same network and operated by the same org. `WAN <-> Router A <-> Router B <-> Downstream hosts`. Router A announces a `/64` with, say, `M=1, O=1, A=0`, router B relays but sets `M=0, O=0, A=1`. NDP/RA proxying and DAD will still prevent address conflicts. The choice of `SLAAC` on a given network segment is a policy decision and I don't see why it would be a `MUST NOT`?

Example 2:
Assuming an ISP cfg which seems common in parts of Asia (Japan, Korea) where a single large prefix is advertised and the ISP expect all devices behind the home router to get globally routable addresses directly from that prefix. Overriding `A` to `A=0` in that case would break the ability for downstream hosts to get a GUA (and that is allowed with the current logic in this patch).

So why is one example kosher and the other one isn't?

Alphix:
> Please don't **remove** the relay funcionality here.

I don't think that has been suggested.

Alphix:
@systemcrash the new version looks better...I would still suggest a separate loop to determine `pfx_count` followed by a single stack allocation (it'll just be one ptr per PIO) though.

Also, a `Signed-off-by` by @Shine- is missing...?

src/router.c
```diff
@@ -1092,6 +1092,11 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 	// MTU option
 	struct nd_opt_mtu *mtu_opt = NULL;
 	uint32_t ingress_mtu_val = 0;
+	/* PIO L/A flag and RA M/O Flags */
+	uint8_t raflags;
```
Noltari: `ra_flags` to be in line with other variable names.

src/router.c
```diff
@@ -1113,13 +1118,28 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+			pio_flags_new = realloc(pio_flags, (pio_count + 1) * sizeof(*pio_flags));
+			if (pio_flags_new) {
+				pio_flags = pio_flags_new;
+				pio_flags[pio_count].ptr = &opt->data[1];
+				pio_count++;
+			} else {
+				error("Couldn't allocate memory for relayed IPv6 prefix options");
+				free(pio_flags);
+				pio_flags = NULL;
+				break;
```
Noltari: This `break` can be removed.

src/router.c
```diff
@@ -1134,6 +1154,27 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 		if (mac_ptr)
 			odhcpd_get_mac(c, mac_ptr);
 
+		/* Set the A flag for each PIO where downlink interface allows SLAAC */
+		for (size_t i = 0; i < pio_count; i++) {
+			uint8_t flags = 0;
+			/* L bit: prefix is not on-link when relayed */
+			/* A bit: depends on whether we allow SLAAC */
```
Noltari:
```
/* L bit: prefix is not on-link when relayed
 * A bit: depends on whether we allow SLAAC
 */
```

src/router.c
```diff
@@ -1113,13 +1118,28 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
...
+			}
+			break;
 		}
 	}
 
 	info("Got a RA on %s", iface->name);
 
 	/* Indicate a proxy, however we don't follow the rest of RFC 4389 yet */
-	adv->nd_ra_flags_reserved |= ND_RA_FLAG_PROXY;
+	/* store original upstream RA state */
```
Noltari:
```
/* Indicate a proxy, however we don't follow the rest of RFC 4389 yet
 * store original upstream RA state
 */

File: src/router.c [hunk 12]
```diff
@@ -1092,6 +1092,11 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 	// MTU option
 	struct nd_opt_mtu *mtu_opt = NULL;
 	uint32_t ingress_mtu_val = 0;
+	/* PIO L/A flag and RA M/O Flags */
+	uint8_t raflags;
+	size_t pio_count = 0;
+	struct { uint8_t *ptr; } *pio_flags = NULL, *pio_flags_new;
```
Noltari: Personally I don't like these types of declarations.
Can we add a proper struct declaration for this and use it here?

Noltari (changes requested):

src/router.c
Alphix: Yeah, a single-member struct is a bit weird...use an array?

Noltari:
> Wow, I'd love to see this applied, after almost 5 years :-)

Well, to be fair you never submitted a Pull Request (attaching patches to issues isn't the best way to get things fixed).

@Shine- please provide your SoB so that @systemcrash can add it to the commit.
`Signed-off-by: Real Name <real@email>`

Shine-:
> > Please don't **remove** the relay funcionality here.
> 
> I don't think that has been suggested.

The current plan seems to be to completely remove the ability to relay the A flag, and hardcode that into the code. That is feature removal in my opinion.

Alphix:
@systemcrash I wonder if your observation wrt clearing the L flag might help with some of the relaying issues that are reported in the repo right now...

src/router.c
Shine-: Example 1: It doesn't matter who is actually operating Router A. If it's router A's policy to not allow self-allocation of IPv6 addresses in their network, and Router B relays their RA, the org behind Router B still has to abide by Router A's policy. They can't arbitrarily set A=1 and illegaly allocate IPv6 addresses from Router A's range. I'm confident that there is also some RFC that disallows such rogue behavior.

Example 2: In that case, you simply don't override A=0, but relay it as-is. It's **choice** after all. In my proposal at least.

Is any of you actually **using** DHCPv6/NDP relay, or are you all just talking theoretically, without having an actual use case?

Shine-:
> > Wow, I'd love to see this applied, after almost 5 years :-)
> 
> Well, to be fair you never submitted a Pull Request (attaching patches to issues isn't the best way to get things fixed).
> 
> @Shine- please provide your SoB so that @systemcrash can add it to the commit. `Signed-off-by: Real Name <real@email>`

I'm aware of that.

But here comes my point: I don't want to see my real name and valid email address stamped into the commit history forever, so no _From:_ or _SoB:_ from me. I will still happily contribute, and my contributions are fully Public Domain unless noted otherwise. So - as happening here - whoever thinks my contributions are useful can re-PR or commit them under their name. I don't need or even want credit.

In addition, this was, I think, my very first contribution to OpenWrt. Nowadays, I don't attach ZIP files, I either add my examples to \<code\> blocks, so they're easily readable, or I actually raise a PR, with a note that I won't provide SoB, yet anybody can re-use the code freely and submit it under their own SoB.

src/router.c
Shine-: How large would you define the array? How would you make sure to only allocate as much memory as needed, not too much and not too few? That was my reason to use struct here.

But I'm not a Pro coder, perhaps nowadays there's better style to achieve what I wanted to achieve. My code can often seem 80s style, because that's where it comes from :-)

Noltari:
> But here comes my point: I don't want to see my real name and valid email address stamped into the commit history forever, so no _From:_ or _SoB:_ from me. I will still happily contribute, and my contributions are fully Public Domain unless noted otherwise. So - as happening here - whoever thinks my contributions are useful can re-PR or commit them under their name. I don't need or even want credit.

I get your point but it's still against the OpenWrt rules: see [Submitting patches](https://openwrt.org/submitting-patches#sign_your_work)

Shine-:
> > But here comes my point: I don't want to see my real name and valid email address stamped into the commit history forever, so no _From:_ or _SoB:_ from me. I will still happily contribute, and my contributions are fully Public Domain unless noted otherwise. So - as happening here - whoever thinks my contributions are useful can re-PR or commit them under their name. I don't need or even want credit.
> 
> I get your point but it's still against the OpenWrt rules: see [Submitting patches](https://openwrt.org/submitting-patches#sign_your_work)

Yes, fully aware of that.
I could also choose to not contribute at all, and keep my code private. But I don't, I choose to release my code into the Public Domain instead. I don't think there's anything wrong with that, or is there?

Noltari:
> Yes, fully aware of that. I could also choose to not contribute at all, and keep my code private. But I don't, I choose to release my code into the Public Domain instead. I don't think there's anything wrong with that, or is there?

You're right, there's nothing wrong with that.
But since you don't want to follow the rules I won't even look at your contributions. However, others can still look into them and prepare proper Pull Requests, which I will happily review and accept even if your SoB is missing.

src/router.c
Alphix: > How large would you define the array? How would you make sure to only allocate as much memory as needed, not too much and not too few?

By looping twice, there is an example above

src/router.c
Alphix: > Example 1: It doesn't matter who is actually operating Router A.

The reason I mentioned it is because if the same org operates router A and B, they are in charge of the policies of both.

> I'm confident that there is also some RFC that disallows such rogue behavior.

Excellent! That is exactly the kind of reference I wanted to see and would be a great way to lay this discussion to rest. Which RFC is it?

> Is any of you actually **using** DHCPv6/NDP relay, or are you all just talking theoretically, without having an actual use case?

That's not very constructive, is it?

Shine-: It's very constructive, it was a question whether you're looking at it from a theoretical point of view (your insisting on the corner case above suggests it), or you're actually coming from a use-case scenario of DHCPv6/NDP relay. No offence intended.

I'm not going to argue anymore. I made my point: Uplink network is usually not yours, you can't allocate IPv6 addresses from it in other ways than uplink's policy. And if uplink network _is_ yours (corner case - you wouldn't have to resort to things like NDP relay if you had control over uplink), you can configure your uplink network properly, instead of messing with the relayed RAs.

My proposal is in #172, it's explained in detail there, with complete reasoning, and I also explained it again here multiple times. That's it from me, my proposal stands and it's free to use for anyone, for any purpose.

Alphix:
@systemcrash perhaps you or I should do a new spin on this, and focus only on modifying the L/M/O flags?

systemcrash:
That we (downstream) can only disable the A flag if enabled (upstream) is reasonable; the implications for routing and/or DHCPv6 aren't immediately obvious, but are reasons for RA advertising a PIO, without the A flag.

IOW: "If the Autonomous flag is not set, silently ignore the PIO with regard to SLAAC." (but for other purposes, throw that PIO around).

Problem scenario. 

Router A no A bit, no L bit (this PR), downstream router B sets A bit. Router B starts forwarding traffic from B to A from that new SLAAC address, but A has only provided route info to B on how to reach that subnet, yet receives traffic *from* that subnet. 

TLDR: no A-bit, no L-bit == "You can reach this prefix through me, but it’s not part of your local subnet."
That's the implication, but behaviour may vary based on implementation.

src/router.c
systemcrash: I think for the loops and scaffolding, the extra work is not worth it for the few potential bytes of memory saved.

I contemplated an array of booleans instead, exclusively for the A flags, but it's not really worth it :/

Noltari: > I think for the loops and scaffolding, the extra work is not worth it for the few potential bytes of memory saved.
> 
> I contemplated an array of booleans instead, exclusively for the A flags, but it's not really worth it :/

But you can still add a new struct declaration:
```
struct fwd_pio_flags {
	uint8_t *ptr;
	uint8_t flags;
};
```
and use it there:
```
struct fwd_pio_flags *pio_flags = NULL, *pio_flags_new;
```

Alphix: > I think for the loops and scaffolding, the extra work is not worth it for the few potential bytes of memory saved.

It's not about saving bytes, it's about avoiding heap allocations when stack allocations are feasible. 

Doing repeated `realloc` calls each and every time a RA is fwded isn't pretty.
 
> I contemplated an array of booleans instead, exclusively for the A flags, but it's not really worth it :/

If you clear the L flag in the switch statement and leave A alone, you can simplify this a lot. That would be my suggestion ATM...

Alphix:
> That we (downstream) can only disable the A flag if enabled (upstream) is reasonable; the implications for routing and/or DHCPv6 aren't immediately obvious, but are reasons for RA advertising a PIO, without the A flag.

I think it's more reasonable to just leave A alone instead of honouring one value for `ra_slaac` and ignoring the other value...

systemcrash:
> > That we (downstream) can only disable the A flag if enabled (upstream) is reasonable; the implications for routing and/or DHCPv6 aren't immediately obvious, but are reasons for RA advertising a PIO, without the A flag.
> 
> I think it's more reasonable to just leave A alone instead of honouring one value for `ra_slaac` and ignoring the other value...

L makes no sense downstream. And PIOs have only those two bits defined, ya?

Alphix:
> > > That we (downstream) can only disable the A flag if enabled (upstream) is reasonable; the implications for routing and/or DHCPv6 aren't immediately obvious, but are reasons for RA advertising a PIO, without the A flag.
> > 
> > 
> > I think it's more reasonable to just leave A alone instead of honouring one value for `ra_slaac` and ignoring the other value...
> 
> L makes no sense downstream. And PIOs have only those two bits defined, ya?

LARP (X, probably won't go anywhere) and whatever the future might bring...

Alphix:
> L makes no sense downstream.

I agree with that... And removing it might fix some bugs. (And it can also be done early, once in the switch statement)

Shine-:
Not touching A in this PR makes sense.
Ie. focus only on bug fixing, not additional features.

Alphix:
> Not touching A in this PR makes sense. Ie. focus only on bug fixing, not additional features.

@Shine- I'm happy we agree (and I don't mean that ironically or sarcastically)

Shine-:
Thanks for your approval! Always feels better to agree with others than to disagree. :-)

After re-reading my original description in #172, I figured that messing with A wasn't originally included anyway - I raised #172 because I noticed the incorrect behavior regarding O and M. I added the additional bit with A to my patch "while I was at it", and later on, was quite happy that I could deal with shorter IPv6 addresses (PREFIX::1234 instead of PREFIX:1111:2222:3333:4444) if I set A=0 in my use case. ;-)

But in fact, the old behavior with M and O is the only actual bug (leaving L aside, that didn't matter to me, so I didn't touch it either). Being able to control A at downlink is just an additional gimmick. So the two parts should be handled separately, not mixed together within one PR or commit.

Alphix:
> Always feels better to agree with others than to disagree. :-)

Absolutely :)

> I added the additional bit with A to my patch "while I was at it", and later on, was quite happy that I could deal with shorter IPv6 addresses (PREFIX::1234 instead of PREFIX:1111:2222:3333:4444) if I set A=0 in my use case. ;-)

You can still have that by using [IPv6 Tokens](https://datatracker.ietf.org/doc/html/draft-chown-6man-tokenised-ipv6-identifiers-02), which will give you short and memorable IPv6 addresses even with SLAAC.

There's support for it in e.g.:
 - [the Linux kernel](https://man7.org/linux/man-pages/man8/ip-token.8.html)
 - [NetworkManager](https://networkmanager.dev/docs/api/latest/settings-ipv6.html) (see `token` at the end of the page)
 - [systemd-networkd](https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#%5BIPv6AcceptRA%5D%20Section%20Options) (see `Token`)
 - and other software as well

It's what I use to get predictable `ULA` and `GUA` addresses for e.g. servers and other things where memorable addresses are desirable, and dynamic IPv6 prefixes are used (boo).

Shine-:
> > I added the additional bit with A to my patch "while I was at it", and later on, was quite happy that I could deal with shorter IPv6 addresses (PREFIX::1234 instead of PREFIX:1111:2222:3333:4444) if I set A=0 in my use case. ;-)
> 
> You can still have that by using [IPv6 Tokens](https://datatracker.ietf.org/doc/html/draft-chown-6man-tokenised-ipv6-identifiers-02), which will give you short and memorable IPv6 addresses even with SLAAC.
> 
> There's support for it in e.g.:
> 
> * [the Linux kernel](https://man7.org/linux/man-pages/man8/ip-token.8.html)
> * [NetworkManager](https://networkmanager.dev/docs/api/latest/settings-ipv6.html) (see `token` at the end of the page)
> * [systemd-networkd](https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#%5BIPv6AcceptRA%5D%20Section%20Options) (see `Token`)
> * and other software as well
> 
> It's what I use to get predictable `ULA` and `GUA` addresses for e.g. servers and other things where memorable addresses are desirable, and dynamic IPv6 prefixes are used (boo).

I have, in this particular network (guest network, so I have the freedom to play around as much as I want), uplink that offers DHCPv6 with the last 16 bits, and SLAAC, which is 64 bits by definition. So I can relay just SLAAC (running into the M/O problems with odhcpd) or relay both, but then every client will get >= 2 IPv6 addresses (DHCPv6, SLAAC, temporary). Setting A=0 **in my use case** allowed me to effectively relay just DHCPv6. Now every client has only one IPv6 address. That's, in a nutshell, my use case, for which I created my patch.

Edit: I just checked, currently I'm using SLAAC-only there, so not even I am using the A=0 bit anymore that my patch allows ;-)

systemcrash:
Soo... @Shine- this patch is a hybrid to form a good solution. It compiles, but I have not tested it. I wonder whether you can take it for a test-drive on current master to verify its usefulness.

Shine-:
I knew this would come ;-)

I can, but I'll be away for this weekend and maybe a few days more, so it can take a bit longer until I can even start.

The network where I'm using Relay in production is currently very low-frequented, so memleaks or similar would only be apparent in due time - ie. not a good test subject. But I can build a test infrastructure where I also have control over uplink, and can perform some behavior testing there, including Wireshark logging. Just like I originally tested mine.

Will take some time, though, so I'd appreciate if others could join testing.

Since this was already waiting for >4 years, it wouldn't be a biggie if it has to wait for another few weeks until it's ready to merge, would it?

Alphix:
> Soo... @Shine- this patch is a hybrid to form a good solution. It compiles, but I have not tested it. I wonder whether you can take it for a test-drive on current master to verify its usefulness.

@systemcrash does that mean you disagree with removing the `A` modification functionality?

systemcrash:
Yeah, disabling A is useful here. The patch *works* :)

Edit: I think it gives us a good structure for other bits. The P bit is interesting, which we've been musing about in an issue thread. 

Maybe I can add that here too, so we advertise the P flag when DHCPv6 is doing its thing (serv/relay).

Alphix:
> Yeah, disabling A is useful here. The patch _works_ :)

At the risk of being annoying...I still disagree...I think it's a confusing footgun.

> Edit: I think it gives us a good structure for other bits. The P bit is interesting, which we've been musing about in an issue thread.
>
> Maybe I can add that here too, so we advertise the P flag when DHCPv6 is doing its thing (serv/relay).

`P` can't be unconditionally enabled just because we're doing DHCPv6, it's a policy decision which needs explicit cfg. I've been thinking about the `P` flag (and things like static routes, and the `prefix_filter` option)...I have some kind of rough idea in mind, I think we might want to add a new option down the line...something like `list prefix_config` which takes `<prefix> <options>`.

That way, `<options>` can be e.g.:
- `ignore` (superseding `prefix_filter`, addresses https://github.com/openwrt/odhcpd/issues/243)
- `priority` (addresses https://github.com/openwrt/odhcpd/issues/105)
- `autonomous` (`A` flag, addresses https://github.com/openwrt/odhcpd/issues/59)
- `pdpref` (`P` flag, addresses https://github.com/openwrt/odhcpd/issues/253)
- `static` (addresses https://github.com/openwrt/odhcpd/issues/74)
- `onlink`
- etc

I think you get my drift...

Edit: and `<prefix>` could be treated like a mask, so any prefix falling within the scope of `prefix` gets matched...giving us the possibility to also work with dynamic/unpredictable prefixes...so you could e.g. match on your ISPs whole prefix...

systemcrash:
Updated: 
-disable the P flag on RA forward if dhcpv6_pd is disabled
-send the P flag on RA send if dhcpv6_pd is enabled (it is by default per interface, can be disabled) and dhcpv6 is not disabled.

Fully tested and working in both RA send and RA forward scenarios.

Edit: I chained it to the existing `dhcpv6_pd` boolean.

systemcrash:
> At the risk of being annoying...I still disagree...I think it's a confusing footgun.

It may be confusing, but it's no footgun. Once one understand the logic, it's fine. Thus ample code comments :)

> `P` can't be unconditionally enabled just because we're doing DHCPv6...

`P` can be unconditionally enabled - it's an indicator that the admin prefer the downstream clients use DHCPv6-PD. The clients can certainly request, and try, or, altogether ignore the P flag. ( see my choice commit message comments or RFC :)

Just as `A` is today: you can enable SLAAC, whether or not you have any prefixes. Same thing. 

I don't disagree that changing the `dhcpv6_pd` and thus `P` could be chained to availability of prefixes or other conditions. But this is a big on/off switch right now.

systemcrash:
> something like `list prefix_config` which takes `<prefix> <options>`.

I think this would be great. There is The oldest PR in this repo which had a similar idea :)

Alternatively a `(ipv6)prefix` config entry with its various options as parameters? e.g.

```
# /etc/config/dhcp
config ipv6prefix
        option flags '...'
        option ignore '0'
        option prefix '.../64'
```

Alphix:
> Updated: -disable the P flag on RA forward if dhcpv6_pd is disabled -send the P flag on RA send if dhcpv6_pd is enabled (it is by default per interface, can be disabled) and dhcpv6 is not disabled.
> 
> Fully tested and working in both RA send and RA forward scenarios.
> 
> Edit: I chained it to the existing `dhcpv6_pd` boolean.

Um...`dhcpv6_pd` and the `P` flag are not related at all?

systemcrash:
> > Updated: -disable the P flag on RA forward if dhcpv6_pd is disabled -send the P flag on RA send if dhcpv6_pd is enabled (it is by default per interface, can be disabled) and dhcpv6 is not disabled.
> > Fully tested and working in both RA send and RA forward scenarios.
> > Edit: I chained it to the existing `dhcpv6_pd` boolean.
> 
> Um...`dhcpv6_pd` and the `P` flag are not related at all?

RFC9762 Use RA to Signal DHCPv6 PD ... [to] use the deployment model in RFC 9663

RFC9663: ... individual nodes connected to large broadcast networks are allocated unique prefixes via DHCPv6 Prefix Delegation (DHCPv6-PD), as specified in RFC 8415.

RFC8415: DHCPv6-PD ( https://www.rfc-editor.org/rfc/rfc8415.html#section-6.3 )

From the readme:

| Option		| Type	|Default| Description |
| :-------------------- | :---- | :---- | :---------- |
| dhcpv6_pd		|bool	| 1	| DHCPv6 stateful addressing hands out IA_PD - Internet Address - Prefix Delegation |


Take the patch-set for a spin ;)

Alphix:
RFC9762, abstract:

> ...The flag is used to indicate that the network prefers that clients use the deployment model in RFC 9663 instead of using individual addresses in the on-link prefix assigned using Stateless Address Autoconfiguration (SLAAC) or DHCPv6 address assignment.
> 
> This document updates RFC 4862 to indicate that the Autonomous flag in a PIO needs to be ignored if the PIO has the P flag set...
> 

Keyword: ...the network **prefers** that clients use [DHCPv6-PD]...

The `dhcpv6_pd` option _enables_ PD support (when possible), the `P` flag is a different thing, it tells clients that PD is _preferred_ over other options. Simply enabling PD doesn't mean that it is preferred (or even possible).

Alphix:
> > something like `list prefix_config` which takes `<prefix> <options>`.
> 
> I think this would be great. There is The oldest PR in this repo which had a similar idea :)
> 
> Alternatively a `(ipv6)prefix` config entry with its various options as parameters? e.g.
> 
> ```
> # /etc/config/dhcp
> config ipv6prefix
>         option flags '...'
>         option ignore '0'
>         option prefix '.../64'
> ```

Yeah, I considered that as well...the advantage of `list prefix_config` (which could take both IPv4 and IPv6 prefixes) is that it makes it very clear which `interface` a given `prefix` cfg belongs to. With `config ipv6prefix`, there would also have to be some kind of `interface` option, and renaming the interface would lose the prefix cfg.

But these kind of things (which exact cfg format makes more sense) usually become clearer while actually writing the code, so I guess it's a case of TBD...

Alphix:
> > At the risk of being annoying...I still disagree...I think it's a confusing footgun.
> 
> It may be confusing, but it's no footgun. Once one understand the logic, it's fine. Thus ample code comments :)

Code comments doesn't really help users...honouring `ra_slaac` in one case and ignoring it in another case isn't exactly intuitive. The point of relaying NDP is essentially to give clients on two different network segments the impression that they're on the same network segment. Fiddling with `A` breaks that illusion, and I agree with what @Shine- said above "Being able to control A at downlink is just an additional gimmick. So the two parts should be handled separately, not mixed together within one PR or commit."

> > `P` can't be unconditionally enabled just because we're doing DHCPv6...
> 
> `P` can be unconditionally enabled - it's an indicator that the admin prefer the downstream clients use DHCPv6-PD. The clients can certainly request, and try, or, altogether ignore the P flag. ( see my choice commit message comments or RFC :)

No, `P` cannot, it is a policy decision that is orthogonal to whether `PD` is supported/enabled or not.

Just as an example: `P` tells clients to ignore `A` and to not try to get an `IA_NA` addr via DHCPv6. So if `dhcpv6_pd` is enabled, `P` is set and PD isn't possible, clients might not configure an address at all...(they _might_ fallback to `IA_NA` or `SLAAC`, but that's not a given).

The `P` flag can also create a situation (assuming e.g. a `::/60` prefix from upstream) where no more than 15 devices can be connected to the network.

> Just as `A` is today: you can enable SLAAC, whether or not you have any prefixes. Same thing.

I'm afraid I have to disagree, it's apples and oranges.

systemcrash:
> Keyword: ...the network **prefers** that clients use [DHCPv6-PD]...
> 
> The `dhcpv6_pd` option _enables_ PD support (when possible), the `P` flag is a different thing, it tells clients that PD is _preferred_ over other options. Simply enabling PD doesn't mean that it is preferred (or even possible).

Well aware - that's the comment I put in the commit message.

I still have that other branch where I put this whole P thing behind another boolean to *prefer* this. 


> The `P` flag can also create a situation (assuming e.g. a `::/60` prefix from upstream) where no more than 15 devices can be connected to the network.

Yes, I recall. Get more space :)

Alphix:
> > Keyword: ...the network **prefers** that clients use [DHCPv6-PD]...
> > The `dhcpv6_pd` option _enables_ PD support (when possible), the `P` flag is a different thing, it tells clients that PD is _preferred_ over other options. Simply enabling PD doesn't mean that it is preferred (or even possible).
> 
> Well aware - that's the comment I put in the commit message.

Your commit message says:

> This commit sets the RA PIO P flag if dhcpv6_pd is enabled, thereby signalling
to downstream clients that DHCPv6-PD is available. Whether they prefer it is up
to them.

And that is _not_ what the `P` flag means. It does not mean that PD is available.

You also mentioned earlier that `dhcpv6_pd` means we can unconditionally enable the `P` flag. I strongly disagree.

> > The `P` flag can also create a situation (assuming e.g. a `::/60` prefix from upstream) where no more than 15 devices can be connected to the network.
> 
> Yes, I recall. Get more space :)

Eh...I don't know what to say to that...

Alphix:
@systemcrash I suggest you split this PR, keep this one to fiddling with M/O/L flags, and move your A/P logic to a different PR. It'll make the code changes in this PR trivial and uncontroversial. And I think the M/O/L stuff could be quite useful for some users (I'm especially curious to see how the L changes would fare in the field).

systemcrash:
👌

Noltari (approved):
LGTM
@Alphix any objections since @systemcrash has dropped the A/P logic?

src/router.c
```diff
@@ -1157,6 +1191,26 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t
 		if (mac_ptr)
 			odhcpd_get_mac(c, mac_ptr);
 
+		if (pio_count > 0)
+			debug("RA forward: Rewriting RA PIO flags");
+
+		for (size_t i = 0; i < pio_count; i++) {
+			/* restore the flags byte to its upstream state before applying per-interface policy */
+			*pio_flags[i].ptr = pio_flags[i].flags;
+			/* ensure L flag (on-link) cleared; relayed == no longer on link, so ND is bad */
```
Alphix: This comment sounds a bit funny...it's not just that "ND is bad" (downlink hosts might not reach other hosts on this prefix without going via the router)...suggest just truncating it to `...; relayed == not on-link */`

Alphix:
Other than that, shoot

src/router.c
systemcrash: K. Done.

Noltari:
Merged, thanks @systemcrash!

PR #301, opened 2025-11-08
odhcpd: Implement RFC9762 DHCPv6 PD Preferred flag for PIOs

systemcrash:
The RFC defines the P flag for use within PIOs, and defines its meaning as "purely a positive indicator, telling nodes that DHCPv6 PD is available and the network prefers that nodes use it".

The RFC abstract lays out:

The (P) flag is used to indicate that the network prefers that clients use the deployment model in RFC 9663 instead of using individual addresses in the on-link prefix assigned using SLAAC, or DHCPv6 address assignment.

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -591,6 +591,15 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_LOW;
 	else if (iface->route_preference > 0)
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_HIGH;
+ 
+	if (iface->dhcpv6 != MODE_DISABLED && iface->dhcpv6_pd && iface->dhcpv6_pd_preferred) {
+		/* RFC9762 § 5
+		 * If the network desires to delegate prefixes to devices that support
+		 * DHCPv6 prefix delegation but do not support the P flag, it SHOULD 
+		 * also set the M or O bits in the RA to 1
+		 */
+		adv.h.nd_ra_flags_reserved |= ND_RA_FLAG_MANAGED;
+	}
 
 	adv.h.nd_ra_reachable = htonl(iface->ra_reachabletime);
 	adv.h.nd_ra_retransmit = htonl(iface->ra_retranstime);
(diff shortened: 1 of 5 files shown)


systemcrash:
ping @Alphix - this should satisfy our requirements

Edit: not per-prefix, but as a 'global' option, it's a start.

Alphix:
I don't want to be a PITA, but I disagree with this approach...this needs to be a per-prefix setting, so adding a global flag is unnecessary churn when we later implement a per-prefix flag (potentially along the lines of what I outlined [here](https://github.com/openwrt/odhcpd/pull/300#issuecomment-3503917190)) and once per-prefix _is_ implemented, it'd just be harder to remove the global option again.

The `P` flag is very recent (published June 2025) and doesn't really have widespread client support yet (I realise that this can change quickly). There's no rush in implementing it, especially not as a stopgap measure.

systemcrash:
But it *does* need a global on/off switch. P is becoming mandated in https://www.ietf.org/archive/id/draft-ietf-v6ops-rfc7084bis-04.html

This is implemented exactly as A and L flags today, which I presume we'd also want to handle similarly 'per-prefix'. So we're in no worse off position with this. In this P case, per-prefix is a refinement beyond what the majority of users need. 

We don't have a per-prefix RA route priority option yet. Which... wouldn't be bad. I think we should open a design discussion in an issue about how best to implement per-prefix stuff.

Alphix:
> But it _does_ need a global on/off switch. P is becoming mandated in https://www.ietf.org/archive/id/draft-ietf-v6ops-rfc7084bis-04.html

Yes, I'm not saying we don't need to provide the ability to set the `P` flag, not at all. Just that we should focus on the per-prefix part...as per [section 6 of RFC9762](https://www.rfc-editor.org/rfc/rfc9762.html#section-6)...

> This is implemented exactly as A and L flags today, which I presume we'd also want to handle similarly 'per-prefix'. So we're in no worse off position with this. In this P case, per-prefix is a refinement beyond what the majority of users need.

Yeah, I had a second look at the patch, and I must admit that I read it a bit too quickly at first. You're just setting the `P` flag for our "own" routes...and not making changes to relayed stuff...and this serves as a global default. I think this is ok, so 👍🏻 

> We don't have a per-prefix RA route priority option yet. Which... wouldn't be bad. I think we should open a design discussion in an issue about how best to implement per-prefix stuff.

I agree. We could create an issue to track that discussion. There's a lot of per-prefix stuff we could control with a separate option. I guess it would make the

src/router.c
```diff
@@ -591,6 +591,18 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_LOW;
 	else if (iface->route_preference > 0)
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_HIGH;
+ 
+ 	if ((iface->dhcpv6 != MODE_DISABLED && iface->dhcpv6_pd && iface->dhcpv6_pd_preferred) &&
+		!(iface->ra_flags & (ND_RA_FLAG_MANAGED | ND_RA_FLAG_OTHER))) {
```
Alphix: Wonky braces...I think this can be simplified to:

```C
if (iface->dhcpv6 != MODE_DISABLED && iface->dhcpv6_pd && iface->dhcpv6_pd_preferred) {
    /* ... comments ... */
    adv.h.nd_ra_flags_reserved |= ND_RA_FLAG_MANAGED;
}
```

I know the RFC says `M` or `O`, but semantically `M` makes more sense...and testing if a flag is set or not isn't really necessary...just set `M`

Alphix:
And some `README.md` changes are missing :)

src/router.c
systemcrash: Agreed about M. The rich list for those flags in the gui specifically mentions dhcpv6 for the M flag I think.

systemcrash:
OK. I've defaulted the P flag to off. Maybe on is better? 

Ready for @Noltari review.

Alphix:
> OK. I've defaulted the P flag to off. Maybe on is better?

Personally, I think `off` is the saner (and safer) default. So that's good.

Edit: and it's also the default that's recommended by the RFC

src/router.c
```diff
@@ -591,6 +591,15 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_LOW;
 	else if (iface->route_preference > 0)
 		adv.h.nd_ra_flags_reserved |= ND_RA_PREF_HIGH;
+ 
+ 	if ((iface->dhcpv6 != MODE_DISABLED && iface->dhcpv6_pd && iface->dhcpv6_pd_preferred)) {
```
Alphix: Nit: remove the double parens `((...))`, they don't add anything

Noltari: Fixed this NIT myself and a couple whitespace + tab NITs.

Noltari:
Merged, thanks @systemcrash!

Issue #237, opened 2025-04-24
Wishlist: Support multiple instances and/or separate leasefiles


pabl0:
OpenWRT allows for the operation of multiple instances of dnsmasq across different subnets. This feature is useful in small SOHO networks, such as when you want to set up a guest network that is fully isolated from other subnets.

Dnsmasq conceals dynamic names created by DHCP from other instances. However, for DHCPv6, all entries are stored in a single lease file (default is `/tmp/hosts/odhcpd`). If you do not set `dhcp.INSTANCE_dns.ignore_hosts_dir='1'`, those names can be seen by other instances. Conversely, enabling "ignore hosts directory" means you will also lose the ability to query AAAA names on the same subnet as well. In typical "guest" network scenarios, this may not pose a significant issue since users often only access the public Internet, but there might be use-cases where this would be a nice feature. And not all networks are dual-stack anymore...

Maybe a single odhcpd instance could handle this by just writing the leases to separate `/tmp/hosts/odhcpd.INSTANCE` files, if it knows about dnsmasq instances?

Alphix:
Multiple instances is quite contrary to how odhcpd is designed but multiple per-interface hosts files should be quite easy to implement. 

Need to do some refactoring to the state/hostsfile logic first though. And we need to sort out the interaction with dnsmasq...

Alphix:
@pabl0 see PR #302 which implements per-interface hosts (not lease) files. If you have the time and ability to test the PR, that would be great :)

PR #302, opened 2025-11-09
odhcpd: statefile and hostsfile improvements

Alphix:
The gist is that state and hosts file writing is moved from `src/dhcpv6-ia.c` into a separate file `src/statefiles.c` (since state and hosts file writing is something which is used by both the DHCPv4 and DHCPv6 servers). This also makes `src/dhcpv6-ia.c` a bit easier to work with (shrinking it from ~1700 lines to ~1300 lines, my rule of thumb is that files start getting difficult to grok when they exceed 1000 lines).

Anyway, there are some important fixes here:
 - Right now, the `hostsfile` cfg option has no effect if `leasefile` isn't set (not documented, but mentioned in 4bbc6e74248feeb756bd03dc500fb4f446ccfc49)
 - If using multiple subnets and instances of `dnsmasq`, there is no way to tell which hosts are available on which subnet, meaning that `dnsmasq` will "leak" the hostnames across subnets (so your guest network can see hosts from other networks, see issue #237
 - With that in mind, the `hostsfile` option is changed to `hostsdir`, where per-interface hosts files will be written

With regard to the `hostsdir` change, note that while `odhcpd` can write a hostsfile as well as a leasefile, the former is typically not used at the moment in a default install. `dnsmasq` will instead read the `leasefile` to get the DNS records.

After this change, `dnsmasq` can be migrated to use the `hostsdir` instead.

dnsmasq already supports reading several hosts files from a directory (via the `--hostsdir=` option), so some changes to dnsmasq's init script will also be necessary (depending on whether the user wants dnsmasq to support resolution of all hosts or just the hosts on a given interface).
    
Making hosts files (rather than odhcpd's lease/state file) the primary interface for dnsmasq is also an important first step towards making the state/leasefile properly private (the other important consumer is LuCI, via its rpcd plugin, I already have some prototype patches to let LuCI get the information by talking to `odhpcd` over ubus instead of parsing its state files), which is necessary if we want to support persisting leases across restarts of odhcpd (or even reboots, if writing to flash is deemed acceptable) in the future (since I want to extend the state/leasefile to capture the full state of leases so that they can be properly reconstructed on reboot/restart).

In addition, the use of `--hostsdir=` in dnsmasq could make the `leasetrigger` unnecessary in the future, since the latter restarts `dnsmasq` to make it aware of the changes to the hostsfile (and `hostsdir` means that `dnsmasq` will monitor the dir with `inotify` and automagically pickup any changes). That still leaves the case when dnsmasq uses multiple instances and only wants to read a specific hostsfile (which, AFAIK, doesn't support automatic detection of when that specific file changes), but that's something to address down the road (the `leasescript` functionality is still there with this PR applied).

There's also a substantial amount of cleanup work to make the statefile code easier to read and parse. There's a log of patches here, and I would suggest that you read the whole diff if you want to save some time, since it'll show the resulting statesfile.c quite nicely.

### DIFF
--- a/src/dhcpv6-ia.h
+++ b/src/dhcpv6-ia.h
@@ -1,0 +1,37 @@
+/**
+ * Copyright (C) 2013 Steven Barth <steven@midlink.org>
+ * Copyright (C) 2016 Hans Dedecker <dedeckeh@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License v2 as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ */
+
+#ifndef _DHCPV6_IA_H_
+#define _DHCPV6_IA_H_
+
+#define ADDR_ENTRY_VALID_IA_ADDR(iface, i, m, addrs) \
+    ((iface)->dhcpv6_assignall || (i) == (m) || \
+     (addrs)[(i)].prefix > 64)
+
+size_t get_preferred_addr(const struct odhcpd_ipaddr *addrs, const size_t addrlen);
+
+struct in6_addr in6_from_prefix_and_iid(const struct odhcpd_ipaddr *prefix, uint64_t iid);
+
+static inline bool valid_prefix_length(const struct dhcpv6_lease *a, const uint8_t prefix_length)
+{
+	return a->length > prefix_length;
+}
+
+static inline bool valid_addr(const struct odhcpd_ipaddr *addr, time_t now)
+{
+	return (addr->prefix <= 96 && addr->valid_lt > (uint32_t)now && addr->preferred_lt > (uint32_t)now);
+}
+
+#endif /* _DHCPV6_IA_H_ */
--- a/src/odhcpd.c
+++ b/src/odhcpd.c
@@ -42,7 +42,9 @@
 #include <sys/random.h>
 
 #include <libubox/uloop.h>
+
 #include "odhcpd.h"
+#include "dhcpv6-ia.h"
 
 static int ioctl_sock = -1;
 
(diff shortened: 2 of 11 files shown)


src/statefiles.c
```diff
@@ -0,0 +1,350 @@
...
+		ctxt->iface->ifname, hexhwaddr,
+		(lease->flags & OAF_BROKEN_HOSTNAME) ? "broken\\x20" : "",
+		(lease->hostname ? lease->hostname : "-"),
+		(lease->valid_until > ctxt->now ?
+		 (int64_t)(lease->valid_until - ctxt->now + ctxt->wall_time) :
+		 (INFINITE_VALID(lease->valid_until) ? -1 : 0)),
+		ntohl(lease->addr), ipbuf);
+}
+
+/* Returns true if there's changes that should be written to the hosts file(s) */
```
systemcrash: 'if there are changes to be written' :)

src/odhcpd.c
```diff
@@ -663,6 +665,63 @@ void odhcpd_bmemcpy(void *av, const void *bv, size_t bits)
...
+			continue;
+
+		/* Filter Out Prefixes */
+		if (ADDR_MATCH_PIO_FILTER(&addrs[i], iface)) {
+			char addrbuf[INET6_ADDRSTRLEN];
+			info("Address %s filtered out on %s",
+			     inet_ntop(AF_INET6, &addrs[i].addr.in6, addrbuf, sizeof(addrbuf)),
+			     iface->name);
+			continue;
+		}
```
systemcrash, line 689: This chunk also exists elsewhere. It seems to prevent handing out PIO if any part of the public prefix we received is assigned to a local interface. I've been attempting RFC6603. This can go in, but may need changing to support 6603.

src/dhcpv6-ia.h
```diff
@@ -0,0 +1,34 @@
...
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ */
+
+#pragma once
```
systemcrash: I think Noltari pulled these out and opted for `#ifndef _dhcpv6_ia_h_` kinda thing.

Alphix: I'm also more used to include guards, but...

```shell
$ fgrep -r pragma *.h
dhcpv4.h:#pragma once
dhcpv6.h:#pragma once
dhcpv6-pxe.h:#pragma once
odhcpd.h:#pragma once
router.h:#pragma once
```

🤷🏻‍♂️

src/odhcpd.c
Alphix: It's not something I've changed in this PR, that's just moved around.

But it checks against `iface->pio_filter_addr`, which is set from `IFACE_ATTR_PREFIX_FILTER`, i.e. config option `prefix_filter`, so it's a check against the user-defined filter...

src/statefiles.c
Alphix: Oki

src/dhcpv6-ia.h
systemcrash: I think if we're in the business of getting rid of "sort of supported sometimes" things for standardised things, we can fix this one now. Maybe @Noltari has an opinion on that.

Noltari: > I think if we're in the business of getting rid of "sort of supported sometimes" things for standardised things, we can fix this one now. Maybe @Noltari has an opinion on that.

@systemcrash fixed at https://github.com/openwrt/odhcpd/pull/304

Alphix: And in this PR as well now

Alphix:
Beside the code review, I should point out again that this will require some careful handling of UCI defaults in the `openwrt/openwrt` repo, [odhcpd.defaults](https://github.com/openwrt/openwrt/blob/main/package/network/services/odhcpd/files/odhcpd.defaults) has these lines as default:

```
set dhcp.odhcpd.leasefile=/tmp/hosts/odhcpd
set dhcp.odhcpd.leasetrigger=/usr/sbin/odhcpd-update
```

With no `dhcp.odhcpd.hostsfile` set. So after this PR, the defaults would change to something like this:

```
set dhcp.odhcpd.leasefile=/tmp/odhcpd.leases
set dhcp.odhcpd.leasetrigger=/usr/sbin/odhcpd-update
set dhcp.odhcpd.hostsdir=/tmp/hosts/
```

Optionally we could omit `leasetrigger` if we expect the few (?) people who use `dnsmasq` instances to set it manually (if I'm correct in that `dnsmasq` won't monitor changes to single files, but only directories)...most people shouldn't need a `leasetrigger` anymore.

And I think we might need to update existing cfgs along the lines of:
```shell
local hostsdir="/tmp/hosts/"

if [ -n "$(uci -q get dhcp.odhcpd.hostsfile)" ]; then
    hostsdir="$(dirname $(uci -q get dhcp.odhcpd.hostsfile))"
    uci delete dhcp.odhcpd.hostsfile
fi

if [ "$(uci -q get dhcp.odhcpd.leasefile)" = "/tmp/hosts/odhcpd" ]; then
    uci set dhcp.odhcpd.leasefile=/tmp/odhcpd.leases
fi

if [ -n "$(uci -q get dhcp.odhcpd.hostsdir)" ]; then
    uci set "dhcp.odhcpd.hostsdir=${hostsdir}"
fi

```

Noltari (approved):

systemcrash:
Yeah, it's fine. Looks elegant. 

@Alphix the leasetrigger thing fires every time I reload or restart odhcpd, and it spams my logs (once for each instance). I figure leave it, but that doesn't affect this PR.

```
daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.cfg01411c - 13 names
daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.cfg01411c - 18 names
daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.light - 18 names
daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.light - 13 names
```

@Noltari LGTM.

Alphix:
> Yeah, it's fine. Looks elegant.

Danke. Merci. Gracias. Etc.

> @Alphix the leasetrigger thing fires every time I reload or restart odhcpd, and it spams my 
logs (once for each instance). I figure leave it, but that doesn't affect this PR.

Yeah, long term I hope we can get rid of it...but not yet.

> ```
> daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.cfg01411c - 13 names
> daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.cfg01411c - 18 names
> daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.light - 18 names
> daemon.info dnsmasq[1]: read /tmp/hosts/dhcp.light - 13 names
> ```

Is `dhcp.light` your `statefile` or a file created via `hostsdir`?

systemcrash:
dhcp.cfg01411c = default instance, light = second named instance. Since host entries are not bound to a specific instance. 

Make a couple of dnsmasq instances and you'll see how it behaves.

Noltari:
Merged, thanks @Alphix!

PR #303, opened 2025-11-10
odhcpd: add support for the "ignore" option

### LINKED ISSUE #198: ip=ignore is ignored

Alphix:
This is inspired by @Kasoo's work in PR https://github.com/openwrt/odhcpd/pull/234, but slightly different.

The "ip" option now takes "ignore" as a value , but it'll only disable DHCPv4 for a matching client.

Similarly, the "hostid" option now also understands "ignore", which will disable DHCPv6 for a matching client.

Of course, this is all based on client-reported MAC addresses/DUIDs/client IDs/etc, so it's not exactly a security feature.

Closes: https://github.com/openwrt/odhcpd/pull/234
Closes: https://github.com/openwrt/odhcpd/issues/198

Draft for now, since it's based on top of #299, which needs to go in first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,10 +128,10 @@ and may also receive information from ubus
 ### Sections of type host (static leases)
 | Option		| Type	|Default| Description |
 | :-------------------- | :---- | :---- | :---------- |
-| ip			|string	|(none) | IPv4 host address |
+| ip			|string	|(none) | IPv4 host address or `ignore` to ignore any DHCPv4 request from this host |
 | mac			|list\|string|(none) | HexadecimalMACaddress(es) |
 | duid			|list\|string|(none) | Hexadecimal DUID(s), or DUID%IAID(s) |
-| hostid		|string	|(none)	| IPv6hostidentifier |
+| hostid		|string	|(none)	| IPv6 tokenised IID or `ignore` to ignore any DHCPv6 request from this host |
 | name			|string	|(none) | Hostname |
 | leasetime		|string	|(none) | DHCPv4/v6leasetime |
 
--- a/src/config.c
+++ b/src/config.c
@@ -670,15 +670,26 @@ int config_set_lease_cfg_from_blobmsg(struct blob_attr *ba)
 			goto err;
 	}
 
-	if ((c = tb[LEASE_CFG_ATTR_IP]))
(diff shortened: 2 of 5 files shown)


Alphix:
@systemcrash and @Noltari ^

systemcrash:
Ah, there. Conflicts. I figured it was familiar, that other commit. Rebase for review plz :)

Alphix:
Rebased, should be much easier to review now.

I should add that there's one small snafu here. DHCPv6 will just ignore the client if `hostid=ignore` (and not reply at all), while DHCPv4 will send a NAK to the client if `ip=ignore`, but I wanted to keep the code changes as minimal as possible. This should already be enough for most people's needs (I guess).

Also, this approach means that we could support further directives in the future (e.g. support both "reject" and "ignore", if that's something people have a genuine use-case for).

systemcrash (approved):

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #306, opened 2025-11-12
odhcpd: misc fixes

Alphix:
This PR includes a bunch of smallish commits that I felt didn't really warrant a PR per commit.

Use an AVL to store leases, rename some struct members to more descriptive names, and update the ubus method/broadcast messages to include additional information that we didn't include before (like DUID/IAID/interface for DHCPv4 leases)

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -195,7 +195,7 @@ const struct uci_blob_param_list interface_attr_list = {
 };
 
 const struct blobmsg_policy lease_cfg_attrs[LEASE_CFG_ATTR_MAX] = {
-	[LEASE_CFG_ATTR_IP] = { .name = "ip", .type = BLOBMSG_TYPE_STRING },
+	[LEASE_CFG_ATTR_IPV4] = { .name = "ip", .type = BLOBMSG_TYPE_STRING },
 	[LEASE_CFG_ATTR_MAC] = { .name = "mac", .type = BLOBMSG_TYPE_ARRAY },
 	[LEASE_CFG_ATTR_DUID] = { .name = "duid", .type = BLOBMSG_TYPE_ARRAY },
 	[LEASE_CFG_ATTR_HOSTID] = { .name = "hostid", .type = BLOBMSG_TYPE_STRING },
@@ -670,12 +670,12 @@ int config_set_lease_cfg_from_blobmsg(struct blob_attr *ba)
 			goto err;
 	}
 
-	if ((c = tb[LEASE_CFG_ATTR_IP])) {
+	if ((c = tb[LEASE_CFG_ATTR_IPV4])) {
 		const char *ip = blobmsg_get_string(c);
 
 		if (!strcmp(ip, "ignore"))
 			lease_cfg->ignore4 = true;
-		else if (inet_pton(AF_INET, blobmsg_get_string(c), &lease_cfg->ipaddr) < 0)
+		else if (inet_pton(AF_INET, blobmsg_get_string(c), &lease_cfg->ipv4) != 1)
 			goto err;
 	}
 
@@ -691,7 +691,7 @@ int config_set_lease_cfg_from_blobmsg(struct blob_attr *ba)
 				goto err;
 		}
 	} else {
-		uint32_t i4a = ntohl(lease_cfg->ipaddr) & 0xff;
+		uint32_t i4a = ntohl(lease_cfg->ipv4.s_addr) & 0xff;
 		lease_cfg->hostid = ((i4a / 100) << 8) | (((i4a % 100) / 10) << 4) | (i4a % 10);
 	}
 
@@ -1791,8 +1791,8 @@ static void lease_cfg_change(struct lease_cfg *lease_cfg_old, struct lease_cfg *
 		update = true;
 	}
 
-	if (lease_cfg_old->ipaddr != lease_cfg_new->ipaddr) {
-		lease_cfg_old->ipaddr = lease_cfg_new->ipaddr;
+	if (lease_cfg_old->ipv4.s_addr != lease_cfg_new->ipv4.s_addr) {
+		lease_cfg_old->ipv4 = lease_cfg_new->ipv4;
 		dhcpv4_free_lease(lease_cfg_old->dhcpv4_lease);
 	}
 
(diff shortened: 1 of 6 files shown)


Alphix:
@Noltari and @systemcrash ^

src/config.c
```diff
@@ -670,12 +670,12 @@ int config_set_lease_cfg_from_blobmsg(struct blob_attr *ba)
...
 	}
 
-	if ((c = tb[LEASE_CFG_ATTR_IP])) {
+	if ((c = tb[LEASE_CFG_ATTR_IPV4])) {
 		const char *ip = blobmsg_get_string(c);
 
 		if (!strcmp(ip, "ignore"))
 			lease_cfg->ignore4 = true;
-		else if (inet_pton(AF_INET, blobmsg_get_string(c), &lease_cfg->ipaddr) < 0)
+		else if (inet_pton(AF_INET, blobmsg_get_string(c), &lease_cfg->ipv4) != 1)
```
systemcrash, line 678: Just noting that this is a behaviour change, although encompassing 0 also is a good thing here. 

https://www.man7.org/linux/man-pages/man3/inet_pton.3.html#RETURN_VALUE

src/config.c
```diff
@@ -1069,6 +1069,11 @@ static int parse_dnr_str(char *str, struct interface *iface)
 	return -1;
 }
 
+static int avl_ipv4_cmp(const void *k1, const void *k2, _unused void *ptr)
+{
+	return memcmp(k1, k2, sizeof(struct in_addr));
+}
+
```
systemcrash, line 1076: I have a patch lined up with something similar for IPv6. This is good.

src/odhcpd.h
```diff
@@ -278,23 +278,23 @@ struct dhcpv6_lease {
...
 
 	unsigned int flags;
 	uint32_t leasetime;
 	char *hostname;
 
-	uint16_t clid_len;
-	uint8_t clid_data[];
+	uint32_t iaid;
+	uint16_t duid_len;
+	uint8_t duid[];
```
systemcrash, line 289: I think a suitable comment that this is (of course) the Client ID (what to look for in the RFC). 

Are there any dhcpv6_lease structs inited in the code that don't know about this re-alignment (iaid)?

src/dhcpv4.c
```diff
@@ -272,12 +272,19 @@ static void dhcpv4_fr_send(struct dhcpv4_lease *lease)
...
-	else
+		      odhcpd_print_mac(lease->hwaddr, sizeof(lease->hwaddr)),
+		      inet_ntop(AF_INET, &dest, ipv4_str, sizeof(ipv4_str)));
+	} else {
+		char ipv4_str[INET_ADDRSTRLEN];
+
 		debug("Sent %s to %s - %s", dhcpv4_msg_to_string(fr_msg.data),
-		      odhcpd_print_mac(lease->hwaddr, sizeof(lease->hwaddr)), inet_ntoa(dest.sin_addr));
+		      odhcpd_print_mac(lease->hwaddr, sizeof(lease->hwaddr)),
+		      inet_ntop(AF_INET, &dest, ipv4_str, sizeof(ipv4_str)));
```
systemcrash, line 286: Perhaps an indent

src/dhcpv4.c
```diff
@@ -272,12 +272,19 @@ static void dhcpv4_fr_send(struct dhcpv4_lease *lease)
...
 	if (dhcpv4_send_reply(iov, ARRAY_SIZE(iov), (struct sockaddr *)&dest, sizeof(dest),
-			      &lease->iface->dhcpv4_event.uloop.fd) < 0)
+			      &lease->iface->dhcpv4_event.uloop.fd) < 0) {
+		char ipv4_str[INET_ADDRSTRLEN];
+
 		error("Failed to send %s to %s - %s: %m", dhcpv4_msg_to_string(fr_msg.data),
-		      odhcpd_print_mac(lease->hwaddr, sizeof(lease->hwaddr)), inet_ntoa(dest.sin_addr));
-	else
+		      odhcpd_print_mac(lease->hwaddr, sizeof(lease->hwaddr)),
+		      inet_ntop(AF_INET, &dest, ipv4_str, sizeof(ipv4_str)));
```
systemcrash, line 280: same, indent.

systemcrash (commented):
`: Commit subject line MUST start with '<package name>: ' (dhcpv4, ubus: update ubus DHCPv4 events/methods)` - maybe just `dhcpv4: ...`

src/config.c
Alphix: Yeah, there's no reason to not error out on a `0` return value (which also signals an error), so I'm fairly certain that this is just an oversight, which my patch fixes

src/odhcpd.h
Alphix: > I think a suitable comment that this is (of course) the Client ID (what to look for in the RFC).

Hm, I'm not convinced, the DUID is such a central concept in RFC8415(bis) that it has a whole section of its own (section §11), talking about CLID here would feel like a bit of apochrypha...

> Are there any dhcpv6_lease structs inited in the code that don't know about this re-alignment (iaid)?

No, that would only be relevant if they were init:ed without specifying the members, inited via something else than `memset(...'\0'...)` or generated from something like `memcpy()`...and there are no such cases...but it's good that you point that out

src/dhcpv4.c
Alphix: I think it is correctly indented? (if you use an editor where tab = 8 spaces)

src/dhcpv4.c
Alphix: Same, `insert witty comment about vim being the one and only editor here`

Alphix:
> `: Commit subject line MUST start with '<package name>: ' (dhcpv4, ubus: update ubus DHCPv4 events/methods)` - maybe just `dhcpv4: ...`

Yeah, reasonable, I almost had a heart attack at first when I realised that there were tests that were failing and wondering why I hadn't caught them...fixed now

src/dhcpv4.c
systemcrash: Edit: ballz. yeah, ignore XD

systemcrash (approved):

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #307, opened 2025-11-13
src: improve attributes

Noltari:
src: consolidate and improve fallthrough
src: improve attributes

CC @Alphix @systemcrash

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ target_compile_options(${PROJECT_NAME} PRIVATE -Werror=implicit-function-declara
 target_compile_options(${PROJECT_NAME} PRIVATE -Wformat)
 target_compile_options(${PROJECT_NAME} PRIVATE -Werror=format-security)
 target_compile_options(${PROJECT_NAME} PRIVATE -Werror=format-nonliteral)
+target_compile_options(${PROJECT_NAME} PRIVATE -Wimplicit-fallthrough=5)
 target_compile_options(${PROJECT_NAME} PRIVATE -Wno-unused-parameter)
 target_compile_options(${PROJECT_NAME} PRIVATE -Wmissing-declarations)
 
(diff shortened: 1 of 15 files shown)


Alphix:
Same comment here as for the other PR, I don't think libubox should expose _identifiers in it's public API...

Alphix:
Not a fan of the `_o_` prefix, `_` already gives a clear hint that this is internal and things like `_packed` are common enough in many code bases that people would instinctively know what it is...

Also, you seem to have removed a bunch of fallthrough statements?

Noltari:
> Not a fan of the `_o_` prefix, `_` already gives a clear hint that this is internal and things like `_packed` are common enough in many code bases that people would instinctively know what it is...

As stated by @systemcrash [here](https://github.com/openwrt/odhcp6c/pull/120#issuecomment-3529136237) -> https://github.com/openwrt/odhcp6c/commit/e5690c1f13ed3605ece68d870abb3c4a2f09d571
There's an `_unused` declaration on a PowerPC header, and since double underscores are forbidden it's the best workaround.

> Also, you seem to have removed a bunch of fallthrough statements?

Yes, those weren't needed since there was no code between the two cases.
I've enabled the fallthrough warning so any missing fallthrough would cause the compiler to complain.

Alphix:
> > Not a fan of the `_o_` prefix, `_` already gives a clear hint that this is internal and things like `_packed` are common enough in many code bases that people would instinctively know what it is...
> 
> As stated by @systemcrash [here](https://github.com/openwrt/odhcp6c/pull/120#issuecomment-3529136237) -> [openwrt/odhcp6c@e5690c1](https://github.com/openwrt/odhcp6c/commit/e5690c1f13ed3605ece68d870abb3c4a2f09d571) There's an `_unused` declaration on a PowerPC header, and since double underscores are forbidden it's the best workaround.
> 
> > Also, you seem to have removed a bunch of fallthrough statements?
> 
> Yes, those weren't needed since there was no code between the two cases. I've enabled the fallthrough warning so any missing fallthrough would cause the compiler to complain.

Ah, all good

PR #309, opened 2025-11-14
config: fix memleak during odhcpd reload

lvdefei:
- The memset in close_interface reset the pios pointer before it could be freed, causing a memory leak. Relocate the free call to clean_interface to ensure proper deallocation.

- Use realloc instead of malloc in config_load_ra_pio() This function may be called multiple times during odhcpd reload, and using malloc without freeing the previous allocation was causing memory leaks.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -352,6 +352,7 @@ static void clean_interface(struct interface *iface)
 		free(iface->dnr[i].svc);
 	}
 	free(iface->dnr);
+	free(iface->pios);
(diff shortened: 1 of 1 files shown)


Noltari (approved):
LGTM
@Alphix @systemcrash any objections?

systemcrash:
Yeah, good!

Alphix:
Looks good, but just gave it a cursory glance

Noltari:
Merged, thanks @lvdefei!

PR #312, opened 2025-11-16
odhcpd: ubus and uloop fixes

Alphix:
This all started with me getting frustrated that `odhcpd` would go into an unkillable (well, `kill -9` works, of course) loop when built with `ubus`. So one patch adds some developer convenience by allowing `ubus` to be disabled at runtime even if it is built in.

The other patch improves the `uloop` signal handling so that we don't end up in unkillable loop and don't do a bunch of unnecessary work.

### DIFF
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -536,7 +536,7 @@ int odhcpd_get_flags(const struct interface *iface);
 struct interface* odhcpd_get_interface_by_index(int ifindex);
 void odhcpd_urandom(void *data, size_t len);
 
-void odhcpd_run(void);
+int odhcpd_run(void);
 time_t odhcpd_time(void);
 ssize_t odhcpd_unhexlify(uint8_t *dst, size_t len, const char *src);
 void odhcpd_hexlify(char *dst, const uint8_t *src, size_t len);
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -65,6 +65,10 @@
 #define _o_unused __attribute__((unused))
 #endif /* _o_unused */
 
+#ifndef _o_noreturn
+#define _o_noreturn __attribute__((__noreturn__))
+#endif /* _o_noreturn */
+
 #define ALL_IPV6_NODES "ff02::1"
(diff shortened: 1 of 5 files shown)


Alphix:
And I know that there's a `__attribute__((__noreturn__))` hiding in here, I'll adapt that after #307 lands

@Noltari @systemcrash please have a look

src/odhcpd.h
```diff
@@ -571,11 +571,21 @@ static inline int ubus_init(void)
 	return 0;
 }
 
+static inline const char *ubus_get_ifname(const char *name)
+{
+	return NULL;
```
systemcrash, line 589: just NULL?

src/odhcpd.h
```diff
@@ -571,11 +571,21 @@ static inline int ubus_init(void)
...
+}
+
 static inline void ubus_apply_network(void)
 {
 	return;
 }
 
+static inline bool ubus_has_prefix(const char *name, const char *ifname)
+{
+	return false;
```
systemcrash, line 599: always false?

src/odhcpd.h
Noltari: > just NULL?

This is for the case that `WITH_UBUS` is false, so the values are correctly forced.

src/odhcpd.h
Noltari: > always false?

This is for the case that `WITH_UBUS` is false, so the values are correctly forced.

src/odhcpd.h
Alphix: Yeah, it's dummy functions for when ubus is compiled out, not for when it's disabled at runtime

Noltari (approved):

Noltari:
> And I know that there's a `__attribute__((__noreturn__))` hiding in here, I'll adapt that after #307 lands
> 
> @Noltari @systemcrash please have a look

#307 is now in, so please rebase this and I'll merge it.

Alphix:
> #307 is now in, so please rebase this and I'll merge it.

Done

Noltari:
Merged, thanks @Alphix!

PR #313, opened 2025-11-16
dhcpv4: simplify random address selection

Alphix:
Simplify the generation of random addresses, and make the code easier to follow, by using the same approach as in dhcpv6.

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -395,78 +395,91 @@ static bool dhcpv4_insert_lease(struct avl_tree *avl, struct dhcpv4_lease *lease
 		return false;
 }
 
+static bool dhcpv4_assign_random(struct interface *iface,
+				 struct dhcpv4_lease *lease)
+{
+	uint32_t pool_start = ntohl(iface->dhcpv4_start_ip.s_addr);
+	uint32_t pool_end = ntohl(iface->dhcpv4_end_ip.s_addr);
+	uint32_t pool_size = pool_end - pool_start + 1;
+	unsigned short xsubi[3];
+	uint32_t try;
+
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari @systemcrash please have a look

src/dhcpv4.c
```diff
@@ -395,78 +395,91 @@ static bool dhcpv4_insert_lease(struct avl_tree *avl, struct dhcpv4_lease *lease
...
+		if (try > pool_end)
+			try = pool_start;
+
+		in_try.s_addr = htonl(try);
+
+		if (config_find_lease_cfg_by_ipv4(in_try))
+			continue;
+
+		if (dhcpv4_insert_lease(&iface->dhcpv4_leases, lease, in_try))
+			return true;
```
systemcrash, line 424: Is it a bad idea to have `dhcpv4_insert_lease` implicitly handle `config_find_lease_cfg_by_ipv4`? So everything is just an insert attempt, now with your beautiful AVLs? :)

So, insert will tell us the pool is full or it's not and grants a lease.

`config_find_lease_cfg_by_ipv4` returns a lease_cfg but we never do anything with those.

Alphix: `dhcpv4_lease()` uses `config_find_lease_cfg_by_ipv4()` (and the returned `struct lease_cfg`) without necessarily wanting to insert any new leases...so I don't think that would work, unfortunately

Noltari (approved):
LGTM
@systemcrash any objections?

systemcrash:
LGTM

Noltari:
Merged, thanks @Alphix!

PR #314, opened 2025-11-17
dhcpv6: prevent network loop scenario

systemcrash:
It's possible for odhcpd to talk with its (odhcp6c) self in a network loop scenario with strange results. Prevent this by returning early.

ping @Alphix

### DIFF
--- a/src/dhcpv6.c
+++ b/src/dhcpv6.c
@@ -708,6 +708,12 @@ static void handle_client_request(void *addr, void *data, size_t len,
 		}
 	}
 
+	if (dest.serverid_length == clientid.len && 
+	    !memcmp(clientid.buf, dest.serverid_buf, dest.serverid_length)) {
+		/* Bail if we are in a network loop where we talk with ourself */
+		return;		
+	}
+
 	if (!IN6_IS_ADDR_MULTICAST((struct in6_addr *)dest_addr) && iov[IOV_NESTED].iov_len == 0 &&
 	    (hdr->msg_type == DHCPV6_MSG_REQUEST || hdr->msg_type == DHCPV6_MSG_RENEW ||
 	     hdr->msg_type == DHCPV6_MSG_RELEASE || hdr->msg_type == DHCPV6_MSG_DECLINE)) {


Alphix:
How did you manage to trigger that?

src/dhcpv6.c
```diff
@@ -708,6 +708,11 @@ static void handle_client_request(void *addr, void *data, size_t len,
 		}
 	}
 
+	if (memcmp(clientid.buf, dest.serverid_buf, dest.serverid_length) == 0) {
+		/* If we are in a network loop where we talk with ourself, bail */
```
Noltari:
```C
/* Bail if we are in a network loop where we talk with ourself */
```

systemcrash:
> How did you manage to trigger that?

VirtualBox and bridged interfaces. Was staring at pcaps going "wtf, why are servid and cliid the same?"

Alphix:
> > How did you manage to trigger that?
> 
> VirtualBox and bridged interfaces. Was staring at pcaps going "wtf, why are servid and cliid the same?"

I think you're actively trying to earn your nickname 😄

src/dhcpv6.c
```diff
@@ -708,6 +708,11 @@ static void handle_client_request(void *addr, void *data, size_t len,
 		}
 	}
 
+	if (memcmp(clientid.buf, dest.serverid_buf, dest.serverid_length) == 0) {
```
Alphix: This isn't enough, you have to check if the length of clientid and serverid are the same, otherwise the `if` statement is true when `clientid` starts with the same data as `serverid` but `clientid` is longer.

Also, small nit, but the idiomatic way to do `memcmp()` to test equality is usually `if (!memcmp(a, b, len))`

systemcrash: Fair. Altho I like == 0. It's clear. :)

systemcrash:
> I think you're actively trying to earn your nickname 😄

😉

src/dhcpv6.c
Alphix: I'll probably change `== 0` in some distant patch without even thinking about it :D

Alphix:
Still don't like `== 0`, but LGTM

src/dhcpv6.c
systemcrash: Cruella de Alphix :D

Noltari:
Merged, thanks @systemcrash!
@Alphix I changed it to `!memcmp` and fixed the indentation ;)

PR #315, opened 2025-11-17
all: implement RFC8910 captive portal (CP) option

systemcrash:
https://www.rfc-editor.org/rfc/rfc8910.html

RFC8910 defines a captive portal API URI for a CP client to consume.

With captive_portal_uri set to `https://test.example.com`

Produces in RA:
```
ICMPv6 Option (DHCP Captive-Portal)
    Type: DHCP Captive-Portal (37)
    Length: 4 (32 bytes)
    Captive Portal: https://test.example.com
```
And in DHCPv6 Reply:
```
Captive Portal
    Option: Captive Portal (103)
    Length: 24
    Captive Portal: https://test.example.com
```

I have a mirror PR for odhcp6c. There are some compliant efforts out there that can utilise this. 
RFC8908 (API) 
RFC8910 (API URI option)
RFC8952 (Architecture)

ping @Alphix @Noltari


Noltari (approved):
LGTM
@Alphix any objections?

Alphix:
Some random thoughts:

Why a global option and not a per-iface one? (Think guest vs internal network)

README.md should say which services this applies to?

`_capt_port` style naming might be a bit confusing? (No ports involved). `_captive_uri` or `_cp_uri`?

systemcrash:
> Some random thoughts:
> 
> Why a global option and not a per-iface one? (Think guest vs internal network)

Didn't think of that. Although the CP tends to apply globally. I can't think why a CP admin would want to have several different CP on the same network, but as an on/off for specific interfaces, yeah.

> README.md should say which services this applies to?

There are a few projects related to openwrt which utilise these RFC. There is travelmate - but it doesn't look like it utilises this option yet.

systemcrash:
OK. All fixed.

src/router.c
```diff
@@ -1015,6 +1024,31 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 	iov[IOV_RA_ADV_INTERVAL].iov_base = (char *)&adv_interval;
 	iov[IOV_RA_ADV_INTERVAL].iov_len = adv_interval.nd_opt_adv_interval_len * 8;
 
+	/* RFC 8910 Captive Portal */
+	uint8_t *captive_portal_uri = (uint8_t *)iface->captive_portal_uri;
+	capt_portal_uri_len = strlen(iface->captive_portal_uri);
```
Alphix: I suggest you add `iface->captive_portal_uri_len` so that you can skip doing `strlen()` on every single RA/DHCPv6 packet...

src/router.c
```diff
@@ -1015,6 +1024,31 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
...
+	if (capt_portal_uri_len > 0) {
+		/* compute pad so that (header + data + pad) is a multiple of 8 */
+		size_t hdr = sizeof(struct nd_opt_capt_portal); /* should be 2 */
+		size_t total_without_pad = hdr + capt_portal_uri_len;
+		size_t pad = (8 - (total_without_pad % 8)) % 8;
+
+		size_t data_padded = capt_portal_uri_len + pad;
+		capt_portal_sz = hdr + data_padded;
```
Alphix: These lines can be replaced with:
`capt_portal_sz = (sizeof(struct nd_opt_capt_portal) + iface->captive_portal_uri_len + 7) & ~7;`

src/dhcpv6.c
```diff
@@ -645,6 +662,8 @@ static void handle_client_request(void *addr, void *data, size_t len,
 		[IOV_DNR] = {dnrs, dnrs_len},
 		[IOV_RELAY_MSG] = {NULL, 0},
 		[IOV_DHCPV4O6_SERVER] = {&dhcpv4o6_server, 0},
+		[IOV_CAPT_PORTAL] = {&capt_portal, (capt_portal_len) ? sizeof(capt_portal) : 0},
+		[IOV_CAPT_PORTAL_URI] = {capt_portal_ptr, (capt_portal_len) ? capt_portal_len : 0},
```
Alphix: The parens around the first part of the ternary operators can be dropped, they make it look like it's a macro

src/config.c
```diff
@@ -1304,6 +1308,12 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 		}
 	}
 
+	if ((c = tb[IFACE_ATTR_CAPTIVE_PORTAL_URI])) {
+		iface->captive_portal_uri = strdup(blobmsg_get_string(c));
+		info("Set RFC8910 captive portal URI: '%s' for interface '%s'",
```
Alphix: Should we really blort out `info()` level messages saying we're doing what the user told us to do? Feels more like `debug()` to me...

Alphix:
> > README.md should say which services this applies to?
> 
> There are a few projects related to openwrt which utilise these RFC. There is travelmate - but it doesn't look like it utilises this option yet.

Sorry, I was being too terse...I meant, `README.md` should maybe note that this is currently enabled for DHCPV6/RA (since the same thing exists for DHCPv4 as well).

src/config.c
systemcrash: Sure. Fixed.

src/dhcpv6.c
systemcrash: Fair.

src/router.c
systemcrash: :O

src/router.c
systemcrash: Fair. Was hoping to avoid extra iface variables. Your wall is full :D

systemcrash:
I completely omitted DHCPv4. for me odhcpd is 'ipv6'. Maybe it's not a bad idea to add it for DHCPv4 also.

Added DHCPv4 for your reviewing pleasure. :D

Alphix:
> I completely omitted DHCPv4. for me odhcpd is 'ipv6'. Maybe it's not a bad idea to add it for DHCPv4 also.

It is my sinister long-term plan to replace `dnsmasq` with `odhcpd` as the default DHCPv4 daemon in OpenWrt ;)

src/config.c
```diff
@@ -1304,6 +1308,17 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 		}
 	}
 
+	if ((c = tb[IFACE_ATTR_CAPTIVE_PORTAL_URI])) {
+		iface->captive_portal_uri = strdup(blobmsg_get_string(c));
+		iface->captive_portal_uri_len = strlen(iface->captive_portal_uri);
+		if (iface->captive_portal_uri_len > 254) {
```
Alphix: 254? It's not NULL-terminated so I think you want 255...:
`if (iface->captive_portal_uri_len > UINT8_MAX)`

src/dhcpv4.c
```diff
@@ -1289,6 +1295,14 @@ void dhcpv4_handle_msg(void *src_addr, void *data, size_t len,
 			iov[IOV_DNR_BODY].iov_base = dnrs;
 			iov[IOV_DNR_BODY].iov_len = dnrs_len;
 			break;
+
+		case DHCPV4_OPT_CAPTIVE_PORTAL:
+			captive_portal.len = iface->captive_portal_uri_len;
+			if (captive_portal.len > 254)
+				break;
```
Alphix: Check len first, then do stuff if it's ok

src/dhcpv4.c
```diff
@@ -1289,6 +1295,14 @@ void dhcpv4_handle_msg(void *src_addr, void *data, size_t len,
...
+			iov[IOV_CAPTIVE_PORTAL].iov_len = sizeof(captive_portal) + iface->captive_portal_uri_len;
+			iov[IOV_CAPTIVE_PORTAL].iov_base = iface->captive_portal_uri;
```
Alphix: You need two IOVs, right? Header and body/data separately...

src/config.c
systemcrash: My bad - I assumed length was included in that count. But it's because length is a single octet field... Got it.

src/dhcpv4.c
systemcrash: That's one way of doing it, but I did mix that up. Should be better now. Thanks!

Alphix:
LGTM

Noltari:
Merged, thanks @systemcrash!

PR #316, opened 2025-11-17
github: ci: add MIPS64, PowerPC64 and RISCV64

Noltari:
MIPS64, PowerPC64 and RISCV64 are popular OpenWrt archs.
Refactor the sizes build step to generate the table programatically.


Noltari:
@systemcrash I also inverted the table columns/rows like you suggested ;)

PR #317, opened 2025-11-18
Relay flags

systemcrash:
Continuation of #300 whereby we grant control of A and P flags on RA relay, based on interface settings. 

Discuss? @Alphix @Noltari 

A seems to be contentious while P seems necessary.


systemcrash:
opinions on this @Alphix @Noltari ?

Noltari (approved):
LGTM

Noltari:
@Alphix any objections?

Alphix:
> @Alphix any objections?

I still find the logic a bit quirky. But I don't think that's reason enough to hold this up (and the behaviour can always be tweaked later). So shoot...

Noltari:
Merged, thanks @systemcrash!

PR #318, opened 2025-11-19
dhcpv4: use BPF to filter incoming requests

Alphix:
This uses a BPF to let the kernel do the preliminary checks for us, meaning `odhcpd` won't have to care about some bogus packets and e.g. messages from other DHCP servers.

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -25,9 +25,11 @@
 #include <net/if.h>
 #include <net/if_arp.h>
 #include <netinet/ip.h>
+#include <netinet/udp.h>
 #include <sys/ioctl.h>
 #include <sys/timerfd.h>
 #include <arpa/inet.h>
+#include <linux/filter.h>
 
 #include <libubox/md5.h>
 
@@ -953,14 +955,6 @@ void dhcpv4_handle_msg(void *src_addr, void *data, size_t len,
 	if (iface->dhcpv4 == MODE_DISABLED)
 		return;
 
-	/* FIXME: would checking the magic cookie value here break any clients? */
-
-	if (len < offsetof(struct dhcpv4_message, options) ||
-	    req->op != DHCPV4_OP_BOOTREQUEST ||
-	    req->htype != ARPHRD_ETHER ||
-	    req->hlen != ETH_ALEN)
-		return;
-
 	debug("Got DHCPv4 request on %s", iface->name);
 
 	if (!iface->dhcpv4_start_ip.s_addr && !iface->dhcpv4_end_ip.s_addr) {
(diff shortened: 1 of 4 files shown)


Alphix:
@systemcrash @Noltari ping

src/dhcpv4.c
```diff
@@ -1435,15 +1429,53 @@ static int dhcpv4_setup_addresses(struct interface *iface)
 	return 0;
 }
 
-int dhcpv4_setup_interface(struct interface *iface, bool enable)
+struct dhcpv4_packet {
+        struct udphdr udp;
```
Noltari: This should be indented with tabs instead of whitespaces.

src/dhcpv4.c
```diff
@@ -1435,15 +1429,53 @@ static int dhcpv4_setup_addresses(struct interface *iface)
...
+bool dhcpv4_setup_interface(struct interface *iface, bool enable)
 {
-	int ret = 0;
-	struct sockaddr_in bind_addr = {
+	/* Note: we could check more things (but buggy clients exist), e.g.:
+	 *  - DHCPV4_MIN_PACKET_SIZE
+	 *  - yiaddr zero
+	 *  - siaddr zero
+	 */
+	struct sock_filter filter[] = {
```
Noltari: Same comment as https://github.com/openwrt/odhcpd/pull/319#discussion_r2545211923.
Should we declare this as static in order to reduce stack usage?

Noltari (changes requested):

systemcrash:
Is this a viable avenue to fix that crash we saw in the forum? A cursory check using BPF in the kernel against the udp length header and the actual packet length?

systemcrash (approved):

Alphix:
> Is this a viable avenue to fix that crash we saw in the forum? A cursory check using BPF in the kernel against the udp length header and the actual packet length?

I dont think so. I've gotten some more pcaps and there seems to be some renewal events that succeed and some that don't (and the trailing garbage seems to be gone from the pcaps). I'll get SSH access to the VM, but it'll have to wait until next week or so when I have more time...

systemcrash:
> (and the trailing garbage seems to be gone from the pcaps)

The observer effect 😱

src/dhcpv4.c
Alphix: Uff, did some minor edits on a Windows client...

src/dhcpv4.c
Alphix: Yeah, `static const` it is...

Alphix:
> > (and the trailing garbage seems to be gone from the pcaps)
> 
> The observer effect 😱

My theory is that the user moved the VM that caused the crash to a different machine, so a different version of qemu/kvm/whatever...seems to have fixed the garbage but not the crash...

Alphix:
Updated

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #319, opened 2025-11-19
ndp: correctness fix for BPF filter

Alphix:
Note that AF_PACKET sockets start receiving packets as soon as they are created.  Thus, a packet can arrive between the creation of the socket and the time the real filter is installed. Fix this using the same technique as used in libpcap, i.e. by installing a drop-all filter, removing any packets from the socket, then installing the real filter later (an atomic operation which replaces the drop filter).

### DIFF
--- a/src/ndp.c
+++ b/src/ndp.c
@@ -39,17 +39,6 @@ static void setup_addr_for_relaying(struct in6_addr *addr, struct interface *ifa
 static void handle_solicit(void *addr, void *data, size_t len,
 		struct interface *iface, void *dest);
 
-/* Filter ICMPv6 messages of type neighbor solicitation */
-static struct sock_filter bpf[] = {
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari and @systemcrash ^

src/ndp.c
```diff
@@ -67,6 +56,24 @@ int ndp_init(void)
 
 int ndp_setup_interface(struct interface *iface, bool enable)
 {
+	/* Drop everything */
+	static struct sock_filter bpf_drop_filter[] = {
```
Noltari: Why is this one static but the rest aren't?
I can't really check right now, but should the others be static in order to reduce the stack usage?

systemcrash (approved):

src/ndp.c
Alphix: Actually, I wanted to make everything `static const `... I'll update the PR

Alphix:
Ok, should be fixed now

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #320, opened 2025-11-23
odhcpd: refactor, improve pool handling, fix realloc errors

Alphix:
This is a bunch of little changes that I didn't really feel deserved individual PRs.
  - The DHCPv4 pool logic is simplified
  - DHCPv4 now works with static-only addresses
  - Fixed all realloc() calls tree-wide
  - Renamed a bunch of variables/structs, making the naming clearer
  - Removed some more magic values
  - Minor correctness fix to the RA DNSSL option

Needs a bit more testing, but posting this here now for review.


Alphix:
@Noltari and @systemcrash ^

Noltari (approved):
LGTM
@systemcrash any objections?

src/dhcpv6.c
```diff
@@ -835,7 +838,8 @@ static void relay_server_response(uint8_t *data, size_t len)
 	int32_t ifaceidx = 0;
 	struct sockaddr_in6 target = {AF_INET6, htons(DHCPV6_CLIENT_PORT),
 		0, IN6ADDR_ANY_INIT, 0};
-	int otype, olen;
+	int otype;
```
systemcrash: can we go ahead and make this uint16_t? I don't recall this var doing anything special here.

systemcrash:
I've given these a once over and these look decent. I'll give it another look tomorrow.

systemcrash (commented):
Does this prevent /8 or /16 dhcp subnets? Or do we just build upon present behaviour?

src/config.c
```diff
@@ -1420,8 +1420,14 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 		iface->dhcpv4_forcereconf = blobmsg_get_bool(c);
 
 	if ((c = tb[IFACE_ATTR_DHCPV6_RAW])) {
+		void *tmp;
+
+		tmp = realloc(iface->dhcpv6_raw, blobmsg_data_len(c) / 2);
```
systemcrash: maybe a var for `blobmsg_data_len(c) / 2` since that's done twice here.

Alphix:
> Does this prevent /8 or /16 dhcp subnets? Or do we just build upon present behaviour?

Behavior shouldn't change, anything with a prefix length <= 24 gets treated the same way...

src/dhcpv6.c
Alphix: Yeah, that would also match the other functions which use `uint16_t`

src/config.c
Alphix: Fixed

Alphix:
@systemcrash @Noltari new version pushed, fixed a minor bug with DHCP pool calculation, addressed the review suggestions...

systemcrash:
I think this is okay, although I get the feeling that one day someone will come along and go "we need more granular control over DHCP subnetting" :)

Alphix:
> I think this is okay, although I get the feeling that one day someone will come along and go "we need more granular control over DHCP subnetting" :)

Yeah, including things like "it's now possible to have several IPv4 subnets on a given interface these days, so we need to be able to pick which one is actually used for the DHCPv4 pool"...etc...and I actually have a simple patch to also allow `/29` subnets to also be used (8 addresses, 2 taken for network/broadcast, 1 taken for the DHCPv4 server, 5 addresses remaining)...`/30` would theoretically be possible (grand total of one address to hand out)...buuuuut....this PR doesn't change the existing logic

Noltari:
Merged, thanks @Alphix!

Issue #321, opened 2025-11-23
odhcpd: segfaults in x86/64 with 6.12.58


GoetzGoerisch:
Since the last update on openwrt-main I'm seeing segfault other the course of a day.

```
[71802.880039] odhcpd[22696]: segfault at 78 ip 000055accd096596 sp 00007ffee6cd5d70 error 4 in odhcpd[c596,55accd08e000+f000] likely on CPU 1 (core 1, socket 0)
[71802.883955] Code: 0f 84 ba 00 00 00 66 81 7c 24 60 00 19 0f 84 dd 00 00 00 48 39 6b 28 0f 85 d3 00 00 00 45 31 db 4c 89 5c 24 38 48 8b 44 24 38 <48> 3b 45 78 0f 83 bc 00 00 00 4c 69 6c 24 38 8c 00 00 00 48 8b 85
[74506.650803] odhcpd[22094]: segfault at 78 ip 00005626f9e85596 sp 00007ffe32bba580 error 4 in odhcpd[c596,5626f9e7d000+f000] likely on CPU 0 (core 0, socket 0)
[74506.657931] Code: 0f 84 ba 00 00 00 66 81 7c 24 60 00 19 0f 84 dd 00 00 00 48 39 6b 28 0f 85 d3 00 00 00 45 31 db 4c 89 5c 24 38 48 8b 44 24 38 <48> 3b 45 78 0f 83 bc 00 00 00 4c 69 6c 24 38 8c 00 00 00 48 8b 85
[76951.851428] odhcpd[31238]: segfault at 78 ip 0000556cc228a596 sp 00007ffed90360a0 error 4 in odhcpd[c596,556cc2282000+f000] likely on CPU 1 (core 1, socket 0)
[76951.856279] Code: 0f 84 ba 00 00 00 66 81 7c 24 60 00 19 0f 84 dd 00 00 00 48 39 6b 28 0f 85 d3 00 00 00 45 31 db 4c 89 5c 24 38 48 8b 44 24 38 <48> 3b 45 78 0f 83 bc 00 00 00 4c 69 6c 24 38 8c 00 00 00 48 8b 85
[78709.800993] odhcpd[9463]: segfault at 78 ip 000056534bbea596 sp 00007ffce0b61950 error 4 in odhcpd[c596,56534bbe2000+f000] likely on CPU 1 (core 1, socket 0)
[...]

PR #322, opened 2025-11-23
statefiles: fix off-by-one-bug

Alphix:
The string with the final hostsfile name is `%s.%s`, so we need two extra bytes for the separating `.` and for the trailing null byte.

See: https://github.com/openwrt/openwrt/issues/20910
Maybe: https://github.com/openwrt/odhcpd/issues/321

### DIFF
--- a/src/statefiles.c
+++ b/src/statefiles.c
@@ -103,7 +103,7 @@ static void statefiles_write_hosts(time_t now)
 	avl_for_each_element(&interfaces, ctxt.iface, avl) {
 		char *hostsfile;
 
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

hnyman:
@Noltari 
When you import this (and other today's changes) to OpenWrt, could you update also the Maintainer field in the odhcpd Makefile?
@dedeckeh is still listed there (and in several other packages) as the maintainer, although he has not committed anything since 2022.

PR #323, opened 2025-11-24
odhcpd: supports rewrite DNS server to local IPv6 DNS server in relay mode

jjm2473:
Just like server mode.

My ISP provides IPv6 DNS in the RA. In odhcpd relay mode, internal network devices use the IPv6 DNS advertised by the ISP, causing internal DNS requests to bypass OpenWRT's DNS server. Therefore, relay mode needs to support overriding the DNS server.

In relay mode, `dns_service` is false by default, so this change doesn't affect the original logic. A future PR will be given to LuCI to allow modification of `dns_service` in relay mode.

### DIFF
--- a/src/dhcpv6.c
+++ b/src/dhcpv6.c
@@ -885,7 +885,7 @@ static void relay_server_response(uint8_t *data, size_t len)
 	}
 
 	/* Rewrite DNS servers if requested */
-	if (iface->always_rewrite_dns && dns_ptr && dns_count > 0) {
+	if ((iface->always_rewrite_dns || iface->dns_service) && dns_ptr && dns_count > 0) {
 		if (is_authenticated)
 			return; /* Impossible to rewrite */
 
--- a/src/router.c
+++ b/src/router.c
@@ -1240,7 +1240,7 @@ static void forward_router_advertisement(const struct interface *iface, uint8_t 
 		}
 
 		/* If we have to rewrite DNS entries */
-		if (c->always_rewrite_dns && dns_ptr && dns_count > 0) {
+		if ((c->always_rewrite_dns || c->dns_service) && dns_ptr && dns_count > 0) {
 			const struct in6_addr *rewrite = c->dns;
 			struct in6_addr addr;
 			size_t rewrite_cnt = c->dns_cnt;


Noltari:
@Alphix @systemcrash can you take a look at this?

systemcrash:
This is a low-cost fix, but the initial post and commit message need rewriting to support the commit's reasoning. 

It's possible to have dnsmasq capture everything on port (8)53 and send that upstream via DoH or whatever, but this might not be viable for everyone.

Maybe something like:

In odhcpd RA/DHCPv6 relay mode, downstream network devices receive the upstream IPv6 DNS. This commit enables overriding the DNS server in relay modes (with the exception of authenticated DHCPv6), whereby the local device is substituted as the DNS server. Since `dns_service` is false by default, this change does not affect the default behaviour.

jjm2473:
I was wrong. `dns_service` doesn't default to false. Switching the mode to relay on LuCI only hides the `dns_service` option; it doesn't become false. 

If `dns_service` is not specified, it will always be true (in `set_interface_defaults`):
https://github.com/openwrt/odhcpd/blob/650a5df88ebae82d956831c2f87c7c6eedd8c283/src/config.c#L329

jjm2473:
Should we first enable LuCI to support configuring `dns_service` in relay mode?
I want to override the ISP's IPv6 DNS with the local DNS in relay mode, which theoretically should be configured using `dns_service`.

Alphix:
> Should we first enable LuCI to support configuring `dns_service` in relay mode? I want to override the ISP's IPv6 DNS with the local DNS in relay mode, which theoretically should be configured using `dns_service`.

I haven't looked into this in detail yet, but there's (from the top of my head) two options that are relevant.

One is if `dns` has been set explicitly, the other one is if `dns_service` has been set.

My main concern is not whether this makes sense...I agree that being able to override the ISP's DNS servers is desirable. It's more about what the cfg should look like and how it should behave, both in LuCI and in the cfg files.

(and forgive me if I missed something simple here...I haven't looked into the relevant parts of the code, I'm busy with other hacks)

systemcrash:
Hi @jjm2473 have you given this some more thought?

jjm2473:
Hi @systemcrash :
This patch currently meets the needs of most users, but in relay mode, it overrides the DNS provided by the RA/DHCPv6 by default, which is different from before. However, I think this is reasonable because the description of `dns_service` in LuCI is 'Announce this device as IPv6 DNS server' It shouldn't differ based on server or relay mode.

This patch has another issue: it only overrides DNS provided by the upstream RA/DHCPv6. If the upstream RA/DHCPv6 itself doesn't provide DNS, then the downstream won't receive IPv6 DNS (it will generally fall back to IPv4 DNS).

Actually, for me, as long as it overridden the DNS provided by the ISP, even completely deleting it is fine, because the ISP-provided DNS allows OpenWrt's DNS to be bypassed by clients, which is the most important issue.

PR #325, opened 2025-11-26
CI: fix formal workflow permissions

GeorgeSapkin:
Fix formality check permissions that are needed to post optional summaries back to the PR.

Posting comments is disabled here. Otherwise repo's permissions need to be updated for actions to modify PRs:
- https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment

Related:
- https://github.com/openwrt/actions-shared-workflows/pull/64

### DIFF
--- a/.github/workflows/formal.yml
+++ b/.github/workflows/formal.yml
@@ -5,8 +5,14 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   build:
     name: Test Formalities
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari not sure if you've seen this one, but if it fixes the error message I get with every PR, it would be nice to have it committed :)

Noltari:
Merged, thanks @GeorgeSapkin!

PR #326, opened 2025-11-27
dhcpv6-ia: fix realloc bug

Alphix:
Commit b9db4d7061a08bf82a25222074065cce71973d0c introduced a bug, the "hostname" variable used for the realloc would shadow the real hostname defined at the beginning of the function. Fix this by using a different variable name.


Alphix:
@Noltari ^

Noltari:
Merged, thanks @Alphix!

PR #330, opened 2025-11-28
router: improve RA PIO search logic

Noltari:
We need to properly handle when the user modifies the prefix length assigned to an interface by:
- Checking if the prefix matches the address taking into account:
    - The biggest length of address and PIO.
    - A minimum PIO of /64.
- Updating addresses and lengths for existing PIOs.

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -449,14 +449,20 @@ struct nd_opt_capt_portal {
 };
 
 /* IPv6 RA PIOs */
+inline static int router_compare_pio_addr(const struct ra_pio *pio, const struct odhcpd_ipaddr *addr)
+{
+	uint8_t cmp_len = max(64, max(pio->length, addr->prefix_len));
+
+	return odhcpd_bmemcmp(&pio->prefix, &addr->addr.in6, cmp_len);
+}
+
 static struct ra_pio *router_find_ra_pio(struct interface *iface,
 	struct odhcpd_ipaddr *addr)
 {
 	for (size_t i = 0; i < iface->pio_cnt; i++) {
 		struct ra_pio *cur_pio = &iface->pios[i];
 
-		if (addr->prefix_len == cur_pio->length &&
-		    !odhcpd_bmemcmp(&addr->addr.in6, &cur_pio->prefix, cur_pio->length))
+		if (!router_compare_pio_addr(cur_pio, addr))
 			return cur_pio;
 	}
 
@@ -471,6 +477,23 @@ static void router_add_ra_pio(struct interface *iface,
 
 	pio = router_find_ra_pio(iface, addr);
 	if (pio) {
+		if (memcmp(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr)) != 0 ||
+		    pio->length != addr->prefix_len)
+		{
+			char new_ipv6_str[INET6_ADDRSTRLEN];
+
+			iface->pio_update = true;
+			warn("rfc9096: %s: changed %s/%u -> %s/%u",
+			     iface->ifname,
+			     inet_ntop(AF_INET6, &pio->prefix, ipv6_str, sizeof(ipv6_str)),
+			     pio->length,
+			     inet_ntop(AF_INET6, &addr->addr.in6, new_ipv6_str, sizeof(new_ipv6_str)),
+			     addr->prefix_len);
+
+			memcpy(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr));
+			pio->length = addr->prefix_len;
+		}
+
 		if (pio->lifetime) {
 			pio->lifetime = 0;
 
@@ -492,7 +515,7 @@ static void router_add_ra_pio(struct interface *iface,
 	pio = &iface->pios[iface->pio_cnt];
 	iface->pio_cnt++;
 
-	memcpy(&pio->prefix, &addr->addr.in6, sizeof(pio->prefix));
+	memcpy(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr));
 	pio->length = addr->prefix_len;
 	pio->lifetime = 0;
 
@@ -659,8 +682,7 @@ static int send_router_advert(struct interface *iface, const struct in6_addr *fr
 			for (size_t j = 0; j < valid_addr_cnt; j++) {
 				struct odhcpd_ipaddr *cur_addr = &addrs[j];
 
-				if (cur_pio->length == cur_addr->prefix_len &&
-				    !odhcpd_bmemcmp(&cur_pio->prefix, &cur_addr->addr.in6, cur_pio->length)) {
+				if (!router_compare_pio_addr(cur_pio, cur_addr)) {
 					pio_found = true;
 					break;
 				}


Noltari:
I'm not really sure if just doing:
```
uint8_t pfx_len = min(addr->prefix_len, cur_pio->length);
if (!odhcpd_bmemcmp(&addr->addr.in6, &cur_pio->prefix, pfx_len))
	...
```
would be safe for some strange cases...

ping @Alphix @systemcrash

Alphix:
> I'm not really sure if just doing:
> 
> ```
> uint8_t pfx_len = min(addr->prefix_len, cur_pio->length);
> if (!odhcpd_bmemcmp(&addr->addr.in6, &cur_pio->prefix, pfx_len))
> 	...
> ```
> 
> would be safe for some strange cases...

You mean to detect `addr_changed` and `len_changed` at the same time?

src/router.c
```diff
@@ -455,6 +455,11 @@ static struct ra_pio *router_find_ra_pio(struct interface *iface,
 	for (size_t i = 0; i < iface->pio_cnt; i++) {
 		struct ra_pio *cur_pio = &iface->pios[i];
 
+		/* Same exact address */
+		if (!memcmp(&addr->addr.in6, &cur_pio->prefix, sizeof(struct in6_addr)))
+			return cur_pio;
```
Alphix: This means that `router_find_ra_pio()` will consider e.g. `2001:db8::/32` and `2001:db8::/48` as being the same prefix? Is that intentional?

Noltari: > This means that `router_find_ra_pio()` will consider e.g. `2001:db8::/32` and `2001:db8::/48` as being the same prefix? Is that intentional?

Yes, but we're actually comparing the full address here, so in normal cases it would be `2001:db8::1/32` and `2001:db8::1/48`.
Currently, if the user changes the delegated prefix length from `/64` to `/60`, odhcpd will anounce `2001:db8::1/64` as stale and `2001:db8::1/60` as valid, which is incorrect and most devices will loose IPv6 connectivity.

Noltari:
> You mean to detect `addr_changed` and `len_changed` at the same time?

I mean to replace both if blocks:
```
/* Same exact address */
if (!memcmp(&addr->addr.in6, &cur_pio->prefix, sizeof(struct in6_addr)))
			return cur_pio;

/* Same prefix address */
		if (addr->prefix_len == cur_pio->length &&
		    !odhcpd_bmemcmp(&addr->addr.in6, &cur_pio->prefix, cur_pio->length))
			return cur_pio;
```
with:
```
uint8_t pfx_len = min(addr->prefix_len, cur_pio->length);
if (!odhcpd_bmemcmp(&addr->addr.in6, &cur_pio->prefix, pfx_len))
	return cur_pio;
```

src/router.c
```diff
@@ -471,6 +476,28 @@ static void router_add_ra_pio(struct interface *iface,
...
+			     pio->length,
+			     inet_ntop(AF_INET6, &addr->addr.in6, new_ipv6_str, sizeof(new_ipv6_str)),
+			     addr->prefix_len);
+		}
+
+		if (addr_changed)
+			memcpy(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr));
+		if (len_changed)
+			pio->length = addr->prefix_len;
+
```
Alphix: This could be written as:
```C
if (addr->prefix_len != pio->length || odhcpd_bmemcmp(&addr->in6, &pio->prefix, pio->length)) {
    char new_ipv6_str[INET6_ADDRSTRLEN];
    warn("blabla...");
    memcpy(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr));
    pio->length = addr->prefix_len;
}
```

But also...doesn't this mean that if the prefix that's being announced changes from e.g. `2001:db8::/32` to `2001:db8::/48`, we'll just announce the new value and ignore the old value (rather than advertising `::/32` as stale and `::/48` as live), meaning downstream devices will still think that the `::/32` prefix is valid and on-link?

src/router.c
Alphix: But if I understand the logic correctly, with this patch applied, odhcpd will make the opposite mistake if the prefix shrinks?

I.e. a change from `/60` to `/64` will mean that the `/64` will be announced as valid and the `/60` won't be announced as stale?

src/router.c
Alphix: (The advantage of doing `addr->prefix_len != pio->length` first is that `odhcpd_bmemcmp()` is skipped altogether if the prefix changes, the advantage of `odhcpd_bmemcmp()` is that changing the address of an interface from e.g. `2001:db8::1/64` to `2001:db8::2/64` will not be seen as a prefix change)

Noltari: > This could be written as:
> 
> ```c
> if (addr->prefix_len != pio->length || odhcpd_bmemcmp(&addr->in6, &pio->prefix, pio->length)) {
>     char new_ipv6_str[INET6_ADDRSTRLEN];
>     warn("blabla...");
>     memcpy(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr));
>     pio->length = addr->prefix_len;
> }
> ```
> 
> But also...doesn't this mean that if the prefix that's being announced changes from e.g. `2001:db8::/32` to `2001:db8::/48`, we'll just announce the new value and ignore the old value (rather than advertising `::/32` as stale and `::/48` as live), meaning downstream devices will still think that the `::/32` prefix is valid and on-link?

I might need to retest what happens if devices receive a stale ::/32 and a valid ::/48 at the same time, but I think that most devices loose IPv6 connectibity.

Anyway, I think I'll mark this as draft for now and I'll try to test this thoroughly...

Alphix: > I might need to retest what happens if devices receive a stale ::/32 and a valid ::/48 at the same time, but I think that most devices loose IPv6 connectibity.

That might be entirely true...I haven't tested that...

Noltari: After more testing:
1. Debian 13 -> IPv6 connectivity, but some addresses are marked with preferred_lt of 0:
```
$ sudo ip -6 addr show dev enp1s0
2: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    altname enx88c9b3b40b94
    inet6 2a02:9142:XXXX:XXXX:EUI64/64 scope global dynamic mngtmpaddr noprefixroute
       valid_lft 5358sec preferred_lft 2658sec
    inet6 2a02:9142:XXXX:XXXX::X/64 scope global deprecated dynamic mngtmpaddr proto kernel_ra
       valid_lft 5358sec preferred_lft 0sec
    inet6 fd91::EUI64/64 scope global dynamic mngtmpaddr noprefixroute
       valid_lft 5358sec preferred_lft 2658sec
    inet6 fd91::X/64 scope global deprecated dynamic mngtmpaddr proto kernel_ra
       valid_lft 5358sec preferred_lft 0sec
    inet6 fe80::EUI64/64 scope link proto kernel_ll
       valid_lft forever preferred_lft forever
```
3. Windows -> no IPv6 addresses
4. iOS -> OK

Maybe the issue is that the valid PIO is before the stale PIO.
I have to test if inverting the PIOs and sending the stale PIOs first fixes the issue...

Noltari: @Alphix so I tested reordering the stale PIOs to the beginning and it fixes the issue with windows devices -> https://github.com/Noltari/odhcpd/commit/c4dd3f38c5a6b6ce8d142db772a0b4432e3416a8

However, I still think that we should identify `2001:db8::1/32` and `2001:db8::/48` as the same PIO, since it will be announced as `2001:db8::/64` anyway...

EDIT: you can test that by manually adding a /63 with the same IP address to your `piofolder` interface file.

Alphix: Hm...now that I think about it...I'm a bit confused...we only support PIOs with `/64`, right? Anything else breaks SLAAC. So there shouldn't even be a PIO with `2001:db8::1/32` or `2001:db8::1/48` (that should be the RIO)...and a change in the prefix size shouldn't have any effect on the PIOs (unless we have a completely new prefix)...or have I missed/overlooked/misunderstood something?

Alphix: It seems we do support PIOs with prefixes longer than /64 as well, but it still seems a bit weird that the files storing the PIO info contains stuff like `"length":56`...that seems incorrect? We only ever advertise `/64`?

src/router.c
Noltari: We're discussing the same on two different threads, so I'll mark this one as resolve and continue on the other.

src/router.c
Noltari: > It seems we do support PIOs with prefixes longer than /64 as well, but it still seems a bit weird that the files storing the PIO info contains stuff like `"length":56`...that seems incorrect? We only ever advertise `/64`?

This was changed in 5eac9c56ff3b0a013c5241f449ca144f70bf4c02 because subnets < 64 were broken.
I believe the current implementation storing the real length instead of the capped one for subnets < 64 is correct, because that's what we use to find the corresponding interface address.

systemcrash: > However, I still think that we should identify `2001:db8::1/32` and `2001:db8::/48` as the same PIO, since it will be announced as `2001:db8::/64` anyway...

I spent a whole weekend hacking through this. netifd gets passed something from upstream odhcp6c, and gives the interface what it has as its length preference set (IPv6 assignment length: "Assign a part of given length of every public IPv6-prefix to this interface"). 

(unimportant paragraph) I was playing with Prefix Exclusions and DHCPv6-PD which *does not work* unless it's a static IPv6 e.g. /32 assigned on e.g. wan6. I was saaaad 😢  (I want to open a discussion for this and see if we can improve the design of this, so we can do smart stuff like prefix exclusions and delegate more of our prefix downstream (oh, and to make PD actually work - but... netifd :)

Yes and no: `2001:db8::1/32` and `2001:db8::/48` are not strictly the same PIO I think. If we get assigned /32 upstream, and we slap /48 or /64 on lan, then we should be able to carve out other prefixes from the /32 and dish those out downstream, each one being a unique PIO - although downstream routing should effectively point to the delegating router (us) to reach the `2001:db8::1/32` net. Same prefix, just slightly different PIOs. RIO are available for this. But for us/odhcpd to ID a prefix, then yeah, they're the same - I think that's what you're referring to here since the PR is about 9096 :)

So I would love to be able to assign anything from /32 to /112. The bit math could do with reworking. I was tempted to go uint_128, but this might crush older 32 bit mips CPUs 🙈 which can maybe only handle uint_64.

systemcrash: IIRC I think there's a prefix compare function which seems a bit underused, it's either here or in odhcp6c. 

Back on Wednesday to be able to think through this case more thoroughly unless you're happy with the results.

Alphix: > Yes and no: `2001:db8::1/32` and `2001:db8::/48` are not strictly the same PIO I think.

The more I look at it, the more I'm convinced that this is wrong.

What we _announce_ is a PIO for SLAAC which is `/64` (even if e.g. a `/48` is assigned to the lan interface), and RFC9096 is all about telling clients that the SLAAC PIO is no longer with us (i.e. the `/64` we told clients to use).

So if we used to have `2001:db8::1/32`, and announced a PIO of `2001:db8::1/64`, and that later changes to `2001:db8::/48`, and we still announce a PIO of `2001:db8::1/64`, RFC9096 doesn't come into play because there is no stale PIO prefix to announce.

There's a separate issue with the code in that prefixes with infinite validity seems to need special handling. We create a PIO RA from that and set the lifetime to something less than infinity, and it's _that_ lifetime which should be the basis for the stale PIO announcements...

> If we get assigned /32 upstream, and we slap /48 or /64 on lan

We shouldn't/can't slap a `/48` on LAN, it'll break SLAAC. Or rather, we can slap a `/whatever` on LAN, but we still announce a PIO of `/64`.

systemcrash: Ah yes, /64 is what SLAAC is supposed to work with. ( My comment also regarded PD which has more flex for lengths). https://www.ietf.org/archive/id/draft-mishra-v6ops-variable-slaac-problem-stmt-06.html

Alphix: It's actually an interesting rabbit hole that I fell into...PIOs and RIOs and having the luxury of more than a `/64` to play with...but my conclusion so far from going through the RFCs is that right now a prefix less than `/64` needs to be handled with a `/64` PIO and a "real" prefix length RIO....and that prefixes longer than `/64` is something that'll probably change in the future...

systemcrash: To hell with the specs. I want /1 prefixes XD

Noltari: > Back on Wednesday to be able to think through this case more thoroughly unless you're happy with the results.

@systemcrash I'm quite happy with the results and I believe this covers all the possible user changes without breaking anything.
So, if you and @Alphix are OK with this change I'll merge and the other stuff taken into account in this discussion can be addressed in subsequent PRs.

BTW, I've just pushed some minor changes adding a new function: `ra_pio_length()`.

src/odhcpd.h
```diff
@@ -511,6 +511,11 @@ inline static bool ra_pio_expired(const struct ra_pio *pio, time_t now)
 	return pio->lifetime && (now > pio->lifetime);
 }
 
+inline static uint8_t ra_pio_length(const struct ra_pio *pio, const struct odhcpd_ipaddr *addr)
+{
+	return max(64, max(pio->length, addr->prefix_len));
```
systemcrash: So, for now, we want to nail the max to 64? This works today, but with future ideas to loosen this to more arbitrary length, a choice comment like:
```
/* future: relax from 64 to arbitrary length */
```
? Maybe a more descriptive function name... `get_longest_ra_pio_pfx_len`?

Noltari: > So, for now, we want to nail the max to 64? This works today, but with future ideas to loosen this to more arbitrary length, a choice comment like:
> ```
> /* future: relax from 64 to arbitrary length */
> ```
> ? Maybe a more descriptive function name... `get_longest_ra_pio_pfx_len`? 

64 is the minimum length that a PIO can have today and I don't think that's going to change in the future.
If a /48 is assigned to the interface, we want to compare the address with a minimum length of 64, which is the length that's actually announced in the corresponding PIO.

systemcrash: How should the function be named? If the prefix is 96 bits (and the suffix 32), that's longer than /64, while a /32 bit prefix gives a much bigger subnet. Same thing just different viewpoints :) Here we seem to want the prefix length, or at least the longest of the two.

Noltari: > How should the function be named? If the prefix is 96 bits (and the suffix 32), that's longer than /64, while a /32 bit prefix gives a much bigger subnet. Same thing just different viewpoints :) Here we seem to want the prefix length, or at least the longest of the two.

What about `ra_pio_match_addr_len()`?

Noltari: Refactored to `router_compare_pio_addr()`

src/router.c
Alphix: I'll be quite busy until next week, so can't really review in detail...if you're happy with the patch, then shoot

src/odhcpd.h
systemcrash: Better 😂

PR #331, opened 2025-11-28
odhcpd: simplify statefile, remove flags

Alphix:
First, the statefile is simplified by removing the hosts entries from the file. Hosts are already handled by the separate `hostsdir` directive, which is enabled by default.

Second, go through the various `OAF_` flags and remove a bunch of flags which are either redundant, or which can be replaced with a simple `bool`.


Alphix:
@Noltari and @systemcrash ^

Noltari (approved):
LGTM

systemcrash (approved):
Yeah, good.

Noltari:
Merged, thanks @Alphix!

PR #332, opened 2025-11-30
statefiles: change ipv4 MAC statefile syntax

Alphix:
This changes the MAC representation in the lease/statefile so that it reads `aa:bb:cc:dd:ee:ff` instead of `aabbccddeeff`.

No functional changes, but this makes it easier to parse the statefile in `rpcd-mod-luci` (for reporting the MAC addresses in LuCI).

We should probably change `lease->hwaddr` to be a `struct ether_addr`, to make it clearer that we only support MAC addresses as hwaddr, but that's a topic for a different PR.

### DIFF
--- a/src/statefiles.c
+++ b/src/statefiles.c
@@ -18,6 +18,7 @@
 #include <fcntl.h>
 #include <time.h>
 #include <netinet/in.h>
+#include <netinet/ether.h>
 #include <arpa/nameser.h>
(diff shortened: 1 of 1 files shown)


Alphix:
@Noltari trivial PR ^

Noltari (approved):

Noltari:
Merged, thanks @Alphix!

PR #333, opened 2025-11-30
odhcpd: stale PIO handling fixes

Alphix:
In essence, this PR:
 - Moves the PIO file handling to statefiles.c, where we can share the logic for creating tmp files
 - Centralises the handling of dirfds
 - Changes the "piofolder" cfg option to "piodir"
 - Fixes a bug in the PIO file handling when running without ubus


Alphix:
@Noltari I think you'll find this one interesting :)

@systemcrash feel free to have a look as well

systemcrash:
Well, I do like JSON, for narrowing in on data quickly when a parser is available...

Alphix:
> Well, I do like JSON, for narrowing in on data quickly when a parser is available...

It's a ~20% increase in binary size, 200 lines more of code and the JSON version of the file looks something like this:

```
{"slaac":[{"prefix":"2001:db8:b::1","length":48,"lifetime":1764546646},{"prefix":"2001:db8:b::1","length":48}]}
```

versus (with this PR applied):

```
2001:db8:b::1 48 1764546646
2001:db8:b::1 48 -1
```

systemcrash:
I guess that makes sense. Would \t or , be a better separator? a la TSV/CSV. Perhaps a comment header for sanity is good (e.g. # IPv6_prefix prefix_length remaining_lifetime)

Noltari:
> As far as I can tell from the GitHub CI output, `odhcpd` without JSON is about 17% smaller.

@Alphix NACK -> https://github.com/openwrt/odhcpd/pull/334

Let's keep it as JSON please.

Alphix:
> > As far as I can tell from the GitHub CI output, `odhcpd` without JSON is about 17% smaller.
> 
> @Alphix NACK -> #334

That makes sense, the difference in binary size didn't make much sense to me.

> Let's keep it as JSON please.

Not to be confrontational, but could you elaborate? What are the advantages of using JSON here?

Noltari:
> > Let's keep it as JSON please.
> 
> Not to be confrontational, but could you elaborate? What are the advantages of using JSON here?

This is probably a personal opinion, but I think it's a better option for structuring data and is more tolerant to changes (adding or removing existing parameters).

Alphix:
> > > Let's keep it as JSON please.
> > 
> > 
> > ...What are the advantages of using JSON here?
> 
> This is probably a personal opinion, but I think it's a better option for structuring data and is more tolerant to changes (adding or removing existing parameters).

I completely agree with you that something like JSON is better as a data _exchange_ format, e.g. if we (`odhcpd`) want to write data and that data is supposed to be read by some other program. Case in point here is `odhcpd`s statefile, which is read by `rpcd-mod-luci` - an issue that I've tried to resolve for a couple of weeks since I want to extend the statefile so that DHCP leases can survive a restart/reboot with _all_ of the lease context intact (I have an idea in mind, but I need a couple of days of uninterrupted hacking to actually get it working, no idea when I'll have the time).

Here, however, we're storing completely internal data, data that's supposed to be used to do something sane after a crash/power loss/etc. And we don't _want_ to encourage other programs reading that data. In this case, I'd argue that simplicity is key. The `fwrite()` path is, as you can see in this PR, ~200 lines less of code, alloc-free, and so simple that the read and write functions both fit on the screen at the same time. And the format is still extensible to a certain degree (as long as we only add parameters, we're safe, just need to check the return value from the `sscanf()` to see how many elements it parsed).

systemcrash:
You’re onto something with sharing, but unless a source is stable or static, the json consumers read from ubus. JSON is just a really good way to keep things structured, yet amendable. In the name is object notation. Counter point: there’s good reason not to poke around in state files for those of sound mind 😅

Alphix:
> You’re onto something with sharing, but unless a source is stable or static, the json consumers read from ubus.

Yes, ubus is good and something we should push external consumers towards using. But we're not talking about ubus now?

> JSON is just a really good way to keep things structured, yet amendable. In the name is object notation. Counter point: there’s good reason not to poke around in state files for those of sound mind 😅

I'm confused, how is that a counter point? I agree that it would be better if noting poked around in our state files, but that's not the world we live in right now...and I want to fix that. But it was just an example, and kind of orthogonal to this PR...

Noltari:
@Alphix remove https://github.com/openwrt/odhcpd/pull/333/commits/da4417b899701db1595309a31dcad45995ff50ff and rebase it on current master, please.
Let's keep it as JSON, thank you.

Noltari:
BTW, I have already started working on `piofolder` to `piodir` migration:
https://github.com/Noltari/openwrt/commit/06b7775a9164d429fb58339dcecf5a8ba6aa4c52

systemcrash:
Man, that’s a bit of migration baggage to carry just to rename a variable 🙁

Alphix:
>> BTW, I have already started working on `piofolder` to `piodir` migration: [Noltari/openwrt@06b7775](https://github.com/Noltari/openwrt/commit/06b7775a9164d429fb58339dcecf5a8ba6aa4c52)

That's nice, addressing the breakage I'm causing, thanks :)

> Man, that’s a bit of migration baggage to carry just to rename a variable 🙁

The question is if it's *really* necessary? This will only affect people moving from one devel build to another devel build, and not a stable release -> stable release, right?

Can't we just skip the migration (except for the variable name change)?, something like:

```shell
	piofolder=$(uci -q get dhcp.odhcpd.piofolder)
	if [ -n "$piofolder" ]; then
		uci delete dhcp.odhcpd.piofolder
		commit=1
	fi

	piodir=$(uci -q get dhcp.odhcpd.piodir)
	if [ -z "$piodir" ]; then
		uci set dhcp.odhcpd.piodir=/tmp/odhcpd-piodir
		commit=1
	fi
```

It only runs the risk of a one-time-non-announcement of stale prefixes for a devel->devel upgrade, right?

Alphix:
I've dropped the patch removing the JSON format and rebased. Compile-tested, but I haven't had time to do some live tests yet, I'll do that later today.

systemcrash:
> The question is if it's _really_ necessary? This will only affect people moving from one devel build to another devel build, and not a stable release -> stable release, right?

Those who go snapshot -> stable are caught by this. Their configs smuggle in the parameter. I don't run snapshot on my primary. But evidently there are those who do. Developers choice to let them suffer :)

Noltari:
> > Man, that’s a bit of migration baggage to carry just to rename a variable 🙁
> 
> The question is if it's _really_ necessary? This will only affect people moving from one devel build to another devel build, and not a stable release -> stable release, right?
> 
> Can't we just skip the migration (except for the variable name change)?, something like:
> 
> ```shell
> 	piofolder=$(uci -q get dhcp.odhcpd.piofolder)
> 	if [ -n "$piofolder" ]; then
> 		uci delete dhcp.odhcpd.piofolder
> 		commit=1
> 	fi
> 
> 	piodir=$(uci -q get dhcp.odhcpd.piodir)
> 	if [ -z "$piodir" ]; then
> 		uci set dhcp.odhcpd.piodir=/tmp/odhcpd-piodir
> 		commit=1
> 	fi
> ```
> 
> It only runs the risk of a one-time-non-announcement of stale prefixes for a devel->devel upgrade, right?

`piofolder` is also present in 24.10.4, so we need to migrate the files too.
- https://github.com/openwrt/openwrt/commit/be32ab85cff01370da77ec88925edefa97ced1a7
- https://github.com/openwrt/openwrt/blob/v24.10.4/package/network/services/odhcpd/files/odhcpd.defaults

src/statefiles.c
```diff
@@ -40,6 +42,321 @@ struct write_ctxt {
...
+
+	renameat(dirfd, ODHCPD_TMP_FILE, dirfd, filename);
+}
+
+#define JSON_LENGTH "length"
+#define JSON_PREFIX "prefix"
+#define JSON_SLAAC "slaac"
+#define JSON_TIME "time"
+
+static inline time_t config_time_from_json(time_t json_time)
```
Noltari: should we rename this to `statefiles_time_from_json`?

src/statefiles.c
```diff
@@ -40,6 +42,321 @@ struct write_ctxt {
...
+	ref = time(NULL);
+	now = odhcpd_time();
+
+	if (now > json_time || ref > json_time)
+		return 0;
+
+	return json_time + (now - ref);
+}
+
+static inline time_t config_time_to_json(time_t config_time)
```
Noltari: should we rename this to `statefiles_time_to_json`?

src/statefiles.c
```diff
@@ -40,6 +42,321 @@ struct write_ctxt {
...
+{
+	time_t ref, now;
+
+	ref = time(NULL);
+	now = odhcpd_time();
+
+	return config_time + (ref - now);
+}
+
+static inline bool config_ra_pio_enabled(struct interface *iface)
```
Noltari: should we rename this to `statefiles_ra_pio_enabled`?

src/statefiles.c
```diff
@@ -40,6 +42,321 @@ struct write_ctxt {
...
+{
+	return config.ra_piodir_fd >= 0 && iface->ra == MODE_SERVER && !iface->master;
+}
+
+static bool config_ra_pio_time(json_object *slaac_json, time_t *slaac_time)
```
Noltari: should we rename this to `statefiles_ra_pio_time`?

src/statefiles.c
```diff
@@ -40,6 +42,321 @@ struct write_ctxt {
...
+	pio_time = config_time_from_json(pio_json_time);
+	if (!pio_time)
+		return false;
+
+	*slaac_time = pio_time;
+
+	return true;
+}
+
+static json_object *config_load_ra_pio_json(struct interface *iface)
```
Noltari: should we rename this to `statefiles_load_ra_pio_json`?

Noltari:
Merged, thanks @Alphix!

Alphix:
Wrt the renaming, yes, the functions should be renamed, _but_ I plan to use some of them in the statefiles functions as well (the time handling), because the exact same functionality is needed there (converting from monotime to time_t and back). It's a bigger fix though, because I want monotime to use a separate datatype and well...yeah, I just let it be for now 🤷🏻‍♂️

GoetzGoerisch:
Did I miss it or where is the LuCi change for this?
@Alphix @Noltari

Alphix:
> Did I miss it or where is the LuCi change for this? @Alphix @Noltari

Nah, I missed it, I've created a PR now:
https://github.com/openwrt/luci/pull/8148

PR #336, opened 2025-12-01
router: remove duplicated PIOs

Noltari:
There can be duplicated PIOs if the user changed the assigned prefix length in the configuration.

~~Draft until https://github.com/openwrt/odhcpd/pull/330 is merged.~~

### DIFF
--- a/src/router.c
+++ b/src/router.c
@@ -544,9 +544,7 @@ static void router_clear_duplicated_ra_pio(struct interface *iface)
 				     inet_ntop(AF_INET6, &pio_a->prefix, ipv6_str, sizeof(ipv6_str)),
 				     pio_a->length);
 
-				if (j + 1 < iface->pio_cnt)
-					iface->pios[j] = iface->pios[iface->pio_cnt - 1];
-
+				iface->pios[j] = iface->pios[iface->pio_cnt - 1];
 				iface->pio_cnt--;
 			} else {
 				j++;
@@ -577,9 +575,7 @@ static void router_clear_expired_ra_pio(time_t now,
 			     inet_ntop(AF_INET6, &cur_pio->prefix, ipv6_str, sizeof(ipv6_str)),
 			     cur_pio->length);
 
-			if (i + 1 < iface->pio_cnt)
-				iface->pios[i] = iface->pios[iface->pio_cnt - 1];
-
+			iface->pios[i] = iface->pios[iface->pio_cnt - 1];
 			iface->pio_cnt--;
 		} else {
 			i++;
--- a/src/router.c
+++ b/src/router.c
@@ -538,8 +538,7 @@ static void router_clear_duplicated_ra_pio(struct interface *iface)
 		while (j < iface->pio_cnt) {
 			struct ra_pio *pio_b = &iface->pios[j];
 
-			if (pio_a->length == pio_b->length &&
-			    !memcmp(&pio_a->prefix, &pio_b->prefix, sizeof(struct in6_addr))) {
+			if (!memcmp(pio_a, pio_b, ra_pio_cmp_len)) {
 				warn("rfc9096: %s: clear duplicated %s/%u",
 				     iface->ifname,
 				     inet_ntop(AF_INET6, &pio_a->prefix, ipv6_str, sizeof(ipv6_str)),
--- a/src/router.c
+++ b/src/router.c
@@ -526,7 +526,7 @@ static void router_add_ra_pio(struct interface *iface,
 	     pio->length);
 }
 
-static void router_clear_ra_pio(time_t now,
+static void router_clear_expired_ra_pio(time_t now,
 	struct interface *iface)
 {
 	size_t i = 0, pio_cnt = iface->pio_cnt;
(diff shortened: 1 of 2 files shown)


src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
...
+	for (size_t i = 0; i < iface->pio_cnt; i++) {
+		struct ra_pio *pio_a = &iface->pios[i];
+		size_t j = i + 1;
+
+		while (j < iface->pio_cnt) {
+			struct ra_pio *pio_b = &iface->pios[j];
+
+			if (pio_a->length == pio_b->length &&
+			    !memcmp(&pio_a->prefix, &pio_b->prefix, sizeof(struct in6_addr))) {
+				warn("rfc9096: %s: duplicated %s/%u",
```
systemcrash: Perhaps info in the warning that it's being removed also

Noltari: Done in 0320c852f3684b0a20a4ed2088d16335a772fd59

src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
...
+			    !memcmp(&pio_a->prefix, &pio_b->prefix, sizeof(struct in6_addr))) {
+				warn("rfc9096: %s: clear duplicated %s/%u",
+				     iface->ifname,
+				     inet_ntop(AF_INET6, &pio_a->prefix, ipv6_str, sizeof(ipv6_str)),
+				     pio_a->length);
+
+				if (j + 1 < iface->pio_cnt)
+					iface->pios[j] = iface->pios[iface->pio_cnt - 1];
+
+				iface->pio_cnt--;
```
systemcrash: How about skipping the if branch?: 

```
iface->pios[j] = iface->pios[iface->pio_cnt - 1];
iface->pio_cnt--;
```

src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
 	     pio->length);
 }
 
-static void router_clear_ra_pio(time_t now,
+static void router_clear_duplicated_ra_pio(struct interface *iface)
+{
+	size_t pio_cnt = iface->pio_cnt;
+	char ipv6_str[INET6_ADDRSTRLEN];
+
+	for (size_t i = 0; i < iface->pio_cnt; i++) {
```
systemcrash, line 534:
```c
for (size_t i = 0; i < pio_cnt; i++) {
```

src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
...
+		struct ra_pio *pio_a = &iface->pios[i];
+		size_t j = i + 1;
+
+		while (j < iface->pio_cnt) {
```
systemcrash, line 538:
```c
		while (j < pio_cnt) {
```

src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
...
+
+			if (pio_a->length == pio_b->length &&
+			    !memcmp(&pio_a->prefix, &pio_b->prefix, sizeof(struct in6_addr))) {
+				warn("rfc9096: %s: clear duplicated %s/%u",
+				     iface->ifname,
+				     inet_ntop(AF_INET6, &pio_a->prefix, ipv6_str, sizeof(ipv6_str)),
+				     pio_a->length);
+
+				if (j + 1 < iface->pio_cnt)
+					iface->pios[j] = iface->pios[iface->pio_cnt - 1];
```
systemcrash:
```c
					iface->pios[j] = iface->pios[pio_cnt - 1];
```

src/router.c
```diff
@@ -526,7 +526,44 @@ static void router_add_ra_pio(struct interface *iface,
...
+
+	for (size_t i = 0; i < iface->pio_cnt; i++) {
+		struct ra_pio *pio_a = &iface->pios[i];
+		size_t j = i + 1;
+
+		while (j < iface->pio_cnt) {
+			struct ra_pio *pio_b = &iface->pios[j];
+
+			if (pio_a->length == pio_b->length &&
+			    !memcmp(&pio_a->prefix, &pio_b->prefix, sizeof(struct in6_addr))) {
```
systemcrash: Maybe we can just memcmp the ra_pio struct up to lifetime, and skip the separate length compare.

systemcrash:
The whole search thing is quadratic complexity. But we don't typically expect a large number of prefixes handled, so optimising to something more rapid like log O(n) can happen later.

src/router.c
Noltari: Done in https://github.com/openwrt/odhcpd/pull/336/commits/d354ebb66cdcccc3373156468eec1c660fb2922c

src/router.c
Noltari: Done in https://github.com/openwrt/odhcpd/pull/336/commits/79252ed0c0c26659a95729da0f2accaa0ddb458d

src/router.c
Noltari: What's the point in looping over PIOs that have already been identified as duplicated/expired?

systemcrash: ? My suggestion is to use the variable you just created :)

Noltari: > ? My suggestion is to use the variable you just created :)

The variable that I created was meant to evaluate if we needed to do a realloc at the end.
However, for looping we should use iface->pio_cnt since it's the one being decreased.

src/router.c
Noltari: https://github.com/openwrt/odhcpd/pull/336#discussion_r2591649959

src/router.c
Noltari: https://github.com/openwrt/odhcpd/pull/336#discussion_r2591649959

src/router.c
systemcrash: OK. All good - yeah you're decreasing it.

systemcrash:
LGTM.

PR #339, opened 2025-12-04
odhcpd: stale PIO handling fixes (with JSON)

Noltari:
In essence, this PR:
 - Moves the PIO file handling to statefiles.c, where we can share the logic for creating tmp files
 - Centralises the handling of dirfds
 - Changes the "piofolder" cfg option to "piodir"
 - Fixes a bug in the PIO file handling when running without ubus

Supersedes https://github.com/openwrt/odhcpd/pull/333, since it's basically the same without the JSON removal.

CC @Alphix @systemcrash


Noltari:
TODO: rename `config_*_json()` functions in `statefiles.c` to `statefiles_*_json()`

Alphix:
Why didn't you just reply to what I wrote in #333 instead?

Noltari:
> Why didn't you just reply to what I wrote in #333 instead?

@Alphix I already replied 2 times:
https://github.com/openwrt/odhcpd/pull/333#issuecomment-3593893888
https://github.com/openwrt/odhcpd/pull/333#issuecomment-3595096950

But yeah, I will reply there once more, no problem.

PR #340, opened 2025-12-05
Fix escape sequence for broken hostname output

ldir-EDB0:


Noltari:
ping @Alphix @systemcrash

systemcrash (approved):
This looks correct - although I'm not 100%. If we're in the business of introducing a space, is it not simply better to write "broken " or does the space somehow get elided by fprintf?

Noltari:
> This looks correct - although I'm not 100%. If we're in the business of introducing a space, is it not simply better to write "broken " or does the space somehow get elided by fprintf?

That's what I though too but it's always safer to just fix the bug :P.

Noltari:
Merged, thanks @ldir-EDB0!

Alphix:
"\x20" == " ", meaning you'll break the strtok-based parsing in "rpcd-mod-luci"...IOW, I think the original code behaved as intended?

systemcrash:
So....  \x20 is incorrect and becomes a literal space? But \\x20 is 'correct' and becomes \x20 here, and doesn't get chewed up by strtok downstream, but converted at display time?

ldir-EDB0:
In luci I see `eth1	broken\x20-` so it isn't converted at display time.  But my 'fix' isn't a fix it turns out because it breaks the parsing which is based on delimiting by space character.

So I think the fix needs to happen in luci somehow.  My apologies for yet more noise.

Noltari:
> In luci I see `eth1	broken\x20-` so it isn't converted at display time.  But my 'fix' isn't a fix it turns out because it breaks the parsing which is based on delimiting by space character.
> 
> 
> 
> So I think the fix needs to happen in luci somehow.  My apologies for yet more noise.

@ldir-EDB0 no worries

Shall we revert it then?

ldir-EDB0:
@Noltari Yes, a revert is a good move.

Alphix:
> So.... \x20 is incorrect and becomes a literal space? But \x20 is 'correct' and becomes \x20 here, and doesn't get chewed up by strtok downstream, but converted at display time?

@systemcrash Yes, `\` is an escape character, so `"\x20"` in C means hexadecimal value 0x20...which is a normal blank space in ASCII/UTF-8. And the C string `"\\x20"` results in `"\x20"` The convention has been that `odhcpd` writes out `broken\x20<whatever>` for broken hostnames (as a single value) for a long time now.

This is something that can be improved down the line. For example: I suspect I could still break the existing functionality by a sufficiently malicious hostname, and we could report broken hostnames in a prettier manner in LuCI. But that requires more substantial rework. Moving to a `ubus` based communication with LuCI is more in line with a "real" fix, but that's one of my long-term goals. So I agree that a revert is the right thing to do for now.

PR #342, opened 2025-12-09
statefiles: don't write empty hosts files

### LINKED ISSUE #338: odhcpd.hosts files for all interfaces
odhcpd creates a odhcpd.hosts files for every interface, even for interfaces where this does not make eny sense (loopback, wan, wan6). It should only create these files where odhcpd really "listens" to, meaning where you have configured a dhcp section and which is not "ignored".

Alphix:
This is mostly cosmetic, skip writing out hosts files for interfaces where it makes little sense to do so (i.e. where the hosts file will anyway be empty).

Closes: https://github.com/openwrt/odhcpd/issues/338

### DIFF
--- a/src/statefiles.c
+++ b/src/statefiles.c
@@ -415,7 +415,15 @@ static void statefiles_write_hosts(time_t now)
 		return;
 
 	avl_for_each_element(&interfaces, ctxt.iface, avl) {
+		if (ctxt.iface->ignore)
+			continue;
+
+		if (ctxt.iface->dhcpv4 != MODE_SERVER && ctxt.iface->dhcpv6 != MODE_SERVER)
+			continue;
+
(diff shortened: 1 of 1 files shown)


Alphix:
Draft for now since this will cause conflicts with #333

systemcrash:
All set @Alphix

systemcrash:
Do we need those state files for any of the pasthru scenarios?

Alphix:
> Do we need those state files for any of the pasthru scenarios?

No, the hosts files are only used for DNS (e.g. `dnsmasq` or `unbound`) for clients that are known to `odhcpd` because they have obtained a DHCPv4/v6 lease, so relaying (I guess that's what you mean with passthrough?) isn't affected.

systemcrash:
Ok. Good then this looks ready. Maybe we can squeak it in with the update in openwrt main.

Noltari:
Merged, thanks @Alphix!

PR #348, opened 2025-12-13
router: misc fixes

Alphix:
Here's a bunch of little fixes and correctness issues in `src/router.c`

Most should be uncontroversial, but @Noltari, I think the third patch ("only store PIO prefix and length") needs careful scrutiny (and this is your area of expertise). I've tried to test it thoroughly, and it seems to work as expected.

I'm still not sure if we handle cases where the whole change is that the prefix length changes, e.g. `2001:db8::1/48` -> `2001:db8::1/96` or vice versa, but that's not actually changed by this PR (see the discussion in https://github.com/openwrt/odhcpd/pull/330)

### DIFF
--- a/src/statefiles.c
+++ b/src/statefiles.c
@@ -143,23 +143,23 @@ static inline bool statefiles_ra_pio_enabled(struct interface *iface)
 
 static bool statefiles_ra_pio_time(json_object *slaac_json, time_t *slaac_time)
 {
-	time_t pio_json_time, pio_time;
+	time_t pio_json_time, pio_time = 0;
 	json_object *time_json;
 
 	time_json = json_object_object_get(slaac_json, JSON_TIME);
 	if (!time_json)
-		return true;
+		goto out;
 
-	pio_json_time = (time_t) json_object_get_int64(time_json);
+	pio_json_time = (time_t)json_object_get_int64(time_json);
 	if (!pio_json_time)
-		return true;
+		goto out;
 
 	pio_time = statefiles_time_from_json(pio_json_time);
 	if (!pio_time)
 		return false;
 
+out:
 	*slaac_time = pio_time;
-
 	return true;
 }
 
@@ -220,7 +220,7 @@ void statefiles_read_prefix_information(struct interface *iface)
 	for (size_t i = 0; i < pio_cnt; i++) {
 		json_object *cur_pio_json, *length_json, *prefix_json;
 		const char *pio_str;
-		time_t pio_lt = 0;
+		time_t pio_lt;
 		struct ra_pio *pio;
 		uint8_t pio_len;
 
--- a/src/router.c
+++ b/src/router.c
@@ -450,20 +450,21 @@ struct nd_opt_capt_portal {
 };
 
 /* IPv6 RA PIOs */
-inline static int router_compare_pio_addr(const struct ra_pio *pio, const struct odhcpd_ipaddr *addr)
+inline static int
+router_compare_pio_addr(const struct ra_pio *pio, const struct in6_addr *addr, uint8_t prefix_len)
 {
-	uint8_t cmp_len = max(64, max(pio->length, addr->prefix_len));
+	uint8_t cmp_len = max(64, max(pio->length, prefix_len));
 
-	return odhcpd_bmemcmp(&pio->prefix, &addr->addr.in6, cmp_len);
+	return odhcpd_bmemcmp(&pio->prefix, addr, cmp_len);
 }
 
-static struct ra_pio *router_find_ra_pio(struct interface *iface,
-	struct odhcpd_ipaddr *addr)
+static struct ra_pio *
+router_find_ra_pio(struct interface *iface, const struct in6_addr *addr, uint8_t prefix_len)
 {
 	for (size_t i = 0; i < iface->pio_cnt; i++) {
 		struct ra_pio *cur_pio = &iface->pios[i];
 
-		if (!router_compare_pio_addr(cur_pio, addr))
+		if (!router_compare_pio_addr(cur_pio, addr, prefix_len))
 			return cur_pio;
 	}
 
@@ -476,7 +477,7 @@ static void router_add_ra_pio(struct interface *iface,
 	char ipv6_str[INET6_ADDRSTRLEN];
 	struct ra_pio *new_pios, *pio;
 
-	pio = router_find_ra_pio(iface, addr);
+	pio = router_find_ra_pio(iface, &addr->addr.in6, addr->prefix_len);
 	if (pio) {
 		if (memcmp(&pio->prefix, &addr->addr.in6, sizeof(struct in6_addr)) != 0 ||
 		    pio->length != addr->prefix_len)
@@ -599,7 +600,7 @@ static void router_stale_ra_pio(struct interface *iface,
 	struct odhcpd_ipaddr *addr,
 	time_t now)
 {
-	struct ra_pio *pio = router_find_ra_pio(iface, addr);
+	struct ra_pio *pio = router_find_ra_pio(iface, &addr->addr.in6, addr->prefix_len);
 	char ipv6_str[INET6_ADDRSTRLEN];
 
 	if (!pio || pio->lifetime)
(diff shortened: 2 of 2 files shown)


Noltari:
@Alphix please split commits 5 (527f1f262d558107b8ff4dca43a71064eae2b096) and 6 (7fb1c4c8f8cf9af908336eff720e0b5ac8384ae6) into a separte PR so that we can discuss this properly.

I will take care of fixing the comparison in `router_clear_duplicated_ra_pio()` (commit 4 d02fce006abd4f0761bf3100c9ddd4ce70847edb) by reverting d354ebb66cdcccc3373156468eec1c660fb2922c, which was introduced during the review of https://github.com/openwrt/odhcpd/pull/336.

Therefore this PR should be left with commits 1, 2 and 3.

Alphix:
> @Alphix please split commits 5...and 6...

Yes, makes sense. I'll post a separate PR for that once we're done with this one since I have a feeling there might be merge conflicts otherwise...

> I will take care of...commit 4...

Yeah, I see you've already done that

> Therefore this PR should be left with commits 1, 2 and 3.

Done :)

src/statefiles.c
```diff
@@ -143,23 +143,23 @@ static inline bool statefiles_ra_pio_enabled(struct interface *iface)
...
 
 	pio_time = statefiles_time_from_json(pio_json_time);
 	if (!pio_time)
 		return false;
 
+out:
 	*slaac_time = pio_time;
-
 	return true;
 }
```
Noltari, line 164: I'm not really a fan of `goto`, so I'd prefer rewriting this as:
```C
static time_t statefiles_ra_pio_time(json_object *slaac_json)
{
	json_object *time_json;
	time_t pio_json_time;

	time_json = json_object_object_get(slaac_json, JSON_TIME);
	if (!time_json)
		return 0;

	pio_json_time = (time_t) json_object_get_int64(time_json);
	if (!pio_json_time)
		return 0;

	return statefiles_time_from_json(pio_json_time);
}
```

And then:
```C
void statefiles_read_prefix_information(struct interface *iface)
{
		...
		time_t pio_lt;
		...
		pio_lt = statefiles_ra_pio_time(cur_pio_json);
		if (!pio_lt)
			continue;
		...
}
```

Noltari:
@Alphix from commit 3 (ee7f7ae4a1b8dcd09a667fbf019455d2119537c6)

> This leads to warnings being printed e.g. if the interface address changes
> without the PIO actually changing:
>
>	rfc9096: foo-server: changed fd00:bbbb::1/48 -> fd00:bbbb::2/48
>
> But the PIO hasn't actually changed.
>
> Fix this by only storing the actual prefix and length that have been
> announced, i.e. fd00:bbbb::/64 and not fd00:bbbb::1/48. This also means
> that the data in the JSON file is clearer, since the prefix will match
> what is actually announced via the RAs.

I don't really think this is a problem, but we can supress the warning by using `odhcpd_bmemcmp` and the prefix length, so there's no need to rewrite all the code for that.

I honestly would prefer to keep the full address in the PIO files because it might help debugging issues.
Maybe what we could do is change the JSON files and switch from "prefix" to "address". We can add a conditional to accept both of them as input and switch the output to "address".

src/statefiles.c
Alphix: I don't think that logic would be correct because it would conflate expired PIOs and PIOs without a lifetime?

Alphix:
> @Alphix from commit 3 ([ee7f7ae](https://github.com/openwrt/odhcpd/commit/ee7f7ae4a1b8dcd09a667fbf019455d2119537c6))
> 
> > This leads to warnings being printed e.g. if the interface address changes
> > without the PIO actually changing:
> 
> I don't really think this is a problem, but we can supress the warning by using `odhcpd_bmemcmp` and the prefix length,

It's not just about the address, though, it's about storing what's actually been announced to clients. If we store a different prefix length than what's actually been announced, then we actually *lose* information - i.e. if we store a `/56`, we can't tell after the fact whether a given `/64` has been previously announced as a PIO or not. IOW, I think this makes it easier, not harder, to debug issues down the road.

> so there's no need to rewrite all the code for that.

Sorry about that, I tried to keep the changes as minimal as possible, and if you ignore the noise from using separate `struct in6_addr` and `uint8_t prefix_len` arguments instead of `struct odhcpd_ipaddr *addr`, I think you'll see that it's actually just a few lines that are truly changed.

src/statefiles.c
Noltari: > I don't think that logic would be correct because it would conflate expired PIOs and PIOs without a lifetime?

You're totally right.
Lately I tend to do things on a hurry and it's very difficult for me to focus on something (I guess that's normal with a 1 year-old daughter...).

```
static bool statefiles_ra_pio_time(json_object *slaac_json, time_t *slaac_time)
{
	time_t pio_json_time, pio_time;
	json_object *time_json;

	time_json = json_object_object_get(slaac_json, JSON_TIME);
	if (!time_json) {
		*slaac_time = 0;
		return true;
	}

	pio_json_time = (time_t) json_object_get_int64(time_json);
	if (!pio_json_time) {
		*slaac_time = 0;
		return true;
	}

	pio_time = statefiles_time_from_json(pio_json_time);
	if (!pio_time) {
		*slaac_time = 0; /* I don't think we actually need this one */
		return false;
	}

	*slaac_time = pio_time;

	return true;
}
```

Noltari:
> > @Alphix from commit 3 ([ee7f7ae](https://github.com/openwrt/odhcpd/commit/ee7f7ae4a1b8dcd09a667fbf019455d2119537c6))
> > > This leads to warnings being printed e.g. if the interface address changes
> > > without the PIO actually changing:
> > 
> > 
> > I don't really think this is a problem, but we can supress the warning by using `odhcpd_bmemcmp` and the prefix length,
> 
> It's not just about the address, though, it's about storing what's actually been announced to clients. If we store a different prefix length than what's actually been announced, then we actually _lose_ information

But the current code isn't using that information directly, it's only comparing the actual prefix length.

>  - i.e. if we store a `/56`, we can't tell after the fact whether a given `/64` has been previously announced as a PIO or not. IOW, I think this makes it easier, not harder, to debug issues down the road.

If we store a `/56` we can asume that it was announced as a `/64` because it's only added when the RA message is being created.

Also, these warnings will only happen if the user changes the delegated prefix length or the ifaceid, so it's not like they will start showing up without the user changing stuff.

> 
> > so there's no need to rewrite all the code for that.
> 
> Sorry about that, I tried to keep the changes as minimal as possible, and if you ignore the noise from using separate `struct in6_addr` and `uint8_t prefix_len` arguments instead of `struct odhcpd_ipaddr *addr`, I think you'll see that it's actually just a few lines that are truly changed.

That's the real issue: I don't see any added value in the changes and therefore I prefer to keep the code as is.

src/statefiles.c
Alphix: > > I don't think that logic would be correct because it would conflate expired PIOs and PIOs without a lifetime?
> 
> You're totally right. Lately I tend to do things on a hurry and it's very difficult for me to focus on something (I guess that's normal with a 1 year-old daughter...).

I've been there, I'm impressed you manage to get anything done wrt. OpenWrt to be honest.

Noltari:
> I honestly would prefer to keep the full address in the PIO files because it might help debugging issues. Maybe what we could do is change the JSON files and switch from "prefix" to "address". We can add a conditional to accept both of them as input and switch the output to "address".

@Alphix I meant something like:
https://github.com/Noltari/odhcpd/compare/master...pio-rename-addr

Alphix:
> > I honestly would prefer to keep the full address in the PIO files because it might help debugging issues. Maybe what we could do is change the JSON files and switch from "prefix" to "address". We can add a conditional to accept both of them as input and switch the output to "address".
> 
> @Alphix I meant something like: [Noltari/odhcpd@master...pio-rename-addr](https://github.com/Noltari/odhcpd/compare/master...pio-rename-addr)

That would invalidate the PIOs that are already stored (once), don't really see the the benefit, but 🤷🏻‍♂️

Noltari:
> That would invalidate the PIOs that are already stored (once), don't really see the the benefit, but 🤷🏻‍♂️

https://github.com/openwrt/odhcpd/pull/356/files#r2616989032

PR #353, opened 2025-12-14
dhcpv6: implement RFC9686

### LINKED ISSUE #349: DHCPv6: should add support for RFC9686 (SLAAC address registration)
[RFC9686](https://www.ietf.org/rfc/rfc9686.html) allows SLAAC clients to register the address they picked with the DHCPv6 server, which has the potential of providing a much nicer IPv6 experience for users.

Google [has indicated](https://mailarchive.ietf.org/arch/msg/v6ops/Sq5TadeSsMQ-0uEWrdem3A1wDh0/) that they plan to roll out Android support "Over the next few months" (that was in September 2025).

Realistically, I don't think this will be 25.12 material though, since it'll require some changes to the DHCPv6 server which runs the risk of causing breakage, and we just branched 25.12.

Just leaving this as a placeholder here for now :)

systemcrash:
Registering Self-Generated IPv6 Addresses Using DHCPv6

This functionality registers a client SLAAC or static address in
the DHCPv6 pool, primarily for diagnostic purposes as outlined in §1.

Most of the machinery required for this functionality already exists,
so we perform a few additional sets and checks in the IA code.

The address registration mechanism overview:
```
+--------+        +------------------+       +---------------+
| CLIENT |        | FIRST-HOP ROUTER |       | DHCPv6 SERVER |
+--------+        +---------+--------+       +-------+-------+
    |      SLAAC            |                        |
    |<--------------------> |                        |
    |                       |                        |
    |                                                |
    |  src: link-local address                       |
    | -------------------------------------------->  |
    |    INFORMATION-REQUEST or SOLICIT/...          |
    |       - OPTION REQUEST OPTION                  |
    |          -- OPTION_ADDR_REG_ENABLE             |
    |                                                |
    |    ...                                         |
    |                                                |
    |                                                |
    |<---------------------------------------------  |
    |     REPLY or ADVERTISE MESSAGE                 |
    |       - OPTION_ADDR_REG_ENABLE                 |
    |                                                |
    |                                                |
    |  src: address being registered                 |
    | -------------------------------------------->  |
    |    ADDR-REG-INFORM MESSAGE                     |Register/
    |                                                |log addresses
    |                                                |
    |                                                |
    | <--------------------------------------------  |
    |        ADDR-REG-REPLY MESSAGE                  |
    |                                                |
```

Closes #349

Might require a few tweaks. This code is untested, but simple enough to merge and see production when this option appears (or we add it in odhcp6c and test that way - ULA are supposed to be registered).

@Alphix and @Noltari at your leisure.

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -1330,6 +1330,7 @@ out:
 /* Register or update address registration in the lease database */
 static bool register_ia_addr_in_lease_db(struct sockaddr_in6 *source,
 		const uint8_t *clientid_data, uint16_t clientid_len,
+		const char *clientfqdn_data, size_t clientfqdn_len,
 		const struct dhcpv6_ia_addr *ia_addr,
 		struct interface *iface)
 {
@@ -1388,6 +1389,19 @@ static bool register_ia_addr_in_lease_db(struct sockaddr_in6 *source,
 		lease->duid_len = clientid_len;
 		memcpy(lease->duid, clientid_data, clientid_len);
 		lease->length = 128;
+
+		/* Add optional Client FQDN */
+		if (clientfqdn_len > 0 && clientfqdn_data) {
+			char *tmp = realloc(lease->hostname, clientfqdn_len + 1);
+			if (tmp) {
+				lease->hostname = tmp;
+				memcpy(lease->hostname, clientfqdn_data, clientfqdn_len);
+				lease->hostname[clientfqdn_len] = 0;
+				lease->hostname_valid = odhcpd_hostname_valid(lease->hostname);
+			}
+		}
+
+
 
 		/* Try to find matching interface prefix and extract host ID */
 		for (size_t i = 0; i < iface->addr6_len; i++) {
@@ -1543,6 +1557,8 @@ void handle_ia_addr_reg_inform(struct sockaddr_in6 *source,
 
 	uint8_t *clientid_data = NULL;
 	uint16_t clientid_len = 0;
+	char clientfqdn_data[256];
+	size_t clientfqdn_len = 0;
 	struct dhcpv6_ia_addr *ia_addr = NULL;
 
 	if (len < sizeof(*hdr)) {
@@ -1553,6 +1569,7 @@ void handle_ia_addr_reg_inform(struct sockaddr_in6 *source,
 	/* RFC9686 §4.2: Client MUST include Client Identifier option */
 	/* RFC9686 §4.2: ADDR-REG-INFORM MUST NOT contain Server Identifier */
 	/* RFC9686 §4.2: MUST contain exactly one IA Address option */
+	/* RFC9686 §4.2: MAY include other options, such as the Client FQDN option */
 
 	dhcpv6_for_each_option(opts, opts_end, otype, olen, odata) {
 		switch (otype) {
@@ -1566,6 +1583,16 @@ void handle_ia_addr_reg_inform(struct sockaddr_in6 *source,
 			/* RFC9686 §4.2: Server MUST discard messages with Server ID */
 			notice("ADDR-REG-INFORM: message contains Server Identifier, discarding");
 			return;
+		case DHCPV6_OPT_FQDN:
+			if (olen >= 2 && olen <= 255) {
+				uint8_t fqdn_buf[256];
+				memcpy(fqdn_buf, odata, olen);
+				fqdn_buf[olen++] = 0;
+
+				if (dn_expand(&fqdn_buf[1], &fqdn_buf[olen], &fqdn_buf[1], clientfqdn_data, sizeof(clientfqdn_data)) > 0)
+					clientfqdn_len = strcspn(clientfqdn_data, ".");
+			}
+			break;
 		case DHCPV6_OPT_IA_ADDR:
 			if (olen == sizeof(struct dhcpv6_ia_addr) - 4) {
 				if (ia_addr != NULL) {
@@ -1607,7 +1634,8 @@ void handle_ia_addr_reg_inform(struct sockaddr_in6 *source,
 	debug("Got ADDR-REG-INFORM for %s on %s", addrbuf, iface->name);
 
 	/* Register address in lease database */
-	if(!register_ia_addr_in_lease_db(source, clientid_data, clientid_len, ia_addr, iface))
+	if(!register_ia_addr_in_lease_db(source, clientid_data, clientid_len,
+									clientfqdn_data, clientfqdn_len, ia_addr, iface))
 		return;
 
 	/* Send ADDR-REG-REPLY response */
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -1326,3 +1326,290 @@ proceed:
 out:
 	return response_len;
 }
+
+/* Register or update address registration in the lease database */
+static bool register_ia_addr_in_lease_db(struct sockaddr_in6 *source,
+		const uint8_t *clientid_data, uint16_t clientid_len,
+		const struct dhcpv6_ia_addr *ia_addr,
+		struct interface *iface)
+{
+	/* RFC9686 §4.2.1: Lease lifetime is the valid-lifetime from IA Address option */
+	time_t now = odhcpd_time();
+	uint32_t valid_lt = ntohl(ia_addr->valid_lt);
+	time_t lease_end = now + valid_lt;
+	bool onlink = false;
+	/* variables for logging */
+	char addrbuf[INET6_ADDRSTRLEN];
+	inet_ntop(AF_INET6, &ia_addr->addr, addrbuf, sizeof(addrbuf));
+	char duidbuf[DUID_HEXSTRLEN] = {0};
+	const char *duidstr = duidbuf;
+
+	/* Search for existing binding with this address */
+	struct dhcpv6_lease *binding = NULL, *lease = NULL;
+	list_for_each_entry(binding, &iface->ia_assignments, head) {
+		/* Check if this address is already registered by another client */
+		if ((binding->flags & OAF_DHCPV6_ADDR_REG) &&
+		    memcmp(&binding->peer.sin6_addr, &ia_addr->addr, sizeof(struct in6_addr)) == 0) {
+			/* Found address registration for this address */
+			if (binding->duid_len == clientid_len &&
+			    memcmp(binding->duid, clientid_data, clientid_len) == 0) {
+				/* Same client, update the lease */
+				lease = binding;
+				break;
+			} else {
+				/* RFC9686 §4.2.1: Different client, should log and update binding
+				 * We'll update to the new client */
+
+				if (!binding->duid_len || binding->duid_len > DUID_MAX_LEN)
+					duidstr = "<invalid-duid>";
+				else
+					odhcpd_hexlify(duidbuf, binding->duid, binding->duid_len);
+
+				notice("ADDR-REG-INFORM: address collision for %s: was bound to "
+					"different client DUID(%s); updating", addrbuf, duidstr);
+				lease = binding;
+				break;
+			}
+		}
+	}
+
+	if (!lease) {
+		/* Create new lease for this address registration */
+		lease = dhcpv6_alloc_lease(clientid_len);
+		if (!lease) {
+			warn("ADDR-REG-INFORM: failed to allocate lease for %s", addrbuf);
+			return false;
+		}
+
+		/* Initialize lease structure */
+		lease->iface = iface;
+		lease->peer = *source;
+		lease->peer.sin6_addr = ia_addr->addr;  /* Store full registered address */
+		lease->duid_len = clientid_len;
+		memcpy(lease->duid, clientid_data, clientid_len);
+		lease->length = 128;
+
+		/* Try to find matching interface prefix and extract host ID */
+		for (size_t i = 0; i < iface->addr6_len; i++) {
+			if (!valid_addr(&iface->addr6[i], now))
+				continue;
+
+			if (ADDR_MATCH_PIO_FILTER(&iface->addr6[i], iface))
+				continue;
+
+			/* RFC9686 §4.2.1: the server SHOULD verify that the address 
+			 * is "appropriate to the link" */
+			if (odhcpd_bmemcmp(&ia_addr->addr, &iface->addr6[i].addr.in6,
+								iface->addr6[i].prefix_len) == 0) {
+				/* Address is within this prefix - extract host ID portion */
+				onlink = true;
+				uint64_t host_id = 0;
+				memcpy(&host_id, &ia_addr->addr.s6_addr[8], 8);
+				lease->assigned_host_id = be64toh(host_id);
+				break;
+			}
+		}
+
+		if (onlink) {
+			/* Add to interface's lease list */
+			list_add(&lease->head, &iface->ia_assignments);
+		} else {
+			/* RFC9686 §4.2.1: If the address fails verification, the server SHOULD log this fact */
+			notice("ADDR-REG-INFORM: address %s not on this link: %s; discarding", addrbuf, iface->name);
+			dhcpv6_free_lease(lease);
+			lease = NULL;
+			return onlink;
+		}
+
+	} else {
+		/* Update existing lease */
+		lease->peer = *source;
+		lease->peer.sin6_addr = ia_addr->addr;  /* Update registered address */
+		lease->duid_len = clientid_len;
+		memcpy(lease->duid, clientid_data, clientid_len);
+	}
+
+	/* Update lifetimes */
+	lease->valid_until = lease_end;
+	lease->preferred_until = now + ntohl(ia_addr->preferred_lt);
+	lease->bound = true;
+
+	/* Mark flags - RFC9686 Address Registration lease */
+	lease->flags = OAF_DHCPV6_ADDR_REG;
+
+
+	if (!lease->duid_len || lease->duid_len > DUID_MAX_LEN)
+		duidstr = "<invalid-duid>";
+	else
+		odhcpd_hexlify(duidbuf, lease->duid, lease->duid_len);
+
+	info("ADDR-REG-INFORM: registered %s for client with DUID %s, "
+		"valid until %ld (in %u seconds)", addrbuf, duidstr,
+		lease_end, valid_lt);
+
+	return true;
+}
+
+/* Send RFC9686 ADDR-REG-REPLY message to client */
+static void send_ia_addr_reg_reply(struct sockaddr_in6 *source,
+		const struct dhcpv6_client_header *hdr,
+		const uint8_t *clientid_data,
+		uint16_t clientid_len,
+		const struct dhcpv6_ia_addr *ia_addr,
+		struct interface *iface)
+{
+	/* RFC9686 §4.3: Reply MUST contain:
+	 * - msg_type: ADDR-REG-REPLY (37)
+	 * - transaction_id: copied from ADDR-REG-INFORM
+	 * - IA Address option: identical to the one in the request
+	 */
+
+	struct {
+		uint8_t msg_type;
+		uint8_t tr_id[3];
+	} _o_packed reply = {
+		.msg_type = DHCPV6_MSG_ADDR_REG_REPLY,
+	};
+	/* Copy transaction ID from request */
+	memcpy(reply.tr_id, hdr->transaction_id, sizeof(reply.tr_id));
+
+	struct {
+		uint16_t code;
+		uint16_t len;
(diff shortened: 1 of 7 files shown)


systemcrash:
Note to self: do we want a configure tuneable? Or do we just set `client_addr_reg_enable` to true only if M or O flags are set. `The client SHOULD NOT send the ADDR-REG-INFORM message unless it has received a Router Advertisement (RA) message with either the M or O flags set to 1.`

src/dhcpv6-ia.c
```diff
@@ -1138,9 +1156,31 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
...
 
+			if (hdr->msg_type == DHCPV6_MSG_ADDR_REG_INFORM) {
+				/* RFC9686 - Address Registration
+				 * Handle client-initiated address registration (e.g. for SLAAC addresses).
+				 * The client registers addresses it has configured.
+				 */
+				char addrbuf[INET6_ADDRSTRLEN];
+				inet_ntop(AF_INET6, &addr->sin6_addr, addrbuf, sizeof(addrbuf));
+
+				if (ia_addr_present && !dhcpv6_ia_on_link(ia, a, iface)) {
```
Alphix: I don't this this is enough. I suggest you rework `ia_addr_present` so that it's not a boolean but an unsigned int (with a count of the IAs), then check that it is exactly 1 here (`The ADDR-REG-INFORM message...MUST contain exactly one IA Address option`).

Also, I see no check for PDs here? If the client includes an IA_PD, that should also be rejected, right?

src/dhcpv6-ia.c
```diff
@@ -1295,6 +1351,9 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
 		buf += ia_response_len;
 		buflen -= ia_response_len;
 		response_len += ia_response_len;
+		/* RFC9686 §4.2.1 If the message passes the verification, the server: 
+		 * MUST log the address registration information.
+		 * odhcpd logs this here anyway as part of normal operation. */
```
Alphix: I would suggest you drop this comment, it's kind of misleading since it only concerns one requirement from one RFC, but we log the outcome of any and all kind of DHCPv6 operations in the next line...pointing just to RFC9686 is a bit misleading...

src/dhcpv6-ia.c
```diff
@@ -1052,6 +1057,19 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
...
 
+				if (hdr->msg_type == DHCPV6_MSG_ADDR_REG_INFORM) {
+					struct dhcpv6_ia_addr *ia_addr = (struct dhcpv6_ia_addr *)&sdata[-4];
+					/* RFC9686 §4.2 "fate sharing" or §4.2.1
+					 * Servers MUST discard any ADDR-REG-INFORM messages where
+					 * the IP address in the IA Address option does not match the
+					 * source address of the original ADDR-REG-INFORM message sent
+					 * by the client */
+					if(memcmp(&ia_addr->addr, &addr->sin6_addr, sizeof(struct in6_addr)) != 0)
+						return -1;
```
Alphix: I don't think you're taking relayed messages into account here (i.e. ones where some other server is the relay and `odhcpd` is the final destination), see 9686, §4.2.1, third bullet point

src/dhcpv6-ia.c
```diff
@@ -1138,9 +1156,31 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
 
 		if (hdr->msg_type == DHCPV6_MSG_SOLICIT ||
 				hdr->msg_type == DHCPV6_MSG_REQUEST ||
-				(hdr->msg_type == DHCPV6_MSG_REBIND && !a)) {
+				(hdr->msg_type == DHCPV6_MSG_REBIND && !a) ||
+				(hdr->msg_type == DHCPV6_MSG_ADDR_REG_INFORM && !a)) {
```
Alphix: There's more cases to consider? If `a` is true, then that address is already taken? Also, there might be a static lease configured (even though it might not be in use), we should check for that as well....in both cases, we should log and drop (4.2.1, second last paragraph)

Alphix:
And some more random points that don't fit some specific line(s) in the diff:

 - I think that there's no check that the to-be-registered address is of global scope? (GUA/ULA)
 - I think you should add a bool to `struct dhcpv6_lease` to indicate that this is a self-registered address and not a "normal" lease
 - That status should probably be included in the `ubus` flags for the DHCPv6 lease

(personal opinion) The RFC comments are slowly getting a bit out of hand, while a quick:
```C
/* See RFC527, second paragraph */
if (something_unexpected)
        do_something_drastic();
```

Is helpful, citing ever longer passages from RFCs just blows up the size of the code, and anyone who is interested should _really_ go read the referred-to paragraph/section of the relevant RFC anyway to make sure that they get the whole context.

Alphix:
> Note to self: do we want a configure tuneable?

I don't think it's necessary...if we're running a DHCPv6 server, it doesn't seem to serve any real purpose to not allow clients to inform us of their autoconfigured addresses...?

> Or do we just set `client_addr_reg_enable` to true only if M or O flags are set.

If the `M` and `O` flags aren't set, we shouldn't expect to get any DHCPv6 requests in the first place? But another interesting twist would be if `A` isn't set and a client address registration msg comes in...we've told clients not to do autoconf...

src/dhcpv6-ia.c
systemcrash: This is a pre-existing mechanism. And this is how the system does it today. But I can check we have exactly one, yes.

IA_PD? `Clients MAY include other options`. So we just ignore them if they're present. The spec makes no mention of how to behave if they're present.

src/dhcpv6-ia.c
systemcrash: Agreed. I had a logging element for RFC9686, but took it away and kept the comment.

src/dhcpv6-ia.c
systemcrash: No - and I don't think it's worth our time to unwrap and perform this check for something we relay since the mandates of the RFC are already quite strict. Here we check before the first wrap.

src/dhcpv6-ia.c
systemcrash: This case is taken care of higher up - a is set, as part of normal lease operations, and the lease is updated. These 4 message types just lodge a new lease. Static leases are checked earlier in the code (irrespective of message type).

The spec is clear about 'refreshing timers' and updating leases - that's all implicitly handled (testing will establish this), so we should not drop those - but use those to update our leases.

src/dhcpv6-ia.c
Alphix: > This is a pre-existing mechanism. And this is how the system does it today. But I can check we have exactly one, yes.

I know it's pre-existing, but with this RFC, you have to check that there's *exactly* one IA_NA...see §4.2:

```
The ADDR-REG-INFORM message...MUST contain exactly one IA Address option containing the address being registered.```
```

> `Clients MAY include other options`. So we just ignore them if they're present. The spec makes no mention of how to behave if they're present.

Read the sentence before the one you quoted, "other options" refers to OROs, as I understand the RFC...

src/dhcpv6-ia.c
Alphix: I'm not referring to something *we* relay, I mean messages we receive that have been relayed (by some other host)...then this check will fail...

src/dhcpv6-ia.c
Alphix: Good

src/dhcpv6-ia.c
Alphix: Oh, and one more thing, we need to check `!dhcpv6_ia_on_link` *and* if the address being registered is part of a prefix that's been delegated to the client (§4.2.1, first paragraph after the bullet points)

systemcrash:
> * I think that there's no check that the to-be-registered address is of global scope? (GUA/ULA)

You're right. There is not (yet). The spec defines no server behaviour for this in §4.2.1. 

It's mandated as part of §4.2 - what the client must do. ( It does no harm if the client erroneously lodges a LL with us. It's nothing that we would assign anyway from our available prefixes. IOW, it's also harmless for us to check this somewhere. And the implied behaviour is... that we discard it. )

```c
static inline bool IN6_IS_ADDR_GLOBAL(const struct in6_addr *a)
{
	return !(
		IN6_IS_ADDR_UNSPECIFIED(a) ||
		IN6_IS_ADDR_LOOPBACK(a) ||
		IN6_IS_ADDR_MULTICAST(a) ||
		IN6_IS_ADDR_LINKLOCAL(a) ||
		IN6_IS_ADDR_ULA(a)
	);
}
```
?

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html defines others. 

Edit: I think this is largely handled by `if (ia_addr_present && !dhcpv6_ia_on_link(ia, a, iface)) ... else ... `

> * I think you should add a bool to `struct dhcpv6_lease` to indicate that this is a self-registered address and not a "normal" lease

Not a bad idea. 

> (personal opinion) The RFC comments are slowly getting a bit out of hand, while a quick:

Hmm - understand. I think 2-3 lines for a code-block is fine, with exceptions for complex bits. Because this RFC basically uses existing functionality with some extra ifs, it's much clearer that we cover all of the MUST parts in the server behaviour (e.g. via a code search) since most machinery is already implicit.

In short, it'd be nice to get this tested before making more changes, or... you could write out your implementation and we compare notes :)

src/dhcpv6-ia.c
systemcrash: Moved.

src/dhcpv6-ia.c
systemcrash: 'exactly one' check in latest commit.

systemcrash:
OK @Alphix last commit addresses 'exactly one' and: 

> * I think you should add a bool to `struct dhcpv6_lease` to indicate that this is a self-registered address and not a "normal" lease
> 
> * That status should probably be included in the `ubus` flags for the DHCPv6 lease

Alphix:
> > * I think that there's no check that the to-be-registered address is of global scope? (GUA/ULA)
> 
> You're right. There is not (yet). The spec defines no server behaviour for this in §4.2.1.
> 
> It's mandated as part of §4.2 - what the client must do. ( It does no harm if the client erroneously lodges a LL with us. It's nothing that we would assign anyway from our available prefixes. IOW, it's also harmless for us to check this somewhere. And the implied behaviour is... that we discard it. )
...
> Edit: I think this is largely handled by `if (ia_addr_present && !dhcpv6_ia_on_link(ia, a, iface)) ... else ... `

Hm, yes, `dhcpv6_ia_on_link` might actually be enough. The only thing it could get "wrong" is if the list of addresses that `dhcpv6_ia_on_link` checks against includes our own linklocal address (I haven't checked if it does, but the list comes from `netlink`, so my guess is that it does).

> In short, it'd be nice to get this tested before making more changes, or... you could write out your implementation and we compare notes :)

Yeah, feel free to not do any changes until you've done more tests...`odhcp6c` support would probably be a good step. I always prefer developing bits and pieces like this with a suitable counterpart to test with...

...and no, I'm not going to write an implementation right now :)

src/dhcpv6-ia.c
systemcrash: OK - last point - I made the forward/relay check re-entrant in the latest commit. That should satisfy this. Thanks!

systemcrash:
Thanks for the comprehensive review @Alphix. This already feels like a better result. If those last few changes look good - then I'll squash. 

This is the on_link check https://github.com/openwrt/odhcpd/blob/03c1468355c0ffb8fa965cdf57f0d3fc8f01f006/src/dhcpv6-ia.c#L902-L934

where 

valid_addr is https://github.com/openwrt/odhcpd/blob/03c1468355c0ffb8fa965cdf57f0d3fc8f01f006/src/dhcpv6-ia.h#L33-L36

That could probably be better named as something like `addr_valid_lifetime` but for the prefix bit.

src/ubus.c
```diff
@@ -156,6 +156,10 @@ static int handle_dhcpv6_leases(_o_unused struct ubus_context *ctx, _o_unused st
 
 			if (a->lease_cfg)
 				blobmsg_add_string(&b, NULL, "static");
+
+			if (a->self_registered_addr)
+				blobmsg_add_string(&b, NULL, "self-registered");
```
Alphix: Nit: I'd suggest

```
if (a->self_generated)
    blobmsg_add_string(&b, NULL, "self-generated");
```

Since the RFC speaks of registration of self-generated addresses.

src/dhcpv6-ia.c
```diff
@@ -951,6 +954,7 @@ ssize_t dhcpv6_ia_handle_IAs(uint8_t *buf, size_t buflen, struct interface *ifac
 	uint8_t *duid = NULL, mac[6] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
 	size_t hostname_len = 0, response_len = 0;
 	bool notonlink = false, rapid_commit = false, accept_reconf = false;
+	bool self_reg_only_one = false;
```
Alphix: That function already takes:
 - 6 arguments
 - 17 top-level variables
 - 23 later variable declarations

IOW, I suggest you go with my suggestion to change `ia_addr_present` to an unsigned count of requested IA_NAs...most of the code just does `if (ia_addr_present)`, so it won't need to change...and you'll avoid introducing toplevel variable number 18 :)

Alphix:
> Thanks for the comprehensive review @Alphix. This already feels like a better result. If those last few changes look good - then I'll squash.

I think I'll just keep my mouth shut now until you have the chance to test this code. Feel free to squash/not squash as/when you prefer...

src/ubus.c
systemcrash: Mmm - how about `registered-self-generated`?

systemcrash:
The good news is that most of this is structurally correct, and almost worked on the first try (I got the replies but IA Addr was missing from response because I used type 3 and not 5) :)

Basically, `assign_na()` stuffs IA_NA(type 3)->IA Addr addresses in our linked list (we have T1, T2, constructed host_id), but RFC9686 sends one naked IA Addr (type 5), while odhcpd IA is set up to handle (non-naked) IA_NA->IA Addr. I think we need a new function specific for these IA Addr that come in (since we also receive ULA which have identical host_id portions).

I'll refresh with what I get working so we can chisel out a good way to handle these records.

```
DHCPv6
    Message type: Address Registration Inform (36)
    Transaction ID: 0x7ede1d
    Client Identifier
        Option: Client Identifier (1)
        Length: 18
        DUID: 0004xx
        DUID Type: Universally Unique IDentifier (UUID) (4)
        UUID: xx
    IA Address
        Option: IA Address (5)
        Length: 24
        IPv6 address: 2001:db8:10:1100:1200:27ff:fe59:96ef
        Preferred lifetime: 2147
        Valid lifetime: 2147

...
DHCPv6
    Message type: Address Registration Reply (37)
    Transaction ID: 0x7ede1d
    Server Identifier
        Option: Server Identifier (2)
        Length: 10
        DUID: 0003000xx
        DUID Type: link-layer address (3)
        Hardware type: Ethernet (1)
        Link-layer address: xx
        Link-layer address (Ethernet): xx
    Client Identifier
        Option: Client Identifier (1)
        Length: 18
        DUID: 0004xx
        DUID Type: Universally Unique IDentifier (UUID) (4)
        UUID: xx
    SOL_MAX_RT
        Option: SOL_MAX_RT (82)
        Length: 4
        Data: 0000003c
    DNS recursive name server
        Option: DNS recursive name server (23)
        Length: 16
         1 DNS server address: fd51:1c2a:8909::1
    Domain Search List
        Option: Domain Search List (24)
        Length: 5
        Domain name suffix search list
    IA Address
        Option: IA Address (5)
        Length: 24
        IPv6 address: 2001:db8:10:1100:1200:27ff:fe59:96ef
        Preferred lifetime: 2147
        Valid lifetime: 2147

```

Alphix:
> Basically, `assign_na()` stuffs IA_NA(type 3)->IA Addr addresses in our linked list (we have T1, T2, constructed host_id), but RFC9686 sends one naked IA Addr (type 5), while odhcpd IA is set up to handle (non-naked) IA_NA->IA Addr. I think we need a new function specific for these IA Addr that come in (since we also receive ULA which have identical host_id portions).

What a coincidence, I've been fiddling with the same stuff because of my work with creating better statefiles that can be used to persist leases across a restart/reboot. I think I might have something that can benefit both of us....basically I want to add an array of real IPv6 addresses to `struct dhcpv6_lease`, so that we can know exactly which addresses have been handed out (the `assigned_host_id` is not enough for that...consider a situation where an `iface` gets another prefix after it has already handed out a lease, for example)...

systemcrash:
Yeah, my solution at present is just to jam the peer IP in there (if checks pass), since this RFC mandates the IP Addr == sender addr. Now the ubus lease handling stuff needs tweaking 😅

systemcrash:
OK - the exchanges work with this, although the lease handling is still not yet ready.

Alphix:
> OK - the exchanges work with this, although the lease handling is still not yet ready.

Did you implement it in `odhcp6c` as well, or how are you testing the functionality?

Edit: and if you give me some time, I think I'll have something for you wrt. the lease handling...

systemcrash:
Implemented in 6c. Easiest way to test.

systemcrash:
OK - I've been testing this one out against 6c, and it works fine:

odhcp6c request
```
DHCPv6
    Message type: Address Registration Inform (36)
    Transaction ID: 0x793fe6
    Client Identifier
        Option: Client Identifier (1)
        Length: 18
        DUID: 0004xxx
        DUID Type: Universally Unique IDentifier (UUID) (4)
        UUID: xxx
    IA Address
        Option: IA Address (5)
        Length: 24
        IPv6 address: 2001:db8:20:1100:a00:27ff:fe59:96ef
        Preferred lifetime: 1917
        Valid lifetime: 1917
```
odhcpd reply:
```
DHCPv6
    Message type: Address Registration Reply (37)
    Transaction ID: 0x793fe6
    Server Identifier
        Option: Server Identifier (2)
        Length: 10
        DUID: 0003xx
        DUID Type: link-layer address (3)
        Hardware type: Ethernet (1)
        Link-layer address: xx
        Link-layer address (Ethernet): xx
    Client Identifier
        Option: Client Identifier (1)
        Length: 18
        DUID: 0004xxx
        DUID Type: Universally Unique IDentifier (UUID) (4)
        UUID: xxx
    IA Address
        Option: IA Address (5)
        Length: 24
        IPv6 address: 2001:db8:20:1100:a00:27ff:fe59:96ef
        Preferred lifetime: 1917
        Valid lifetime: 1917
```

and `ubus call dhcp ipv6leases` produces:

```json
{
	"duid": "0004xx",
	"iaid": 0,
	"hostname": "",
	"accept-reconf": false,
	"flags": [
		"bound",
		"self-generated"
	],
	"ipv6-addr": [
		{
			"address": "fd51:1c2a:8909:0:a00:27ff:fe59:96ef",
			"preferred-lifetime": 1325,
			"valid-lifetime": 1325
		}
	],
	"valid": 4297
},
{
	"duid": "0004xx",
	"iaid": 0,
	"hostname": "",
	"accept-reconf": false,
	"flags": [
		"bound",
		"self-generated"
	],
	"ipv6-addr": [
		{
			"address": "2001:db8:20:1100:a00:27ff:fe59:96ef",
			"preferred-lifetime": 1325,
			"valid-lifetime": 1325
		}
	],
	"valid": 1325
}
```

Alphix:
Awesome, I'll post a draft of the thing I'm working on....hopefully this evening (CET) or tomorrow...I hope you'll find that both of our projects will dovetail nicely...

systemcrash:
Let's see - I'm happy with where things are handling leases at present here - but you typically reveal something I hadn't thought of.

Alphix:
> Awesome, I'll post a draft of the thing I'm working on....hopefully this evening (CET) or tomorrow...I hope you'll find that both of our projects will dovetail nicely...

@systemcrash see #364

systemcrash:
I think this is ready for general consumption now. @Alphix you seem to dislike passing loads of parameters, but it feels like we can pass addrbuf once or twice to avoid doing inet_ntop at different steps. 

The 6c half needs some more time - cleanup and hooking in to stats. Currently it triggers on RA, which doesn't seem optimal... but it works. So I can throw it out there so at least others can test this.

src/ubus.c
systemcrash: Now `self-generated`

schuettecarsten:
Any update on this or is it outdated?

PR #356, opened 2025-12-14
src: rename ra_pio prefix to addr

Noltari:
We're storing the full interface address instead of the prefix, so renaming it from prefix to address should make the code more clear.

CC @Alphix @systemcrash

### DIFF
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -346,7 +346,7 @@ struct dnr_options {
 
 // RA PIO - RFC9096
 struct ra_pio {
-	struct in6_addr prefix;
+	struct in6_addr addr;
 	uint8_t length;
 	time_t lifetime;
 };
--- a/src/statefiles.c
+++ b/src/statefiles.c
@@ -108,6 +108,7 @@ static void statefiles_finish_tmp_file(int dirfd, FILE **fpp, const char *prefix
 	renameat(dirfd, ODHCPD_TMP_FILE, dirfd, filename);
 }
 
+#define JSON_ADDRESS "address"
 #define JSON_LENGTH "length"
 #define JSON_PREFIX "prefix"
 #define JSON_SLAAC "slaac"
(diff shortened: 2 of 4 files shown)


src/statefiles.c
```diff
@@ -235,20 +236,22 @@ void statefiles_read_prefix_information(struct interface *iface)
 		if (!length_json)
 			continue;
 
-		prefix_json = json_object_object_get(cur_pio_json, JSON_PREFIX);
-		if (!prefix_json)
+		addr_json = json_object_object_get(cur_pio_json, JSON_ADDRESS);
+		if (!addr_json)
+			addr_json = json_object_object_get(cur_pio_json, JSON_PREFIX);
+		if (!addr_json)
 			continue;
```
Noltari, line 243: @Alphix we're handling both "address" and "prefix", so old JSON files are still valid.

Alphix:
Well, you already know my opinion<sup>[1](#myfootnote1)</sup>, and I don't want to flog a proverbial dead horse here.

<a name="myfootnote1">1</a>:<sub>I think this goes in the opposite direction of where it should go. Knowing the address and prefix the router happened to have when it announced the PIO is just noise compared to knowing the exact PIO which was announced (IMHO).</sub>

src/odhcpd.h
```diff
@@ -346,7 +346,7 @@ struct dnr_options {
 
 // RA PIO - RFC9096
 struct ra_pio {
-	struct in6_addr prefix;
+	struct in6_addr addr;
```
systemcrash, line 349: The RFC calls it a prefix https://www.rfc-editor.org/rfc/rfc8415#section-21.22
but either way is fine. I have no preference.

Noltari: > The RFC calls it a prefix https://www.rfc-editor.org/rfc/rfc8415#section-21.22 but either way is fine. I have no preference.

Then I'll keep it as is.
Thanks @systemcrash!

PR #357, opened 2025-12-14
config: dhcpv6_pd_min_len changes

Alphix:
I intentionally split this into two patches to separate the issues.

First, the dhcpv6_pd_min_len allowed range is changed from `[1,62]` to `[1,64]`. I am not aware of any reason why the admin shouldn't be allowed to restrict PDs to a `/64` now that we have end-user devices (and not just routers) which might request a PD.

Second, change the default from `0` to `62`. This one might be more controversial, but I don't think there is a perfect one-size-fits-all answer, and `62` at least provides some protection against pathological cases (like downstream routers requesting a `/57` out of a `/56` upstream delegation). The default is, of course, a default and can be overridden.

There's also a small bugfix here in that the limit wasn't correctly enforced before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ and may also receive information from ubus
 | dhcpv6_na		|bool	| 1	| DHCPv6 stateful addressing hands out IA_NA - Internet Address - Network Address |
 | dhcpv6_pd		|bool	| 1	| DHCPv6 stateful addressing hands out IA_PD - Internet Address - Prefix Delegation (PD) |
 | dhcpv6_pd_preferred   |bool | 0 | Set the DHCPv6-PD Preferred (P) flag in outgoing ICMPv6 RA message PIOs (RFC9762); requires `dhcpv6` and `dhcpv6_pd`. |
-| dhcpv6_pd_min_len	|integer| -	| Minimum prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,62] |
+| dhcpv6_pd_min_len	|integer| -	| Minimum prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,64] |
 | router		|list	|`<local address>`| IPv4 addresses of routers on a given subnet (provided via DHCPv4, should be in order of preference) |
 | dns			|list	|`<local address>`| DNS servers to announce, accepts IPv4 and IPv6 |
 | dnr			|list	|disabled| Encrypted DNS servers to announce, `<priority> <domain name> [<comma separated IP addresses> <SvcParams (key=value)>...]` |
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ and may also receive information from ubus
 | dhcpv6_na		|bool	| 1	| DHCPv6 stateful addressing hands out IA_NA - Internet Address - Network Address |
 | dhcpv6_pd		|bool	| 1	| DHCPv6 stateful addressing hands out IA_PD - Internet Address - Prefix Delegation (PD) |
 | dhcpv6_pd_preferred   |bool | 0 | Set the DHCPv6-PD Preferred (P) flag in outgoing ICMPv6 RA message PIOs (RFC9762); requires `dhcpv6` and `dhcpv6_pd`. |
-| dhcpv6_pd_min_len	|integer| -	| Minimum prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,64] |
+| dhcpv6_pd_min_len	|integer| 62	| Minimum prefix length to delegate with IA_PD (adjusted, if necessary, to be longer than the interface prefix length).  Range [1,64] |
 | router		|list	|`<local address>`| IPv4 addresses of routers on a given subnet (provided via DHCPv4, should be in order of preference) |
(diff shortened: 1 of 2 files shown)


Alphix:
Another reasonable default might be `/60`....not 100% sure what's better

@Noltari and @systemcrash ^

GoetzGoerisch:
Thank you! /62 is fine as a default from my PoV

Do we need a Luci change as well?

Noltari (approved):
LGTM

@systemcrash any objections?

systemcrash:
> First, the dhcpv6_pd_min_len allowed range is changed from `[1,62]` to `[1,64]`. I am not aware of any reason why the admin shouldn't be allowed to restrict PDs to a `/64` now that we have end-user devices (and not just routers) which might request a PD.

Thank gawd. I was wondering that (62) also. It seems to have been that way since the start. 

It's my understanding that we recently added fixes so that /96 is possible. So I don't think it makes sense to cap a limit there. Especially if we want more flexibility in the future, e.g. /112. There is also that NAT64 /96 thing - how is that affected here?

Abstractly, I think as long as it's between ... say, /16 and /128 then it should be acceptable. PD handling only really seems to work up to /64 though, with those /96 exceptions(?).

Alphix:
> It's my understanding that we recently added fixes so that /96 is possible.

For IA_NA DHCPv6 leases, not PD, as far as I can remember...

> So I don't think it makes sense to cap a limit there. Especially if we want more flexibility in the future, e.g. /112.

We can change it in the future...but I'm not so sure that'll happen (for example, RFC7421 is still there).

> There is also that NAT64 /96 thing - how is that affected here?

NAT64 doesn't use PD, it's just a way to tell clients that "for IPv4 addresses, please combine the IPv4 address with this /96 prefix and send your traffic there". So it's not affected.

> Abstractly, I think as long as it's between ... say, /16 and /128 then it should be acceptable. PD handling only really seems to work up to /64 though, with those /96 exceptions(?).

I'm not aware of /96 exceptions...?

systemcrash:
OK - I'm going by memory here calling out what seemed relevant. This is ready.

README.md
```diff
@@ -89,7 +89,7 @@ and may also receive information from ubus
 | dhcpv6_na		|bool	| 1	| DHCPv6 stateful addressing hands out IA_NA - Internet Address - Network Address |
 | dhcpv6_pd		|bool	| 1	| DHCPv6 stateful addressing hands out IA_PD - Internet Address - Prefix Delegation (PD) |
 | dhcpv6_pd_preferred   |bool | 0 | Set the DHCPv6-PD Preferred (P) flag in outgoing ICMPv6 RA message PIOs (RFC9762); requires `dhcpv6` and `dhcpv6_pd`. |
-| dhcpv6_pd_min_len	|integer| -	| Minimum prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,62] |
+| dhcpv6_pd_min_len	|integer| 62	| Minimum prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,64] |
```
GoetzGoerisch: Would change the doc:

Prefix length to delegate with IA_PD (value is adjusted if needed to be greater than the interface prefix length).  Range [1,64]

As it is no longer possible to request larger prefix then the configured.
e.g., if set to /64 a downstream cannot request a /62.

GoetzGoerisch (approved):
LGTM

README.md
Alphix: I think the README is correct? `62` as *minimum* prefix length means *longer* is still allowed...i.e. if `62` is the *minimum* then `63` and `64` (i.e. bigger/larger prefix lengths) are allowed, `64` being the *maximum*.

Yes, a `62` is a larger prefix in the sense that it encompasses more addresses, but `64` is greater than `62` in *length*.

And saying `Prefix length to delegate...` would be wrong, since longer prefixes than this value can still be requested?

systemcrash: Yeah, that thing. Sometimes confusing. Maybe re-word 'greater' -> 'longer (e.g. 64 > 56)'?

GoetzGoerisch: Agreed.

> Yes, a 62 is a larger prefix in the sense that it encompasses more addresses, but 64 is greater than 62 in length.

Easy to mix.

GoetzGoerisch: Do we need a luci-mod-network update?

Alphix: > Do we need a luci-mod-network update?

Yes, here:
https://github.com/openwrt/luci/blob/ce3e393ec2fe5c2571f1374179569739c0535252/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1048

But I'm not sure if we'll backport this, so I don't know if it goes to LuCI master or master and 25.12...

systemcrash: master initially, and pick to 25 if this commit goes into openwrt/25

Alphix: https://github.com/openwrt/luci/pull/8155

Noltari: > master initially, and pick to 25 if this commit goes into openwrt/25

I'm OK with cherry-picking this for 25.12 if @Alphix agrees.

systemcrash (approved):

README.md
Alphix: Yeah, I think this is small enough...and the functionality might end up being quite useful in the coming year or two...so I agree with a cherry-pick

Noltari:
Merged, thanks @Alphix!

PR #358, opened 2025-12-14
router: fixups

Alphix:
Reduce some log spam, remove some unnecessary memory (re)allocations that are performed on each RA, remove the fallback code for DNSSL which doesn't work (see individual commit messages for details).


Noltari (approved):
LGTM

@systemcrash any objections?

systemcrash:
LGTM

Noltari:
Merged, thanks @Alphix!

PR #361, opened 2025-12-18
Bugfixes

systemcrash:
ping @Alphix @Noltari

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -215,10 +215,14 @@ static int send_reconf(struct dhcpv6_lease *assign)
 
 	memcpy(clientid.data, assign->duid, assign->duid_len);
 
+	size_t serverid_len, clientid_len;
+	serverid_len = sizeof(serverid.code) + sizeof(serverid.len) + ntohs(serverid.len);
+	clientid_len = sizeof(clientid.code) + sizeof(clientid.len) + ntohs(clientid.len);
+
 	struct iovec iov[IOV_TOTAL] = {
 		[IOV_HDR] = { &hdr, sizeof(hdr) },
-		[IOV_SERVERID] = { &serverid, sizeof(serverid) },
-		[IOV_CLIENTID] = { &clientid, sizeof(clientid) },
+		[IOV_SERVERID] = { &serverid, serverid_len },
+		[IOV_CLIENTID] = { &clientid, clientid_len },
 		[IOV_MESSAGE] = { &message, sizeof(message) },
 		[IOV_AUTH] = { &auth, sizeof(auth) },
 	};
--- a/src/config.c
+++ b/src/config.c
@@ -72,8 +72,6 @@ struct sys_conf sys_conf = {
 #define PD_MIN_LEN_MAX		64
 #define PD_MIN_LEN_DEFAULT	62
 
-#define OAF_DHCPV6	(OAF_DHCPV6_NA | OAF_DHCPV6_PD)
-
 enum {
 	IPV6_PXE_URL,
 	IPV6_PXE_ARCH,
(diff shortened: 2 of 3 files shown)


Noltari (approved):
LGTM

@Alphix any objections?

systemcrash:
I've run-tested and validated the length bugfix (in case it's not obvious from the commit message)

src/dhcpv6-ia.c
```diff
@@ -217,11 +217,17 @@ static int send_reconf(struct dhcpv6_lease *assign)
...
+		[IOV_CLIENTID] = { &clientid, 0 },
 		[IOV_MESSAGE] = { &message, sizeof(message) },
 		[IOV_AUTH] = { &auth, sizeof(auth) },
 	};
+	size_t serverid_len, clientid_len;
+
+	serverid_len = sizeof(serverid.code) + sizeof(serverid.len) + ntohs(serverid.len);
+	clientid_len = sizeof(clientid.code) + sizeof(clientid.len) + ntohs(clientid.len);
+	iov[IOV_SERVERID].iov_len = serverid_len;
+	iov[IOV_CLIENTID].iov_len = clientid_len;
```
Alphix: If you move the two `size_t` up to just below the respective structs, you can calculate `clientid_len` right there and `serverid_len` can be calculated in the `if-else` block...then you just set the `IOV_*` values straight away to `serverid_len`/`clientid_len`

Alphix:
Except for that little nit, LGTM

src/dhcpv6-ia.c
systemcrash: That makes two lines of `serverid_len = sizeof(serverid.code) + sizeof(serverid.len) + ntohs(serverid.len);` per if-else, whereas we can do it once after the if-else. So net change is zero. We could skip the size_t altogether and just assign directly to the iov_len. But this is tested and works, and the variables more describe what's going on :)

Alphix: > That makes two lines of `serverid_len = sizeof(serverid.code) + sizeof(serverid.len) + ntohs(serverid.len);` per if-else, whereas we can do it once after the if-else. So net change is zero. We could skip the size_t altogether and just assign directly to the iov_len. But this is tested and works, and the variables more describe what's going on :)

Yes, and that makes it very clear that the server_id_len is actually set depending on the if-else?

```C
struct serverid {
   bla
};
size_t server_id_len = sizeof(serverid.code) + sizeof(serverid.len);
...
if (foo) {
    ...
    server_id_len += x;
} else {
    ...
    server_id_len += y;
}

iovs...
```

While also making it clear in the iov setting that the size_t is never zero or optional or anything. But this is just a suggestion... 🤷🏻‍♂️

Noltari: And what about...?
```C
	size_t serverid_len, clientid_len;
	serverid_len = sizeof(serverid.code) + sizeof(serverid.len) + ntohs(serverid.len);
	clientid_len = sizeof(clientid.code) + sizeof(clientid.len) + ntohs(clientid.len);

	struct iovec iov[IOV_TOTAL] = {
		[IOV_HDR] = { &hdr, sizeof(hdr) },
		[IOV_SERVERID] = { &serverid, serverid_len },
		[IOV_CLIENTID] = { &clientid, clientid_len },
		[IOV_MESSAGE] = { &message, sizeof(message) },
		[IOV_AUTH] = { &auth, sizeof(auth) },
	};
```

Alphix: Yes, I like that one as well.

Noltari: @systemcrash I hope you don't mind that I've changed it myself :)

Noltari:
Merged, thanks @systemcrash!

PR #363, opened 2025-12-19
dhcpv6: set static defines for DHCPv6 option header size

systemcrash:
so the number 4 isn't so mysterious at first glance.

ping @Noltari optional review also from @Alphix

### DIFF
--- a/src/odhcpd.h
+++ b/src/odhcpd.h
@@ -194,6 +194,8 @@ enum odhcpd_assignment_flags {
 	OAF_DHCPV6_PD		= (1 << 1),
 };
 
+#define DHCPV6_OPT_HDR_SIZE 4
+
 /* 2-byte type + 128-byte DUID, RFC8415, §11.1 */
 #define DUID_MAX_LEN 130
 /* In theory, 2 (type only), or 7 (DUID-EN + 1-byte data), but be reasonable */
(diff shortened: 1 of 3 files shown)


src/odhcpd.h
```diff
@@ -194,6 +194,8 @@ enum odhcpd_assignment_flags {
 	OAF_DHCPV6_PD		= (1 << 1),
 };
 
+#define DHCPV6_OPT_HDR_SIZE 4
+#define DHCPV6_OPT_HDR_SIZE_U 4U
```
Noltari: `DHCPV6_OPT_HDR_SIZE_U` seems to be unused, so we can remove it

systemcrash: Fixed

Alphix:
LGTM, only question mark is if the `#define` shouldn't be in `src/dhcpv6.h` instead?

systemcrash:
Doesn't look like it. The other length defines are in `odhcpd.h`.

Alphix:
> Doesn't look like it. The other length defines are in `odhcpd.h`.

Fine by me, LGTM

Noltari (approved):

Noltari:
Merged, thanks @systemcrash!

PR #360, opened 2025-12-16
dhcpv6-ia: IA case refactor

systemcrash:
The semantics and logic are preserved, and hopefully readability is improved by breaking up some of those monster if conditions. 

@Alphix @Noltari

PR #364, opened 2025-12-19
[WIP] Persist leases across reboot/restart

Alphix:
This is a rough draft/WIP of code to store and persist all state of DHCPv[46] leases across a reboot/restart. Requires a lot more work to handle all corner cases (e.g. if interface addresses change across a reboot/restart, we need to handle that properly and trigger reconfiguration events for clients that support it, and probably drop leases for clients that don't, etc, etc, etc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,11 +62,12 @@ and may also receive information from ubus
 | Option	| Type	|Default| Description |
 | :------------ | :---- | :----	| :---------- |
 | maindhcp	| bool	| 0	| Use odhcpd as the main DHCPv4 service |
-| leasefile	| string|	| DHCPv4/6 lease file |
-| leasetrigger	| string|	| Lease trigger script |
+| leasefile	| string|	| DHCPv4/6 lease file (legacy) |
+| leasetrigger	| string|	| Lease trigger script (legacy) |
+| statedir	| string|	| DHCPv4/v6 state directory (for persisting information such as active leases across reboots/restarts) |
 | hostsdir	| string|	| DHCPv4/v6 hostfile directory (one file per interface will be created) |
-| loglevel	|integer| 6	| Syslog level priority (0-7) |
 | piodir	|string |	| Directory to store IPv6 prefix information (to detect stale prefixes, see RFC9096, §3.5) |
+| loglevel	|integer| 6	| Syslog level priority (0-7) |
 | enable_tz |bool | 1 | Toggle whether RFC4833 timezone information is sent to clients, if set in system  |
 
 
--- a/src/config.c
+++ b/src/config.c
@@ -631,28 +631,28 @@ int config_set_lease_cfg_from_blobmsg(struct blob_attr *ba)
 {
 	struct blob_attr *tb[LEASE_CFG_ATTR_MAX], *c;
 	struct lease_cfg *lease_cfg = NULL;
-	int mac_count = 0;
-	struct ether_addr *macs;
-	int duid_count = 0;
+	struct ether_addr *macaddrs;
+	int macaddrs_cnt = 0;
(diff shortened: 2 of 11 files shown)


Alphix:
@systemcrash no need to review the actual code yet, but there's two important changes here that I'd like to draw your attention to, both of them are in the definition of `struct dhcpv6_lease`.

The first one being that the flags have been replaced with an enum to define which kind of lease it is (here I guess you could add another value to the enum for the self-generated addresses).

The other one being that the addresses/prefixes that the client has been given (or with your PR, registered) are stored explicitly in an array rather than being generated on the fly from the addresses available on a given interface (which might actually be misleading/incorrect).

Ignore the pile of patches and look at the changes to `src/odhcpd.h`...does that look like something that could be of use to you?

systemcrash:
I took a deeper dive in this, and I like most of the stuff in here. I like flags, but since we don't really need to check combos, those can become enum. I see you made similar switch/case changes I did in #360

The only thing I'm a bit ick about is the statefile part. That's a great candidate for json. The 9686 thing does things a bit differently, and likely dumps some zeroes there. It does have one perceived benefit: dissuading external consumers away from trying to use/rely on the state-file 😈 

> if interface addresses change across a reboot/restart, we need to handle that properly

IIUC, the present system handles that: reviewing prefixes and if addrs will find that present leases do not fit, so they can just be free'd. 

That `ADDR_MATCH_PIO_FILTER` macro is a good candidate for renaming to e.g. `PIO_EXCLUDES_ADDR_FILTER` since that seems to be what it does.

Alphix:
> I took a deeper dive in this, and I like most of the stuff in here. I like flags, but since we don't really need to check combos, those can become enum.

Not just that we don't need to check combos, any code checking for combos would be buggy because all the "flags" are mutually exclusive (meaning they're not flags in the traditional sense)....hence the enum.

> I see you made similar switch/case changes I did in #360

Yeah, that was one of the reason why I posted this early, as a heads-up that we're mucking around in similar areas right now :)

> The only thing I'm a bit ick about is the statefile part. That's a great candidate for json. 

Not really. Replacing a single line `sscanf`/`printf` with several hundred lines of JSON write/read code, plus lots of allocations and stuff...for no real benefit.

> The 9686 thing does things a bit differently, and likely dumps some zeroes there.

Not a problem at all...the code is meant to store exactly what's in `struct dhcpv[46]_lease`, if the struct contains some zeroes, then some zeroes should end up in the statefile.

> It does have one perceived benefit: dissuading external consumers away from trying to use/rely on the state-file 😈

I almost considered adding a big fat warning at the top of each statefile...but experience tells me that such warnings are ignored 🤷🏻‍♂️ 

> > if interface addresses change across a reboot/restart, we need to handle that properly
> 
> IIUC, the present system handles that: reviewing prefixes and if addrs will find that present leases do not fit, so they can just be free'd.

The present system has no notion of carrying leases across reboot/restart so it can't possibly handle it...but yeah, the current code *does* have a lot of duplication because the logic in `build_ia()` is basically duplicated in `odhcpd_enum_addr6()` (a function that I really want to remove, and this PR is a step towards that).

> That `ADDR_MATCH_PIO_FILTER` macro is a good candidate for renaming to e.g. `PIO_EXCLUDES_ADDR_FILTER` since that seems to be what it does.

Well...it's the code calling `ADDR_MATCH_PIO_FILTER` that does the actual exclusion, so if you want to rename it, it should (IMHO) rather be something like `PREFIX_FILTER_MATCH`, (there's no _PIOs_ involved, just a check against the cfg option called `prefix_filter`, and the `ADDR_` part adds no information since the addr is the argument to the function).

PR #365, opened 2025-12-22
dhcpv6: allow INFO_REQ for SLAAC only

Noltari:
Allow DHCPV6_MSG_INFORMATION_REQUEST for RA (SLAAC) only configurations.

@Alphix @systemcrash not sure if this is correct, but I've noticed that clients keep sending DHCPv6 Information Requests to odhcpd server and it never replies since DHCPv6 is considered as disabled if it's configured as RA only (no DHCPv6 managed mode).

Marking as draft for now...

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1969,9 +1969,14 @@ struct lease_cfg *config_find_lease_cfg_by_ipv4(const struct in_addr ipv4)
 void reload_services(struct interface *iface)
 {
 	if (iface->ifflags & IFF_RUNNING) {
+		bool enable_dhcpv6;
+
+		enable_dhcpv6 = iface->dhcpv6 != MODE_DISABLED;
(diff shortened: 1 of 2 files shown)


systemcrash:
But shouldn't it be like that? Now this forces DHCPv6 to run if RA is in server mode. We should rather disable the O flag if DHCPv6 is off, no?

Alphix:
> But shouldn't it be like that? Now this forces DHCPv6 to run if RA is in server mode. We should rather disable the O flag if DHCPv6 is off, no?

There is no one-to-one correspondence between the flags in the RA and whether we offer DHCP services, because there could still be a DHCP server on the same subnet, but it could be on a different host...

systemcrash:
Then I think it's correct as is. I considered the possibility that radvd is running elsewhere and odhcpd can't control that aspect. I merely mean that if we have control over the O flag that we send, we can disable that.

PR #367, opened 2025-12-30
netlink: clean up sockets, close files

### LINKED ISSUE #366: cleanup
ping @Alphix you've done some memory profiling on this daemon. How does it look for repeated reloads/reinit? Does netlink implicitly close/free sockets on termination?

I see we set up sockets, and only clean them up in failure modes. But I don't see that those are freed on shutdown. I added some general cleanup on shutdown to 6c, perhaps something similar here?

systemcrash:
Sockets are handled for most usage paths, except for shutdown. Handle those at shutdown.

Closes #366

### DIFF
--- a/src/netlink.c
+++ b/src/netlink.c
@@ -53,6 +53,29 @@ static struct event_socket rtnl_event = {
 	.sock_bufsize = 133120,
 };
 
+/* Shut down and free netlink sockets/registration. Safe to call multiple times. */
+static void netlink_shutdown(void)
+{
+	/* Deregister event and free the event socket */
+	if (rtnl_event.sock) {
+		odhcpd_deregister(&rtnl_event.ev);
+
+		if (rtnl_event.ev.uloop.fd >= 0) {
+			close(rtnl_event.ev.uloop.fd);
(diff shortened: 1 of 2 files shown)


systemcrash:
ping @Alphix

Alphix:
Question: `atexit` doesn't have to be called in `main()`. Would it be possible to call it in the respective netlink/router registration code so that the init/shutdown logic stays internal to those files (and that the functions can be static instead of exported)?

systemcrash:
I have no preference either way - but the init routines are likewise not static because they're called from `odhcpd.c` main(). So I figured parity. How about now?

systemcrash:
Welp, I've taken it for a quick spin and nothing exploded :)

systemcrash:
At your leisure, then, @Noltari

Alphix:
Might be a silly question, but why are we bothering with cleanup on exit? Is it because you get spam in some profiling tool otherwise?

systemcrash:
Just good practice, if nothing else.

Noltari:
Merged, thanks @systemcrash!

PR #369, opened 2026-01-12
send DHCPv6 lease events over U-Bus

ekatwired:
odhcpd currently only sends DHCPv4 lease events over U-Bus. For DHCPv6 the only option was to use a trigger hook script and then parse a lease file and diff it against previous version.

This made it very inconvenient to use in cases like adding DHCP hosts to DNS server and the behavior is also inconsistent with DHCPv4 counterpart. Format of the U-Bus message is based on reply of `ipv6leases` U-Bus method.

This patch is largely based on https://github.com/mikma/openwrt-odhcpd/commit/b796d613dfe74fd36a8df8564c51907321f6a187

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -208,7 +208,7 @@ static void dhcpv6_ia_free_assignment(struct dhcp_assignment *a)
 		close(a->managed_sock.fd.fd);
 	}
 
-	if ((a->flags & OAF_BOUND) && (a->flags & OAF_DHCPV6_PD))
+	if (a->flags & OAF_BOUND)
 		apply_lease(a, false);
 
 	if (a->fr_cnt)
@@ -591,6 +591,10 @@ void dhcpv6_ia_write_statefile(void)
 static void __apply_lease(struct dhcp_assignment *a,
 		struct odhcpd_ipaddr *addrs, ssize_t addr_len, bool add)
 {
+	#ifdef WITH_UBUS
+		ubus_bcast_dhcp6_event(add ? "dhcpv6.ack" : "dhcpv6.release", a, addr_len, addrs);
+	#endif
+
 	if (a->flags & OAF_DHCPV6_NA)
 		return;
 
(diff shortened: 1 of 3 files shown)


Noltari:
@erin-desu there are conflicts with current master, so this branch has to be rebased.

systemcrash:
This was part of @mikma's patch-set to make WireGuard more useable,although he found another way to make WireGuard work without this patch, although it may still have some utility. 

@mikma - can you comment?

src/dhcpv6-ia.c
```diff
@@ -412,48 +838,69 @@ static bool is_reserved_ipv6_iid(uint64_t iid)
...
+	struct dhcp_assignment *c;
+	uint32_t seed = 0;
 
 	/* Preconfigured assignment by static lease */
 	if (a->assigned_host_id) {
 		list_for_each_entry(c, &iface->ia_assignments, head) {
 			if (!(c->flags & OAF_DHCPV6_NA) || c->assigned_host_id > a->assigned_host_id ) {
 				list_add_tail(&a->head, &c->head);
+				if (a->flags &OAF_BOUND)
+					apply_lease(a, true);
```
kempniu, line 852: `apply_lease()` is currently used for installing routes to prefixes delegated to DHCPv6-PD clients.  (Ab)using it for sending ubus notifications for address leases not only hurts readability, but it is also a very slippery slope - and it already shows.

For lease renewals, this PR causes an erroneous `dhcpv6.release` notification to be emitted just before a `dhcpv6.ack` one for the same lease.  Why?  `assign_na()` is called from `dhcpv6_ia_handle_IAs()`, shortly after `apply_lease(a, false);` is called [on line 1522][1].  Note that no release actually happens from lease management perspective; this behavior is just an artifact of emitting notifications from `__apply_lease()`.

Please see #393 for what I hope is a more maintainable counterproposal.

[1]: https://github.com/openwrt/odhcpd/blob/5557d01c7170c5ff19dcda20c62e4bb3aa483689/src/dhcpv6-ia.c#L1522

Issue #371, opened 2026-01-13
odhcpd do_page_fault()


updatede:
latest openwrt snapshot image, odhcpd used as ipv4 server, after this page fault ,it doesn't reply client dhcp request.

```
kern.info kernel: [300136.922357] do_page_fault(): sending SIGSEGV to odhcpd for invalid read access from 000000d0
kern.info kernel: [300136.930995] epc = 555b29dd in odhcpd[129dd,555a0000+18000]
kern.info kernel: [300136.936642] ra  = 555b29d5 in odhcpd[129d5,555a0000+18000]
```

mmmspatz:
I think the correct fix is to replace `avl_remove_all_elements()` with `avl_for_each_element_safe()`, not to just omit the `avl_delete()` in `dhcpv4_free_lease()` as #376 does.

~~However, looking at `avl.h`, I see that it's apparently unsafe to call `avl_for_each_element()` or `avl_for_each_element_safe()` on an empty list, since those macros involve `avl_first_element()` and `avl_last_element()`: The documentation for both of those says "This function must not be called for an empty tree."~~

~~So, I think additionally we should add a `if (!avl_is_empty(&iface->dhcpv4_leases))` guard at every spot where we iterate over leases.~~

edit: reading avl.h more closely, it is indeed safe to call `avl_for_each_element()` and the like on empty lists, because the pointer returned by `avl_first_element()` is not dereferenced in that case.

mmmspatz:
This bug is new since 24.10; it was introduced in https://github.com/openwrt/odhcpd/pull/306. It would be great to get a fix in before the final 25.12 release.

PR #372, opened 2026-01-15
dhcpv4:  fix do_page_fault() when interface lost carrier

updatede:
This fix issue #371 . According to avl.h, av_delete can’t be called in remove all elements loop.

### DIFF
--- a/src/dhcpv4.c
+++ b/src/dhcpv4.c
@@ -351,7 +351,6 @@ void dhcpv4_free_lease(struct dhcpv4_lease *lease)
 
 	if (lease->iface) {
 		lease->iface->update_statefile = true;
-		avl_delete(&lease->iface->dhcpv4_leases, &lease->iface_avl);
 	}
 
 	if (lease->lease_cfg)


systemcrash:
ping @Alphix and @Noltari. WDYT?

Noltari:
@updatede please fix the commit title
As reported at https://github.com/openwrt/odhcpd/actions/runs/21016348202/job/60422128645?pr=372, it should be something like:
`dhcpv4: fix do_page_fault() when interface lost carrier`

systemcrash:
@updatede commit title, not PR title (needs force push) :)

systemcrash:
merge commits are not accepted... just rebase on master if you must

Alphix:
Hm, I could just do a quick check on my phone (not very ergonomic), I assume you want to fix the issue with:

```
bool dhcpv4_setup_interface()
   ...
   avl_remove_all_elements(...)
       dhcpv4_free_lease(...)
```

But `dhcpv4_free_lease()` is called from a bunch of other places, have you checked all callers? It seems likely that we'd end up free'ing but not removing leases in some cases. Wouldn't changing `avl_remove_all_elements()` to `avl_for_each_element_safe()` be a better option?

updatede:
no, I haven’t check any other place where free something twice with av_delete, maybe triple , quadruple :) no I haven’t check it. Feel free to provide better solution.

PR #376, opened 2026-01-20
dhcpv4: fix do_page_fault() when interface lost carrier

updatede:
This fix issue #371 . According to avl.h, av_delete can’t be called in “remove all elements” loop.


systemcrash:
Fix your SoB line (and github config to match it) https://openwrt.org/submitting-patches#submission_guidelines

Noltari:
Superseded by #381

PR #380, opened 2026-02-10
dhcpv6-ia: dhcpv6_log_ia_addr(): parse return of snprintf

lynxis:
snprintf() returns also in the failure case how many characters would have been required to print the full string.
Check return of snprintf() and ensure the log_ctxt->buf_idx never points outside of the given log_ctxt->buf.

Fixes: TOB-OWRT-5
Reported-by: Trail of Bits


Noltari (approved):

Noltari:
Merged at 13cec018c0d9bd452f4cb5e583c52fdf206cdc31

PR #383, opened 2026-02-28
ubus: fix truncated field in DHCPv6 lease query

mmmspatz:
The `assigned_subnet_id` field in `struct dhcpv6_lease` is `uint32_t`, not `uint16_t`.

Unlikely to matter in practice, since the number of meaningful bits in the field is `(assignment size) - (interface assignemt size)`, so largest typically be `64 - 48 = 16`. But it is possible to assign a /32 to an interface in openwrt.


Noltari (approved):

Noltari:
Merged, thanks @mmmspatz!

PR #389, opened 2026-03-16
[25.12] backport fixes from master

Noltari:
- odhcpd: ensure zero padding on DNSSL
- odhcpd: ignore NULL in dhcpv4_free_lease() stub
- ubus: fix truncated field in DHCPv6 lease query
- dhcpv4: fix avl_delete on leases not in avl tree
- dhcpv4: fix segfault when disabling interface
- dhcpv6-ia: dhcpv6_log_ia_addr(): parse return of snprintf

CC @lynxis @aparcar


aparcar:
Nice, thanks!

Issue #385, opened 2026-03-11
Extend `prefix_filter` to filter out unrelated prefixes in relay mode for RA, DHCPv6 and NDP


lqf96:
Background: I'm trying to provide IPv6 access to machines in my network. The upstream network does not support DHCPv6 PD so I have to use NDP proxy (and odhcpd relay mode) instead. I'd also like to set up a whole ULA prefix on the router so devices can get truly static and stable address legitimately without any collision issues. To achieve this, and because I need to announce custom routes for the ULA prefix, I opted to use odhcpd in relay mode for ULA, and bird3 router advertisement for GUA. I did use odhcpd to assign the router a single IPv6 address in a /64 ULA subnet on the bridge LAN interface, apart from the usual GUA relay settings. bird3 handles the task of announcing a larger /48 ULA route that encompasses the /64 ULA subnet the router is on, and part of this route will be used to connect the router to other sites through a mesh VPN like tailscale.

With this combination I managed to get machines in my network to gain connectivity for both GUA and ULA. However, I found odhcpd to incorrectly relay ULA addresses on the router despite it should only do so for the GUA. When a machine passes data to or through the router using the ULA address, the router generates a /128 proxy route on the LAN, despite the fact that the larger /64 prefix is already configured for the bridge LAN interface. The router may have also incorrectly relay the NDP of ULA to upstream though I can't confirm.

To investigate the cause of this issue, I skimmed though the NDP proxy code in this repo. I think the issue is the NDP proxy code does not provide a mechanism to filter out IPv6 addresses it wishes to proxy. There are only coarse filters to handle invalid IPv6 addresses. Based on this observation, I'm wondering if we can reuse the `prefix_filter` config item for selectively proxying NDP, RA and DHCPv6. Right now this parameter seems to be only used in server mode for RA, but it could be useful to extend it to work in relay mode for everything.

lqf96:
I have a patch on the `extended-prefix-filter` branch on my own fork, which covers RA and NDP relay for now. I will test them next week and if it works will submit a pull request. Patch on DHCPv6 relay will follow because of the complexity of DHCPv6, and also because I want my current code reviewed to make sure maintainers are good with my approach.

PR #390, opened 2026-03-17
Extend `prefix_filter` to work in relay mode for RA and NDP

lqf96:
# Background
See openwrt/odhcpd#385 for details.

# Changes
## General
* Change the naming of `pio_filter_*` to `prefix_filter_*`, since the `prefix_filter` will no longer be used just for RA.
* Added `IN6_MATCH_PREFIX_FILTER` alongside `ADDR_MATCH_PREFIX_FILTER` to test if a single IPv6 address is in the prefix filter.

## NDP
* Apply `prefix_filter` to neighbor solicitation from downstream in `handle_solicit`.
* Apply `prefix_filter` to neighbor and interface address updates on upstream interface in `ndp_netevent_cb`.

## RA
* Apply `prefix_filter` to filter out PIO and RIO from upstream RA in `forward_router_advertisement`.
* Removing PIO and RIO from RA requires editing the RA message before sending it downstream. For this purpose, a utility called `iov_builder` is added to `odhcpd` to solve the problem of applying small edits to a received message before sending it again. See below for the details and usage of this tool.
* Because of the need of editing the RA message for potential PIO and RIO exclusion, and also because how `iov_builder` works, `forward_router_advertisement` is rewritten to perform option look-up and modification together in the per-interface loop, instead of doing so separately in the current implementation. The modifications for options other than PIO and RIO are kept intact.

## `iov_builder`
* `iov_builder` is a utility to apply small edits to an existing message stored in the buffer.
* It ultimately generates an `iovec` array pointing to chunks of the original message or chunks to added data. The `iovec` array can then be written or sent through vector I/O.
* Use `odhcpd_iov_builder_init` to initalize an `iov_builder`. It takes a `iovec` array called `iov_buf` as well as its `iov_capacity`, and pointer to the original message `data`. It initializes the internal editing pointer `current_iov_base` to `data`, set the current chunk length `current_iov_len` to 0 and inclusion `include_iov` to false.
* Use `odhcpd_iov_builder_append` to add an arbitrary `iovec` to the builder. This is a low-level tool. If the `iov_buf` is full the function returns 1.
* Use `odhcpd_iov_builder_advance` to add a chunk from the original message to the `iov_builder`. This advances the internal editing pointer by `chunk_len`. If `include_chunk` is set to false, the chunk is not added to `iov_buf` but is instead skipped.
* Use `odhcpd_iov_builder_finalize` to finalize the edition in the end. You can now use `iov_buf` and `iov_count` for any vectored I/O.

# Questions
* Do we need to consider the case of service reload for NDP relay? Right now in `ndp_netevent_cb` both addition and deletion are guarded on the `prefix_filter` check. But this might not work if the user changes the `prefix_filter` to a more specific prefix and then reload.
* The updated `forward_router_advertisement` function has lower efficiency and may need to allocate for more memory due to the need of editing RA message. I thought about this but couldn't come up with a better approach here. Let me know what you think about this and if there is a better solution. It might be OK since NDP relay doesn't happen a lot unless you have many devices in your network

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1707,8 +1707,8 @@ int config_parse_interface(void *data, size_t len, const char *name, bool overwr
 
 	if ((c = tb[IFACE_ATTR_PREFIX_FILTER]))
 		odhcpd_parse_addr6_prefix(blobmsg_get_string(c),
-					  &iface->pio_filter_addr,
-					  &iface->pio_filter_length);
+					  &iface->prefix_filter_addr,
+					  &iface->prefix_filter_length);
 
 	if (overwrite && (c = tb[IFACE_ATTR_NTP])) {
 		struct blob_attr *cur;
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -295,7 +295,7 @@ static void __apply_lease(struct dhcpv6_lease *a,
 	for (ssize_t i = 0; i < addr_len; ++i) {
 		struct in6_addr prefix;
 
-		if (ADDR_MATCH_PIO_FILTER(&addrs[i], a->iface))
+		if (ADDR_MATCH_PREFIX_FILTER(&addrs[i], a->iface))
 			continue;
 
 		prefix = addrs[i].addr.in6;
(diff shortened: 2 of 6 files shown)


lqf96:
Some information about testing: I tested this patch today on my router and can confirm that `prefix_filter` works for NDP, and regular RA forwarding still works. However I haven't tested RA `prefix_filter` because in my situation there is nothing in the RA to filter. I would appreciate if anyone can test it in their environment.

PR #394, opened 2026-04-21
router: add option to include static routes in RA

### LINKED ISSUE #74: Support adding static router advertisments

dbnk0:
Introduces a new configuration section `ra_route` for adding custom route(s) in route advertisement (RA) messages.

This is useful when, for example, end devices are given only ULA IPv6 addresses (e.g. from `fc00:0:100::/48`), but one still needs inter-subnet communication to e.g. `fc00:0:200::/48`. Currently, this isn't possible without a static route on each end device, as the odhcpd server doesn't present itself as a default router and no route information is provided other than the on-link local ULA subnet (`fc00:0:100::/48`). Even when odhcpd provides a default route, route customization may be useful in a multi-homed setup to set higher or lower route preferences and enforce different routing behavior (see [RFC4191](https://datatracker.ietf.org/doc/html/rfc4191)).

The new configuration section `ra_route` allows advertising arbitrary routes to end devices, enabling correct routing even when no default route (or public IPv6 prefix) is available.

The configuration section includes four attributes:

| Option	| Type		| Default		| Description |
| :------------ | :------------ | :-------------------- | :---------- |
| prefix	| string	| (none)		| Target IPv6 route prefix, e.g. `2000:abc::/48`. |
| interface	| list		| (none)		| List of interfaces to advertise this route on. If no interface is provided, the route is advertised on all interfaces. |
| preference	| string	| `ra_preference`	| Route preference [medium\|high\|low]. By default interface-specific `ra_preference` value is used. |
| lifetime	| string	| `ra_lifetime`	| Value to be placed in the Route Lifetime field of the Route Information. Implicitly clamped to the interval of sending unsolicited RA (`ra_maxinterval`) as a minimum value. Expected format follows `leasetime`, e.g. 60m, 2h, 3d, 1w. |

**Config example:**

```
# Route advertised only to the `lan` and `guest_wifi` interfaces
config ra_route 'route_branch1'
	option prefix 'fc00:10::/32'
	list interface 'lan'
	list interface 'guest_wifi'
	option preference 'high'
	option lifetime '2h'

# Route advertised to all odhcpd interfaces
config ra_route 'route_branch2'
	option prefix 'fc00:20::/32'
	option preference 'low'
	option lifetime 'infinite'

# Minimal route configuration, defaults are inherited from interfaces
config ra_route 'route_branch3'
	option prefix 'fc00:30::/32'
```

> [!NOTE]
> This patch is a rework of my PR: https://github.com/openwrt/odhcpd/pull/224
>
> The previous implementation used attributes on both global and per-interface basis to configure the static routes. This meant that per-route preference and lifetime couldn't be set, and, most importantly, the original PR didn't take [RFC4191](https://datatracker.ietf.org/doc/html/rfc4191) into account, allowing configuration of static routes only when a default route wasn't present. @systemcrash and I agreed on this redesign (see the discussion of the original PR).

Fixes: https://github.com/openwrt/odhcpd/issues/74 and possibly helps with https://github.com/openwrt/odhcpd/issues/152

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,15 @@ and may also receive information from ubus
 [//]: # "dhcpv6_raw - string - not documented, may change when generic DHCPv4/DHCPv6 options are added"
 
 
+### Sections of type ra_route
+| Option	| Type		| Default		| Description |
+| :------------ | :------------ | :-------------------- | :---------- |
+| prefix	| string	| (none)		| Target IPv6 route prefix, e.g. `2000:abc::/48`. |
+| interface	| list		| (none)		| List of interfaces to advertise this route on. If no interface is provided, the route is advertised on all interfaces. |
+| preference	| string	| `ra_preference`	| Route preference [medium\|high\|low]. By default interface-specific `ra_preference` value is used. |
+| lifetime	| string	| `ra_lifetime`	| Value to be placed in the Route Lifetime field of the Route Information. Implicitly clamped to the interval of sending unsolicited RA (`ra_maxinterval`) as a minimum value. Expected format follows `leasetime`, e.g. 60m, 2h, 3d, 1w. |
+
+
 ### Sections of type host (static leases)
 | Option		| Type	|Default| Description |
 | :-------------------- | :---- | :---- | :---------- |
--- a/src/config.c
+++ b/src/config.c
@@ -215,6 +215,26 @@ const struct uci_blob_param_list lease_cfg_attr_list = {
 };
 
 enum {
+	RA_ROUTE_ATTR_PREFIX,
+	RA_ROUTE_ATTR_INTERFACE,
+	RA_ROUTE_ATTR_PREFERENCE,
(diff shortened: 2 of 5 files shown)


systemcrash:
Untested by me but looks OK. How does this align with what you had in mind @Alphix ?

flokli:
I was missing this functionality in openwrt, the config interface looks good to me. Would be nice to see this for 26.x ;-)

PR #399, opened 2026-05-20
config: set default dhcpv4_pool_start/dhcpv4_pool_end

AndyChiang888:
README.md states that the default 'start' value is 100,
but the actual value was 0.

Set the default dhcpv4_pool_start to 100 and the
dhcpv4_pool_end to 249 for consistency.

Link: https://github.com/openwrt/odhcpd/blob/master/README.md

Ref: https://github.com/openwrt/luci/pull/8642


AndyChiang888:
@jow- PTAL

PR #415, opened 2026-08-23
ubus: fold ephemeral dhcp sections from procd service data

dangowrt:
Fold ephemeral `dhcp` sections from procd instance data into the interface list on every reload: query `service get_data {"type":"dhcp"}` and feed each section through `config_parse_interface()`, sharing defaults and validation with UCI. UCI sections win collisions. The first commit defers a nested `odhcpd_reload()`, since libubus dispatches the pending netifd dump reply while the synchronous request waits.

A container manager can then declare DHCP, RA and NDP service on the container's own procd instance, tied to its lifetime: the section vanishes with the instance, which sends a final zero-lifetime RA and drops the leases. Ephemeral sections enable DHCPv4 regardless of `maindhcp`, which arbitrates UCI pool ownership with dnsmasq. firewall4 consumes service data the same way in `read_ubus()`.

Both commits were produced with an LLM coding agent working from a written brief: fold ephemeral sections from procd instance data into the interface list through the UCI code path, re-derive them on every reload, tear down withdrawn interfaces, let UCI win collisions, and tolerate an absent `service` object. Tested on x86/64 with dnsmasq running and `maindhcp=0`: DHCPv4 from the declared pool, RAs and stateful DHCPv6 on a container bridge, teardown on withdrawal, reconvergence after restart, leases intact across repeated reloads; all four build variants compile.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1098,7 +1098,8 @@ static int avl_ipv4_cmp(const void *k1, const void *k2, _o_unused void *ptr)
 	return memcmp(k1, k2, sizeof(struct in_addr));
 }
 
-int config_parse_interface(void *data, size_t len, const char *name, bool overwrite)
+int config_parse_interface(void *data, size_t len, const char *name,
+			   bool overwrite, bool ephemeral)
 {
 	struct interface *iface;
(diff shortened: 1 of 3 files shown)


blogic:
LGTM @dangowrt

PR #420, opened 2026-09-09
dhcpv6-ia: skip invalid parent prefixes for PD routes

### LINKED ISSUE #418: __apply_lease() failed checking whether the delegated prefix
## Problem

`__apply_lease()` installs a route for every non-filtered IPv6 prefix on the
interface when applying an IA_PD lease, without checking whether the delegated
prefix can actually be derived from that parent prefix.

The DHCPv6 reply path already uses `valid_prefix_length()` and skips parent
prefixes that are not shorter than the delegated prefix.

This mismatch can create a route that was never actually delegated to the
client.

For example, with an interface that has:

- a ULA `/60`
- a GUA `/64`
- a downstream client requesting a `/64` IA_PD

the DHCPv6 lease correctly contains only a `/64` derived from the ULA `/60`,
but `__apply_lease()` also installs a route derived from the GUA `/64`:

```
2600:1700:545c:6e1f::/64 via <client-link-local> dev br-lan metric 1024
```

That route replaces the directly-connected route for the LAN GUA prefix and
causes intermittent IPv6 loss until the stale PD route is removed.

## Fix

Apply the same `valid_prefix_length()` check in `__apply_lease()` before
constructing and installing a PD route.

A parent `/60` with a delegated `/64` remains valid, while a parent `/64` with
a delegated `/64` is skipped.

## Reproduction

Observed on OpenWrt with a LAN carrying both:

```
2600:1700:545c:6e1f::/64
fdb0:373:fb52::/60
```

A downstream Thread Border Router/HomePod requests IA_PD `/64`.
`ubus call dhcp ipv6leases` reports only the valid delegated ULA prefix:

```
fdb0:373:fb52:1::/64
```

but the route table intermittently contains:

```
2600:1700:545c:6e1f::/64 via <client-link-local> dev br-lan metric 1024
```

Removing that route and restoring the direct `dev br-lan` route immediately
restores IPv6 connectivity.

The proposed change makes route application consistent with the existing
IA_PD reply validation.

lightingghost:
The IA_PD reply path uses valid_prefix_length() to skip interface prefixes that cannot contain the delegated prefix, but __apply_lease() does not.

As a result, a /64 PD lease on an interface carrying both a shorter prefix and an on-link /64 can install a route for the on-link /64 via the downstream client, even though that prefix was never delegated.

Apply the same prefix-length validation before installing PD routes.

This keeps route installation consistent with the IA_PD reply path and prevents a delegated /64 from replacing the directly-connected route of a /64 parent prefix.

fixes #418

### DIFF
--- a/src/dhcpv6-ia.c
+++ b/src/dhcpv6-ia.c
@@ -298,6 +298,9 @@ static void __apply_lease(struct dhcpv6_lease *a,
 		if (ADDR_MATCH_PIO_FILTER(&addrs[i], a->iface))
 			continue;
 
+		if (!valid_prefix_length(a, addrs[i].prefix_len))
+			continue;
+
 		prefix = addrs[i].addr.in6;
 		prefix.s6_addr32[1] |= htonl(a->assigned_subnet_id);
 		prefix.s6_addr32[2] = prefix.s6_addr32[3] = 0;


lightingghost:
@Alphix @systemcrash @Noltari Could you please take a look at this? Thanks!

